PR #11, opened 2024-09-09
Additional validation for update

### LINKED ISSUE #5: Update storage collection
Create an endpoint where authenticated users can update an existing storage collection.

- [x] new PUT endpoint `/{customerId}/collections/{collectionId}` 

- [x] only the following fields can be edited:

    - label (optional)
    - slug, if slug value changed must be unique
    - parent

- [x] update basic data in database
- [x] updates are rejected if

    - [x] "behavior" does not include "storage-collection" (see [iiif-behaviors](https://deploy-preview-3--dlcs-docs.netlify.app/api-doc/iiif#iiif-behaviors))
    - [x] latest eTag is not provided in headers
    - [x] IIIF-CS-Show-Extras header is not "All"
    - [x] user not authenticated

See related API docs - https://deploy-preview-3--dlcs-docs.netlify.app/api-doc/iiif#example-create-a-storage-collection-within-a-storage-collection

JackLewis-digirati:
This PR resolves #5

It contains the following changes:

- adds more validation rules for values that should be left empty
- makes label optional
- rejects updates if behavior does not contain `storage-collection`


src/IIIFPresentation/API/Features/Storage/Validators/FlatCollectionValidator.cs
```diff
@@ -9,18 +9,33 @@ public FlatCollectionValidator()
...
+            RuleFor(f => f.Context).Empty().WithMessage("Context cannot be set");
+            RuleFor(f => f.Items).Empty().WithMessage("Items cannot be set");
+            RuleFor(f => f.Tags).Empty().WithMessage("Tags cannot be set");
+            RuleFor(f => f.Thumbnail).Empty().WithMessage("Thumbnail cannot be set");
+            RuleFor(f => f.Type).Empty().WithMessage("Type cannot be set");
+            RuleFor(f => f.View).Empty().WithMessage("View cannot be set");
+            RuleFor(f => f.ItemsOrder).Empty().WithMessage("Items order cannot be set");
+            RuleFor(f => f.CreatedBy).Empty().WithMessage("Created by cannot be set");
+            RuleFor(f => f.ModifiedBy).Empty().WithMessage("Modified By cannot be set");
+            RuleFor(f => f.PartOf).Empty().WithMessage("Part of cannot be set");
```
griffri, line 33: Would these responses be slightly easier to read if the properties were enclosed in single quotes? e.g
> 'Modified by' cannot be set

griffri (approved):

src/IIIFPresentation/API/Features/Storage/Validators/FlatCollectionValidator.cs
```diff
@@ -9,18 +9,33 @@ public FlatCollectionValidator()
...
+            RuleFor(f => f.Parent).NotEmpty().WithMessage("Creating a new collection requires a parent");
         });
         
         RuleSet("update", () =>
         {
-            RuleFor(a => a.Created).Empty().WithMessage("Created cannot be set");
-            RuleFor(a => a.Modified).Empty().WithMessage("Modified cannot be set");
-            RuleFor(a => a.Id).Empty().WithMessage("Id cannot be set");
-            RuleFor(a => a.Parent).NotEmpty().WithMessage("Updating a collection requires a parent to be set");
+            RuleFor(f => f.Created).Empty().WithMessage("Created cannot be set");
```
donaldgray, line 20: Would it be easier to use an alternative view-model/poco where these properties don't exist? Would mean that there's no way that they can be provided, which would reduce validation overhead? When viewing on swagger etc consumers would only see properties that can be updated, rather than a bunch that cannot. Would also avoid needing to keep this list updated if new properties were added.

src/IIIFPresentation/API.Tests/Integration/ModifyCollectionTests.cs
```diff
@@ -242,12 +242,133 @@ public async Task UpdateCollection_CreatesCollection_WhenAllValuesProvided()
...
         fromDatabase.Parent.Should().Be(parent);
-        fromDatabase.Label.Values.First()[0].Should().Be("test collection - updated");
+        fromDatabase.Label!.Values.First()[0].Should().Be("test collection - updated");
         fromDatabase.Slug.Should().Be("programmatic-child");
         fromDatabase.IsPublic.Should().BeTrue();
         fromDatabase.IsStorageCollection.Should().BeTrue();
     }
     
+        [Fact]
+    public async Task UpdateCollection_CreatesCollection_WhenAllValuesProvidedWithoutLabel()
```
donaldgray, line 252: Is `UpdateCollection_CreatesCollection` correct? Also `[Fact]` indentation

src/IIIFPresentation/API.Tests/Integration/ModifyCollectionTests.cs
```diff
@@ -242,12 +242,133 @@ public async Task UpdateCollection_CreatesCollection_WhenAllValuesProvided()
...
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+        fromDatabase.Parent.Should().Be(parent);
+        fromDatabase.Slug.Should().Be("programmatic-child-2");
+        fromDatabase.Label.Should().BeNull();
+        fromDatabase.IsPublic.Should().BeTrue();
+        fromDatabase.IsStorageCollection.Should().BeTrue();
+    }
+    
+        [Fact]
+    public async Task UpdateCollection_FailsToCreateCollection_WhenNotStorageCollection()
```
donaldgray, line 317: `FailsToCreate_WhenNotStorageCollection` doesn't seem correct? Wouldn't expect it to create and `IsStorageCollection = true`?

JackLewis-digirati: Will change it to `FailsToUpdate`, but I think the `IsStorageCollection = true` is correct, this is creating a record that is then updated to have  `IsStorageCollection = false`

src/IIIFPresentation/API/Features/Storage/Validators/FlatCollectionValidator.cs
JackLewis-digirati: not sure if this needs another look, but I've updated this to use a POCO for `update` - I'm working on some fixes to `create` on another PR and will look to get this updated for `POST` along with those changes

PR #13, opened 2024-09-10
rename IIIF-CS-Show-Extra header

### LINKED ISSUE #9: Rename IIIF-CS-Show-Extra header
In order to retrieve extra information from the API, a header of `IIIF-CS-Show-Extras` with the value of `All`.  

There's a possible requirement to show domain specific headers with the prefix of `X-`, so this header should be `X-IIIF-CS-Show-Extras`

**NOTE:** This should be checked by the product owner before implementation

JackLewis-digirati:
Resolves #9 

This PR renames the `IIIF-CS-Show-Extra` header to `X-IIIF-CS-Show-Extras`

### DIFF
--- a/src/IIIFPresentation/API/Attributes/EtagCachingAttribute.cs
+++ b/src/IIIFPresentation/API/Attributes/EtagCachingAttribute.cs
@@ -47,7 +47,7 @@ public class EtagCachingAttribute : ActionFilterAttribute
 
             responseHeaders.CacheControl = new CacheControlHeaderValue() // how long clients should cache the response
             {
-                Public = !requestHeaders.Headers.Keys.Contains("IIIF-CS-Show-Extra"),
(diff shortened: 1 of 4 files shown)


src/IIIFPresentation/API/Attributes/EtagCachingAttribute.cs
```diff
@@ -47,7 +47,7 @@ ResultExecutionDelegate next
 
             responseHeaders.CacheControl = new CacheControlHeaderValue() // how long clients should cache the response
             {
-                Public = !requestHeaders.Headers.Keys.Contains("IIIF-CS-Show-Extra"),
+                Public = !requestHeaders.Headers.Keys.Contains("X-IIIF-CS-Show-Extras"),
```
donaldgray, line 50: Should this use something similar to `ShowExtraProperties()` extension method, so that it's all contained in 1 place?

donaldgray (approved):

griffri (approved):

PR #15, opened 2024-09-11
Adding ability for collection items to page and order by

### LINKED ISSUE #8: Support paging of Collections
Add a pager, effective when header supplied, on Storage Collections

example endpoint: `https://localhost:7230/1/collections/b157e325-3fe5-459b-8b80-b7a62213334a?page=1&pageSize=100`

- Implement paging params in the API
- Use Hydra paging mechanism
- Update database calls to correctly page data
- Allow pageSize to be retrieved from GET
- Set a default pageSize
- Set a max pageSize from settings

---

From the PR (linked below)

This PR resolves #8 

In order to support paging, this PR adds a set of query params to GET storage collections.  These are as follows:

- `page`
  - sets the current page to be looked at
- `pageSize`
  - How large each page should be

There are also default page sizes (set to 100 by default) and max page page sizes (1000 by default) which can be set in settings.

The default page size is used when no query params are set, and the max automatically overrides any value larger than it to the max, so that we avoid requests for millions of items in a single page

Additionally, ordering has been added to paging, so that retrieved items are consistent.  There are 2 query params that support this:

- `orderBy`
  - this the default 
- `orderByDescending`

You can order on the current fields:

- `created`
  - this is the default 
- `id`
- `slug`

An example of using these query params would be something like the following:

```
https://localhost:7230/1/collections/0909385a-d8bb-4254-b7a5-f3f3a7939c3a?page=1&pageSize=3&orderByDescending=slug
```

This PR also adds the values `next` and `last` to `view`, which will appear when there are more, and/or previous pages to view

JackLewis-digirati:
This PR resolves #8 

In order to support paging, this PR adds a set of query params to GET storage collections.  These are as follows:

- `page`
  - sets the current page to be looked at
- `pageSize`
  - How large each page should be

There are also default page sizes (set to 100 by default) and max page page sizes (1000 by default) which can be set in settings.

The default page size is used when no query params are set, and the max automatically overrides any value larger than it to the max, so that we avoid requests for millions of items in a single page

Additionally, ordering has been added to paging, so that retrieved items are consistent.  There are 2 query params that support this:

- `orderBy`
  - this the default 
- `orderByDescending`

You can order on the current fields:

- `created`
  - this is the default 
- `id`
- `slug`

An example of using these query params would be something like the following:

```
https://localhost:7230/1/collections/0909385a-d8bb-4254-b7a5-f3f3a7939c3a?page=1&pageSize=3&orderByDescending=slug
```

This PR also adds the values `next` and `last` to `view`, which will appear when there are more, and/or previous pages to view

Finally, if an `orderBy` parameter is uesd that isn't understood, it will silently override with the defaults


src/IIIFPresentation/Core/Helpers/StringX.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System.Diagnostics.CodeAnalysis;
+
+namespace Core.Helpers;
+
+public static class StringX
```
donaldgray, line 5: File moved from API project - are there tests to move too?

src/IIIFPresentation/Models/API/Collection/View.cs
```diff
@@ -15,4 +15,8 @@ public class View
     public int PageSize { get; set; }
     
     public int TotalPages { get; set; }
+    
+    public string? Next { get; set; }
+    
+    public string? Last { get; set; }
```
donaldgray, line 21: Should `Next` and `Last` be a `Uri?` rather than a `string?`

src/IIIFPresentation/API/Converters/CollectionConverter.cs
```diff
@@ -24,60 +22,78 @@ public static Collection ToHierarchicalCollection(this Models.Database.Collectio
...
-            Context = new List<string> 
-                {
-                    "http://iiif.io/api/presentation/3/context.json", 
-                    "http://tbc.org/iiif-repository/1/context.json" 
-                },
+            Id = id,
+            Context = new List<string>
+            {
+                "http://iiif.io/api/presentation/3/context.json",
+                "http://tbc.org/iiif-repository/1/context.json"
```
donaldgray, line 44: The IIIF Presentation context must come last, see https://iiif.io/api/presentation/3.0/#46-linked-data-context-and-extensions (when this is using `Collection` we can use the `.EnsureXX()` method to enforce this)

src/IIIFPresentation/API/Converters/CollectionConverter.cs
```diff
@@ -1,20 +1,18 @@
...
-        IQueryable<Models.Database.Collections.Collection>? items)
+    public static Collection ToHierarchicalCollection(this Models.Database.Collections.Collection dbAsset, 
+        UrlRoots urlRoots, List<Models.Database.Collections.Collection>? items)
     {
-        var stuff = new Collection()
+        var collection = new Collection()
         {
-            Id = $"{urlRoots.BaseUrl}/{dbAsset.CustomerId}{(dbAsset.FullPath != null ? $"/{dbAsset.FullPath}" : "")}",
+            Id = $"{urlRoots.BaseUrl}/{dbAsset.CustomerId}{(string.IsNullOrEmpty(dbAsset.FullPath) ? string.Empty : $"/{dbAsset.FullPath}")}",
             Context = "http://iiif.io/api/presentation/3/context.json",
```
donaldgray, line 16: As this is a `Collection` from the iiif-net package there's a couple of helpers for adding contexts. It ensures that they're ordered correctly and avoids adding multiple times etc.

```cs
collection.EnsurePresentation3Context();
// which is a shortcut for...
collection.EnsureContext(IIIF.Presentation.Context.Presentation3Context);
```

src/IIIFPresentation/API/Converters/CollectionConverter.cs
```diff
@@ -1,20 +1,18 @@
...
 using Models.API.Collection;
 using Models.Infrastucture;
 
 namespace API.Converters;
 
 public static class CollectionConverter
 {
-    public static Collection ToHierarchicalCollection(this Models.Database.Collections.Collection dbAsset, UrlRoots urlRoots,
-        IQueryable<Models.Database.Collections.Collection>? items)
+    public static Collection ToHierarchicalCollection(this Models.Database.Collections.Collection dbAsset,
```
donaldgray, line 10: There's a lot of Id/Url wrangling in this class - could those be in a central place for reuse? Would make any changes easier to apply too

src/IIIFPresentation/API/Features/Storage/Requests/GetCollection.cs
```diff
@@ -36,12 +47,21 @@ public async Task<CollectionWithItems> Handle(GetCollection request,
...
         
-        IQueryable<Collection>? items = null;
+        List<Collection>? items = null;
+        int total = 0;
 
         if (collection != null)
         {
-            items = dbContext.Collections.Where(s => s.CustomerId == request.CustomerId && s.Parent == collection.Id);
-
+            total = await dbContext.Collections.CountAsync(
```
donaldgray, line 55: Worth using `Task.WhenAll()` to allow these to run in parallel?

src/IIIFPresentation/API/Features/Storage/Requests/GetHierarchicalCollection.cs
```diff
@@ -119,24 +119,37 @@ INNER JOIN
...
         {
-            items = dbContext.Collections.Where(s => s.CustomerId == request.CustomerId && s.Parent == storage.Id);
+            items = await dbContext.Collections
+                .Where(s => s.CustomerId == request.CustomerId && s.Parent == collection.Id)
+                .ToListAsync(cancellationToken: cancellationToken);
 
             foreach (var item in items)
             {
-                item.FullPath = $"{request.Slug}/{item.Slug}";
+                item.FullPath = string.IsNullOrEmpty(request.Slug) ? item.Slug : $"{request.Slug}/{item.Slug}";
```
donaldgray, line 147: This is identical to what's used in `UpdateCollection` - can that go in a helper to manage paths?

src/IIIFPresentation/API/Features/Storage/Requests/UpdateCollection.cs
```diff
@@ -78,6 +83,7 @@ public async Task<ModifyEntityResult<FlatCollection>> Handle(UpdateCollection re
         }
 
         return ModifyEntityResult<FlatCollection>.Success(
-            collectionFromDatabase.ToFlatCollection(request.UrlRoots, settings.PageSize, items));
+            collectionFromDatabase.ToFlatCollection(request.UrlRoots, 1, settings.PageSize, total,
```
donaldgray, line 86: What is `1`?

JackLewis-digirati: `1` is the current page - you can't currently set the page on `Update`, so git assumes the first page for `view` - it's definitely not clear though, so will use something other than a magic number

src/IIIFPresentation/API/Features/Storage/Requests/GetCollection.cs
JackLewis-digirati: I think there are issues with running queries in parallel in EF - https://learn.microsoft.com/en-us/ef/core/dbcontext-configuration/#avoiding-dbcontext-threading-issues

donaldgray (approved):

Issue #7, opened 2024-08-30
Update FlatCollection to use the iiif-net package


sarahos:
This ticket will not contain changes for playwright testing and is more of a tech task

FlatCollection is used to show additional properties to a standard IIIF collection, it would make sense to use inheritance from the centralized iiif-net library collections.  This was attempted earlier, but failed due to issues with serialization

see:
https://github.com/digirati-co-uk/iiif-net
https://github.com/dlcs/iiif-presentation/blob/feature/updateCollection/src/IIIFPresentation/Models/API/Collection/FlatCollection.cs

- rename `FlatCollection` to `PresentationCollection`
- investigate using inheritance between `Collection` and `FlatCollection`
- Possible fix: allow custom serializers in the iiif-net library

donaldgray:
Moving to 'done' as no specific tests required for this.

PR #18, opened 2024-09-12
Hierarchical collection to return with AsJson

### LINKED ISSUE #17: Json properties for @type shown as type + say "1" instead of "Collection"
I think this was introduced as part of using the iiif-net library, which uses Newtonsoft, compared to system.text.json, which necessitated adding Newtonsoft to `Program.cs`.

This has meant the project is using Newtonsoft over System.Text.Json

current issues:

- adding both causes Netwonsoft to take precedence
- using purely System.Text.Json causes collections to not be displayed properly in the hierarchical view

JackLewis-digirati:
Partly resolves #17 

This PR moves hierarchical collects to use `.AsJson` found with iiif-net library over relying on internal serialization.

The reason for this change is that `.AsJson` contains a number of converters designed for the iiif spec which are not available through standard serialization.  Additionally, the `Content-Type` header is now correctly set to that required by the iiif spec

This Pr does not change `FlatCollection` as it's currently being modified to work with the iiif-net library in #7

### DIFF
--- a/src/IIIFPresentation/API/Features/Storage/StorageController.cs
+++ b/src/IIIFPresentation/API/Features/Storage/StorageController.cs
@@ -54,7 +54,7 @@ public class StorageController(IOptions<ApiSettings> options, IMediator mediator
 
         if (Request.ShowExtraProperties())
         {
-            return Ok(storageRoot.Collection.ToHierarchicalCollection(GetUrlRoots(), storageRoot.Items));
+            return Ok(storageRoot.Collection.ToFlatCollection(GetUrlRoots(), Settings.PageSize, storageRoot.Items));
         }
 
         return Content(storageRoot.Collection.ToHierarchicalCollection(GetUrlRoots(), storageRoot.Items).AsJson(),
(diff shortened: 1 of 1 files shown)


src/IIIFPresentation/API/Features/Storage/StorageController.cs
```diff
@@ -50,9 +52,13 @@ public async Task<IActionResult> Get(int customerId, string id)
 
         if (storageRoot.Collection == null) return NotFound();
 
-        return Ok(Request.ShowExtraProperties()
-            ? storageRoot.Collection.ToFlatCollection(GetUrlRoots(), Settings.PageSize, storageRoot.Items)
-            : storageRoot.Collection.ToHierarchicalCollection(GetUrlRoots(), storageRoot.Items));
+        if (Request.ShowExtraProperties())
```
donaldgray, line 55: Not directly related to changes in this PR but should this response include `Vary` header for x-iiif-show-extras?

JackLewis-digirati: https://github.com/dlcs/iiif-presentation/issues/19

Issue #39, opened 2024-09-20
`created` and `modified` timestamps on a new storage collection are different


griffri:
The `created` and `modified` timestamps on a newly created storage collection should be the same, but there's actually a very miniscule difference between them. 

Examples from failing tests:

```
Expected: "2024-09-20T10:28:58.060219Z"
Received: "2024-09-20T10:28:58.0602192Z"
```

```
Expected: "2024-09-20T10:30:59.9647261Z"
Received: "2024-09-20T10:30:59.9647265Z"
```

This might be because `DateTime.Now()` is being called separately when setting both fields.

**Acceptance Criteria:**
- Presentation applies the same timestamp to the `created` and `modified` fields of a newly created storage collection.

PR #40, opened 2024-09-23
Ensure matching 'modified' and 'creation' collection values on create

griffri:
Resolves https://github.com/dlcs/iiif-presentation/issues/39

This PR fixes a minor bug that caused a slight difference between the `created` and `modified` timestamps on a newly created collection, caused by calling `DateTime.UtcNow` twice.


JackLewis-digirati:
Is it worth modifying a test to check these values are the same on POST?

JackLewis-digirati (approved):

PR #44, opened 2024-09-23
 Bump iiif-net and use exposed options with iiif-net derrived collections

p-kaczynski:
Essentially we use `JsonSerializerSettings` used by `iiif-net`, as it provides (now) working converters in correct order to handle all the weirdness of the iiif collections. Additionally, as the `iiif-net` package has no idea about a `FlatCollection`, so we provide it in serialization context.

It's not pretty but kinda "reality of working with json dot net"

60 tests passing on my local env.


src/IIIFPresentation/Models/API/Collection/FlatCollection.cs
```diff
@@ -1,48 +1,45 @@
-﻿using IIIF.Presentation.V3.Strings;
+﻿using IIIF.Presentation.V3;
+using IIIF.Presentation.V3.Strings;
 using Newtonsoft.Json;
 
 namespace Models.API.Collection;
 
-public class FlatCollection
+public class FlatCollection : IIIF.Presentation.V3.Collection
```
JackLewis-digirati, line 7: I think this needs to be renamed to `PresentationCollection`

src/IIIFPresentation/Models/API/Collection/FlatCollection.cs
```diff
@@ -1,48 +1,45 @@
...
     public int? ItemsOrder { get; set; }
-    
-    public List<Item>? Items { get; set; }
-    
-    public List<PartOf>? PartOf { get; set; }
-    
+
     public int TotalItems { get; set; }
 
     public View? View { get; set; }
```
JackLewis-digirati, line 34: I think View might be using `JsonPropertyName` instead of `JsonProperty` for some attributes (my mistake when I built this class)

Additionally I think the `Type` in this class may need to be changed away from `PresentationType`

src/IIIFPresentation/Models/API/Collection/FlatCollection.cs
```diff
@@ -1,48 +1,45 @@
...
-    
-    public List<SeeAlso>? SeeAlso { get; set; }
```
JackLewis-digirati, line 35: if `SeeAlso` has been removed, should the class in `Models` be removed as well?

src/IIIFPresentation/Models/API/Collection/FlatCollection.cs
```diff
@@ -1,48 +1,45 @@
...
 
-    public List<string> Behavior { get; set; } = new ();
+    public override string Type { get; } = nameof(FlatCollection);
 
-    public LanguageMap? Label { get; set; }
+    #endregion
+
+    public string? PublicId { get; set; }
+
+    public PresentationType PresentationType { get; set; }
```
JackLewis-digirati, line 24: I don't think this needs to be returned to the user

src/IIIFPresentation/Models/API/Collection/FlatCollection.cs
```diff
@@ -1,48 +1,45 @@
...
+    public new List<string>? Context
+    {
+        get => base.Context as List<string>;
+        set => base.Context = value;
+    }
+
+    #region Overrides of Collection
 
-    public List<string> Behavior { get; set; } = new ();
+    public override string Type { get; } = nameof(FlatCollection);
```
JackLewis-digirati, line 18: I think this should just be `collection`

p-kaczynski: then there is no way for the `ResourceBaseV3Converter` to know which concrete class deriving from collection to use

src/IIIFPresentation/Models/API/Collection/FlatCollection.cs
p-kaczynski: Ah, yes, probably... unless we want to derive from ExternalResource, but I have no idea if there is any point to this

src/IIIFPresentation/Models/API/Collection/FlatCollection.cs
JackLewis-digirati: might need to talk to Tom about it then, as I think the [docs](https://deploy-preview-3--dlcs-docs.netlify.app/api-doc/iiif#public-and-api-storage-roots) says it should be `Collection`

src/IIIFPresentation/Models/API/Collection/FlatCollection.cs
JackLewis-digirati: I think this still needs removing

src/IIIFPresentation/Models/API/Collection/View.cs
```diff
@@ -1,13 +1,13 @@
...
 
 public class View
 {
-    [JsonPropertyName("@id")]
+    [JsonProperty("@id")]
     public required string Id { get; set; }
 
-    [JsonPropertyName("@type")] 
+    [JsonProperty("@type")] 
     public PresentationType Type { get; set; }
```
JackLewis-digirati, line 11: I think this is still returning `"@type": 0`

PR #45, opened 2024-09-23
Allow flat id to be set

### LINKED ISSUE #24: allow flat ID to be set
from Netlify: 

If your public IIIF resources default to the hierarchical scheme, you may not have any need to assign your own flat identifiers to the IIIF resources.

As mentioned there, you should be able to create a storage collection with a specified identifier

PUT `/{customer}/collections/{my-specified-id}`
- payload has "parent" and "slug".
- flat id is as specified in URL,
- parent and slug properties used to set respective properties.
- "parent" and "slug" are required
- "id" in payload ignored

Acceptance Criteria:
- allow PUT to upsert a collection, instead of reject when not found (i.e.: set the id of the storage collection to Bob)
- When PUT is used for insert, then ETag shouldn't be set
- id has 250 character max length
- `A-Za-z0-9:-_` are allowed characters
- `/` is allowed as well and should allow multiple, so needs to avoid interaction with the api

JackLewis-digirati:
Resolves #24

This PR adds the ability for PUT to be used as an upsert operation, instead of just update.  It does this by checking for an ID from the URL in the database, and creating a record if it can't find one

Additionally, there are changes to ETags to support using a cache over a dictionary, and allowing creation on PUT resources


src/IIIFPresentation/API.Tests/Integration/ModifyCollectionTests.cs
```diff
@@ -292,6 +292,81 @@ public async Task UpdateCollection_UpdatesCollection_WhenAllValuesProvided()
         responseCollection.View.Id.Should().Contain("?page=1&pageSize=20");
     }
     
+     [Fact]
```
griffri, line 295: Whitespace behind `[Fact]`

src/IIIFPresentation/API/Attributes/EtagCachingAttribute.cs
```diff
@@ -66,7 +65,7 @@ ResultExecutionDelegate next
 
                 // Remove all unnecessary headers while only keeping the ones that should be included in a `304` response.
                 foreach (var header in response.Headers)
-                    if (!headersToKeepFor304.Contains(header.Key))
+                    if (!HeadersToKeepFor304.Contains(header.Key))
```
griffri, line 68: Should the operations within this `foreach` be enclosed in a bracket, following the style used in Protagonist?

src/IIIFPresentation/API/Attributes/EtagCachingAttribute.cs
```diff
@@ -100,7 +99,7 @@ private static EntityTagHeaderValue GenerateETag(Stream stream, string path)
             new EntityTagHeaderValue('"' + hashString +
                                      '"');
 
-        etagHashes[path] = enityTagHeader.Tag.ToString();
+        eTagManager.UpsertETag(path, enityTagHeader.Tag.ToString());
         return enityTagHeader;
```
griffri, line 103: Typo - should be `entityTagHeader`

src/IIIFPresentation/Core/WriteResult.cs
```diff
@@ -48,5 +48,10 @@ public enum WriteResult
     /// <summary>
     ///     Predefined storage limits exceeded
```
griffri, line 49: The contents of the `<summary>` blocks in this file are indented, but this style doesn't seem to be used elsewhere.

src/IIIFPresentation/API/Attributes/EtagCachingAttribute.cs
```diff
@@ -100,7 +99,7 @@ private static EntityTagHeaderValue GenerateETag(Stream stream, string path)
             new EntityTagHeaderValue('"' + hashString +
```
griffri, line 99: Could this statement be made into one line?
Also - could string interpolation be used to build this string?

griffri (approved):
Looks good - just a few minor things.

PR #47, opened 2024-09-25
Round down the timestamps returned to the whole second for PresentationCollection

### LINKED ISSUE #46: created (or modified) value from POST is off by a few microseconds from what was retrieved via GET
There can be tiny differences in the smallest numbers of the DateTime i.e.: `2024-09-24T09:46:16.892094Z` when using POST and then retrieving via GET.

This breaks the ETag, as they currently require the file to be identical to generate the same ETag

should really be `2024-09-24T09:46:16`

p-kaczynski:
Resolves #46

### DIFF
--- a/src/IIIFPresentation/API.Tests/Integration/GetCollectionTests.cs
+++ b/src/IIIFPresentation/API.Tests/Integration/GetCollectionTests.cs
@@ -1,4 +1,6 @@
-﻿using System.Net;
+﻿#nullable disable
+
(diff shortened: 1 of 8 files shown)


JackLewis-digirati (approved):
LGTM, but do you mind changing the name of this PR as they show up on our release notes, and if you add the text `Resolves #<issue number>` into the description it will link the issue to this PR for record keeping and tracking

PR #50, opened 2024-09-26
Change id's to be generated by sqids and an injected id generator

### LINKED ISSUE #25: use more human readable method of minting identifiers, over GUIDs
Protagonist uses a simpler method of generating ID's

- consider Sqids as it's already a NuGet package - https://sqids.org/dotnet
  - Sqids is encoded from numbers, need to look at uniqueness with these numbers
  - use customer id + some other numbers?
  - make sure on second attempt, id is different
- if this won't work, create a NuGet package based on the class
- collisions possible, so need to check for uniqueness
- unit test for checking that duplicate ID's are rejected
- Leeds will use a custom ID retrieved from their system

Acceptance Criteria:
- (more) human readable ID is generated
- The ID must be unique
- There are multiple possible ID systems, so this should use dependency injection

JackLewis-digirati:
Resolves #25 

this PR creates an id generator that can be injected into the context to create more human readable id's.

### DIFF
--- a/src/IIIFPresentation/API/Features/Storage/Requests/CreateCollection.cs
+++ b/src/IIIFPresentation/API/Features/Storage/Requests/CreateCollection.cs
@@ -38,7 +38,6 @@ public class CreateCollectionHandler(
     private readonly ApiSettings settings = options.Value;
 
     private const int CurrentPage = 1;
-    private const int maxAttempts = 3;
     
     public async Task<ModifyEntityResult<PresentationCollection>> Handle(CreateCollection request, CancellationToken cancellationToken)
     {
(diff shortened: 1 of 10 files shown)


src/IIIFPresentation/API/Features/Storage/Requests/CreateCollection.cs
```diff
@@ -47,11 +51,24 @@ public async Task<ModifyEntityResult<FlatCollection>> Handle(CreateCollection re
             return ModifyEntityResult<FlatCollection>.Failure(
                 $"The parent collection could not be found", WriteResult.Conflict);
         }
+        
+        string id = string.Empty;
```
griffri, line 55: This variable could defined with `var`

griffri (approved):
LGTM, just one minor suggestion

PR #52, opened 2024-09-26
Add vary heade to flat and hierarchical collection GETs

p-kaczynski:
image

Array can be expanded or moved to config if anything like that ever becomes necessary.

Attr can be easily modified to take the `string[] headerNames` in ctor, if we ever need to vary the vary header in various actions 😆 

For now took the simple approach. Well, simple would be to add `Response.Headers.Append` in the actions themselves, but you know, that's awful :D

### DIFF
--- a/src/IIIFPresentation/API/Infrastructure/Filters/VaryHeaderAttribute.cs
+++ b/src/IIIFPresentation/API/Infrastructure/Filters/VaryHeaderAttribute.cs
@@ -1,0 +1,15 @@
+﻿using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.Net.Http.Headers;
+
+namespace API.Infrastructure.Filters;
+
+public class VaryHeaderAttribute : ActionFilterAttribute
+{
+    private static readonly string[] VaryHeaders = ["X-IIIF-CS-Show-Extras", "Authorization"];
+
+    public override void OnActionExecuted(ActionExecutedContext context)
+    {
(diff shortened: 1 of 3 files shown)


src/IIIFPresentation/API/Infrastructure/Filters/VaryHeaderAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace API.Infrastructure.Filters;
+
+public class VaryHeaderAttribute : ActionFilterAttribute
+{
+    private static readonly string[] VaryHeaders = ["X-IIIF-CS-Show-Extras", "Authorization"];
+
+    public override void OnActionExecuted(ActionExecutedContext context)
+    {
+        if (context.Result is not null)
+            context.HttpContext.Response.Headers.Append(HeaderNames.Vary, VaryHeaders);
```
JackLewis-digirati, line 13: As a coding style, I think we wrap single lines like this in `{}`

JackLewis-digirati (approved):
LGTM

src/IIIFPresentation/API/Infrastructure/Filters/VaryHeaderAttribute.cs
p-kaczynski: I think you can copy the style from https://github.com/dlcs/protagonist/blob/main/src/protagonist/.editorconfig - my IDE will then force me to do it :D

PR #67, opened 2024-10-04
IIIF hierarchical post

### LINKED ISSUE #27: Create IIIF collections from hierarchical
Update existing Create endpoint to support creating IIIF Collections

part 1: 

https://deploy-preview-3--dlcs-docs.netlify.app/api-doc/iiif#example-create-a-iiif-collection-within-a-storage-collection

see #28

create IIIF collection which is empty items

gather some IIIF collections without items from the internet, add a parent and slug and see if they can be POSTed 
- id should be ignored/stripped out of payload

Acceptance Criteria:

- POST `/{customer}/{parent-collection}`
- payload has "id", which is used to get the slug value and is required
- parent is derived from parent-collection in URL
- flat id is auto-gen
- "slug" and "parent" are disallowed
- will only accept valid IIIF - no extra storage collection fields (bad request)

JackLewis-digirati:
Resolves #27 

This PR adds the ability to use POST from the hierarchical URL to create IIIF collections

### DIFF
--- a/src/IIIFPresentation/API/Features/Storage/Helpers/ThumbnailX.cs
+++ b/src/IIIFPresentation/API/Features/Storage/Helpers/ThumbnailX.cs
@@ -1,0 +1,33 @@
+﻿using IIIF.Presentation.V3.Content;
+
+namespace API.Features.Storage.Helpers;
+
+public static class ThumbnailX
+{
+    private const int ThumbnailSize = 100;
+    private const int AcceptableDelta = 2;
+    
+    public static string? GetThumbnailPath(this List<Image> thumbnails)
+    {
+        var byWidth = thumbnails.FirstOrDefault(i => i.Width.CloseEnough())?.Id;
+
+        if (byWidth != null)
+        {
+            return byWidth;
+        }
+        
+        var byHeight = thumbnails.FirstOrDefault(i => i.Height.CloseEnough())?.Id;
+
+        return byHeight ?? thumbnails.MinBy(x => x.Width)?.Id;
+    }
+    
+    private static bool CloseEnough(this int? toCheck)
+    {
+        if (toCheck.HasValue)
+        {
+            return Math.Abs(toCheck.Value - ThumbnailSize) <= AcceptableDelta;
+        }
+        
+        return false;
+    }
+}
--- a/src/IIIFPresentation/API/Features/Storage/Helpers/ThumbnailX.cs
+++ b/src/IIIFPresentation/API/Features/Storage/Helpers/ThumbnailX.cs
@@ -1,33 +1,41 @@
-﻿using IIIF.Presentation.V3.Content;
+﻿using IIIF;
+using IIIF.Presentation.V3.Content;
 
 namespace API.Features.Storage.Helpers;
 
 public static class ThumbnailX
 {
     private const int ThumbnailSize = 100;
-    private const int AcceptableDelta = 2;
     
+    /// <summary>
+    /// Finds the thumbnail path of the closest thumbnail to a set size
+    /// </summary>
+    /// <param name="thumbnails">List of thumbnails to check for closest</param>
+    /// <returns>The id of the closest thumbnail</returns>
     public static string? GetThumbnailPath(this List<Image> thumbnails)
     {
-        var byWidth = thumbnails.FirstOrDefault(i => i.Width.CloseEnough())?.Id;
-
-        if (byWidth != null)
-        {
-            return byWidth;
-        }
-        
-        var byHeight = thumbnails.FirstOrDefault(i => i.Height.CloseEnough())?.Id;
-
-        return byHeight ?? thumbnails.MinBy(x => x.Width)?.Id;
+        return thumbnails.SizeClosestTo(ThumbnailSize).Id;
     }
     
-    private static bool CloseEnough(this int? toCheck)
+    /// <summary>
+    /// From provided sizes, return the Size that has MaxDimension closest to specified targetSize
+    ///
+    /// e.g. [[100, 200], [250, 500] [500, 1000]], targetSize = 800 would return [500, 1000]
+    /// </summary>
+    /// <param name="sizes">List of sizes to query</param>
+    /// <param name="targetSize">Ideal MaxDimension to find</param>
+    /// <returns><see cref="Size"/> closes to specified value</returns>
+    private static Image SizeClosestTo(this IEnumerable<Image> sizes, int targetSize)
     {
-        if (toCheck.HasValue)
-        {
-            return Math.Abs(toCheck.Value - ThumbnailSize) <= AcceptableDelta;
-        }
-        
-        return false;
+        var closestSize = sizes
+            .OrderBy(MaxDimension)
+            .Aggregate((x, y) =>
+                Math.Abs(x.MaxDimension() - targetSize) < Math.Abs(y.MaxDimension() - targetSize) ? x : y);
+        return closestSize;
+    }
+
+    private static int MaxDimension(this Image s)
+    {
+        return s.Width > s.Height ? s.Width ?? 0 : s.Height ?? 0;
     }
 }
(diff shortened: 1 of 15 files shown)


griffri (approved):

src/IIIFPresentation/API/Features/Storage/Helpers/ThumbnailX.cs
```diff
@@ -0,0 +1,33 @@
+﻿using IIIF.Presentation.V3.Content;
+
+namespace API.Features.Storage.Helpers;
+
+public static class ThumbnailX
+{
+    private const int ThumbnailSize = 100;
+    private const int AcceptableDelta = 2;
```
donaldgray, line 8: Does this need to be restricted to a set size, or can it looks for size closest to? (which is how it's done in Protagonist https://github.com/dlcs/protagonist/blob/main/src/protagonist/DLCS.Model/IIIF/IIIFX.cs#L38)

Is it better to find _a_ thumb, rather than the perfect thumb?

src/IIIFPresentation/API/Features/Storage/Helpers/ThumbnailX.cs
```diff
@@ -0,0 +1,33 @@
...
+    
+    public static string? GetThumbnailPath(this List<Image> thumbnails)
```
donaldgray, line 10: Worth adding XML comments, or expanding method name to make it clearer that this will get thumbnail closest to X +/- threshold? (may not be required depending on whether changes made for previous comment)

src/IIIFPresentation/API/Features/Storage/Requests/PostHierarchicalCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+    : IRequestHandler<PostHierarchicalCollection, ModifyEntityResult<Collection, ModifyCollectionType>>
+{
+    private readonly ApiSettings settings = options.Value;
+    
+    public async Task<ModifyEntityResult<Collection, ModifyCollectionType>> Handle(PostHierarchicalCollection request,
+        CancellationToken cancellationToken)
+    {
+        var splitSlug = request.Slug.Split('/');
+
+        var parentSlug = String.Join(String.Empty, splitSlug.Take(..^1));
```
donaldgray, line 54: Why use `String` here, rather than `string` alias? Don't think anywhere else in codebase does this?

src/IIIFPresentation/API/Features/Storage/Requests/PostHierarchicalCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+            Modified = dateCreated,
+            CreatedBy = Authorizer.GetUser(),
+            CustomerId = request.CustomerId,
+            IsPublic = collectionFromBody.Behavior != null && collectionFromBody.Behavior.IsPublic(),
+            IsStorageCollection = false,
+            Label = collectionFromBody.Label,
+            Thumbnail = thumbnails!?.GetThumbnailPath()
+        };
+        
+        await using var transaction =
```
donaldgray, line 102: Is the transcation required here? `TrySaveCollection` calls `SaveChangesAsync` which will [be in a transaction](https://learn.microsoft.com/en-us/ef/core/saving/transactions#default-transaction-behavior) without you needing to do it externally. If `SaveChanges` fails, `saveErrors` will be returned. If `SaveChanges` works then the transaction will always be committed as `WriteToBucket` never throws

src/IIIFPresentation/API/Features/Storage/Requests/PostHierarchicalCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        string id;
+        
+        try
+        {
+            id = await dbContext.Collections.GenerateUniqueIdAsync(request.CustomerId, idGenerator, cancellationToken);
+        }
+        catch (ConstraintException ex)
+        {
+            logger.LogError(ex, "An exception occured while generating a unique id");
```
donaldgray, line 68: Do these logs contain a correlation_id or anything to help know more context around the request?

src/IIIFPresentation/API/Features/Storage/Requests/PostHierarchicalCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+            return ErrorHelper.CannotGenerateUniqueId<Collection>();
+        }
+
+        Collection collectionFromBody;
+        try
+        {
+            collectionFromBody = request.RawRequestBody.FromJson<Collection>();
+            collectionFromBody.Id = $"{request.UrlRoots.BaseUrl}/{request.CustomerId}/{request.Slug}";
```
donaldgray, line 76: Should this be in the class that builds paths/URLs?

src/IIIFPresentation/API/Features/Storage/Requests/PostHierarchicalCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+    public int CustomerId { get; } = customerId;
+
+    public string Slug { get; } = slug;
+    
+    public UrlRoots UrlRoots { get; } = urlRoots;
+    
+    public string RawRequestBody { get; } = rawRequestBody;
+}
+
+public class PostHierarchicalCollectionHandler(
```
donaldgray, line 39: General comment/personal opinion: this method is doing a lot and there's a lot of vars that are used for single purposes. Could it be split up to make it easier to read at a glance?

```cs
var parentCollection = GetParentCollection(request);
if (parentCollection == null) return ErrorHelper.NullParentResponse<Collection>();

var id = GetUniqueId();
if (id == null) return ErrorHelper.CannotGenerateUniqueId<Collection>();

var collection = BuildCollection(request, id, parentCollection);
if (collection == null) return ErrorHelper.CannotValidateIIIF<Collection>();

await SaveStuff()...
```

src/IIIFPresentation/API/Features/Storage/StorageController.cs
```diff
@@ -42,6 +42,20 @@ public async Task<IActionResult> GetHierarchicalCollection(int customerId, strin
...
+    [HttpPost("{*slug}")]
+    [ETagCaching]
+    public async Task<IActionResult> PostHierarchicalCollection(int customerId, string slug)
+    {
+        if (!Request.ShowExtraProperties())
+        {
+            return this.PresentationProblem(statusCode: (int)HttpStatusCode.Forbidden);
+        }
+        
+        var rawRequestBody = await new StreamReader(Request.Body).ReadToEndAsync();
```
donaldgray, line 55: Does `StreamReader` need to be disposed of? Or would that close `Request.Body`? and if it did is that bad?

src/IIIFPresentation/API/Features/Storage/Helpers/ThumbnailX.cs
JackLewis-digirati: swapped to use `ClosestTo`

src/IIIFPresentation/API/Features/Storage/Requests/PostHierarchicalCollection.cs
JackLewis-digirati: it currently just contains the standard exception information - I'll look to add the `CorrelationIdHeaderEnricher` from protagonist

donaldgray: That can be done separately I think to save adding more to this. It's definitely useful to add as we can track requests across iiif-p and protagonist-api/engine etc. Create a ticket and we can pick it up

donaldgray: https://github.com/dlcs/iiif-presentation/issues/70

donaldgray (approved):

PR #68, opened 2024-10-07
Thumbnail can't handle IIIF style thumbnail on POST

### LINKED ISSUE #57: thumbnail expects List<ExternalResource> when trying to save a IIIF collection
When saving a storage collection into the database, the `thumbnail` expects a single string that contains a location of the tumbnail.  However, when saving a IIIF collection, `thumbnail` expects a `List<ExternalResource>`.  This causes the application to break when a `thumbnail` is attached when trying to create a IIIF collection due to the conflicting definition

Additionally, `tags` will get stripped out in the conversion if added

JackLewis-digirati:
Resolves #57

This PR allows thumbnails to be correctly converted between `ExternalResource` and `string` when using POST on a IIIF collection, via a `collections`

### DIFF
--- a/src/IIIFPresentation/API/Features/Storage/Requests/CreateCollection.cs
+++ b/src/IIIFPresentation/API/Features/Storage/Requests/CreateCollection.cs
@@ -103,9 +103,6 @@ public class CreateCollectionHandler(
             collection.Thumbnail = request.Collection.PresentationThumbnail;
         }
         
-        await using var transaction = 
-            await dbContext.Database.BeginTransactionAsync(cancellationToken);
-        
         dbContext.Collections.Add(collection);
 
         var saveErrors =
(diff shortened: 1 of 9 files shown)


src/IIIFPresentation/API/Features/Storage/Requests/CreateCollection.cs
```diff
@@ -80,13 +80,28 @@ public async Task<ModifyEntityResult<PresentationCollection, ModifyCollectionTyp
...
+
+        if (!request.Collection.Behavior.IsStorageCollection())
+        {
+            try
+            {
+                convertedIIIFCollection = ConvertToIIIFCollection(request, collection);
+            }
+            catch (Exception ex)
+            {
+                logger.LogError(ex, "Error attempting to validate collection is IIIF");
```
griffri, line 97: This message feels a little unclear to me - would something like "An error occurred while attempting to validate the collection as IIIF" be easier to read?

griffri (approved):

PR #93, opened 2024-10-13
Prototype/canvas mapping

tomcrane:
This is a separate solution to explore how an incoming manifest is mapped to the canvas_paintings table.

### DIFF
--- a/src/CanvasPaintings/Mapper/Identity/Identifiable.cs
+++ b/src/CanvasPaintings/Mapper/Identity/Identifiable.cs
@@ -1,0 +1,33 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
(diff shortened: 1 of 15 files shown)


src/CanvasPaintings/Mapper/Identity/Identifiable.cs
```diff
@@ -0,0 +1,33 @@
...
+                if (letterFirst && i == 0)
+                {
+                    sb.Append(Letters[random.Next(Letters.Length)]);
+                }
+                else
+                {
+                    sb.Append(All[random.Next(All.Length)]);
+                }
+            }
+            return sb.ToString();
```
stephenwf, line 30: I wonder if we could hash the canvas ID to get an identifier? (Might simplify PUT/update requests)

PR #94, opened 2024-10-14
API gains delegated auth

### LINKED ISSUE #89: Add authentication to the API
For MVP in **iiif-presentation** application this can just be a delegated auth handler - 

 - [ ] detect basic HTTP auth credentials in request
 - [ ] use these to make HTTP request to "legacy" DLCS API for Customer
 - [ ] if 200 OK, cache the credentials in memory cache in iiif-presentation app
 - [ ] on subsequent requests, authenticate the users creds against the memory cache
 
(THIS IS NOT PRODUCTION!)

donaldgray:
Resolves #89. Adds auth to API. This currently takes any `Authorization` header passed and proxies it to a downstream DLCS instance, it will attempt to `GET /customer/{customerId}` and if we receive a 200 request is allowed. Result of call is cached in-memory for a short period.

Implementation notes:
* Add `DelegatedAuthHandler` to API, this allows for use of standard `[Authorize]` attributes. Note that this checks authorization only, checks for `x-iiif-cs-show-extras` header is handles elsewhere. Logic for actual auth logic is is delegated to `IAuthorizer` impl.
* `IAuthorizer` has a single implementation, `DelegatedAuthenticator`, which makes request to DLCS and caches results.
  * _Note that there's no locking around this as we don't expect high call volumes_

Other changes:
* Make `x-iiif-cs-show-extras` header case insensitive 
* Refactored `PostHierarchicalCollection` to make quick/cheap serialiser check first before doing any DB work.
* Slight refactoring of `PresentationAppFactory` - change in dotnet 8 meant values in `appsettings.testing.json` weren't available until the host had been built.
* Sample SQL script for creating root collection

### DIFF
--- a/scripts/create_root.sql
+++ b/scripts/create_root.sql
@@ -1,0 +1,4 @@
+INSERT INTO public.collections 
+    (id, slug, use_path, label, created, modified, created_by, modified_by, is_storage_collection, is_public, customer_id) 
+VALUES
+    ('root', '', false, '{"en": ["(repository root)"]}', now(), now(), 'Admin', 'Admin', true, true, 1);
--- a/src/IIIFPresentation/API/Auth/DelegatedAuthHandler.cs
+++ b/src/IIIFPresentation/API/Auth/DelegatedAuthHandler.cs
@@ -6,10 +6,8 @@ using Microsoft.Extensions.Options;
 namespace API.Auth;
 
 /// <summary>
-/// AuthenticationHandler that hands off calls to DLCS for carrying out authentication. Any auth header provided
-/// here will be proxied to DLCS API - a 200 response = auth successful.
+/// AuthenticationHandler that hands off calls to implementation of <see cref="IAuthenticator"/> for auth logic
 /// </summary>
-/// <remarks>This is temporary and will be replaced in the future by an implementation that has auth logic</remarks>
 public class DelegatedAuthHandler(
     IAuthenticator delegatedAuthenticator,
     IOptionsMonitor<DelegatedAuthenticationOptions> options,
--- a/src/IIIFPresentation/API/Auth/DelegatedAuthHandler.cs
+++ b/src/IIIFPresentation/API/Auth/DelegatedAuthHandler.cs
@@ -1,0 +1,64 @@
+﻿using System.Security.Claims;
+using System.Text.Encodings.Web;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.Options;
+
+namespace API.Auth;
+
+/// <summary>
+/// AuthenticationHandler that hands off calls to DLCS for carrying out authentication. Any auth header provided
+/// here will be proxied to DLCS API - a 200 response = auth successful.
+/// </summary>
+/// <remarks>This is temporary and will be replaced in the future by an implementation that has auth logic</remarks>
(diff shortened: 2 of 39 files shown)


scripts/create_root.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO public.collections 
+    (id, slug, use_path, label, created, modified, created_by, modified_by, is_storage_collection, is_public, customer_id) 
+VALUES
+    ('root', '', false, '{"en": ["(repository root)"]}', now(), now(), 'Admin', 'Admin', true, true, 1);
```
JackLewis-digirati, line 4: s it worth adding a note that customer id might need to be changed based on the delegated auth user?

src/IIIFPresentation/API/Auth/DelegatedAuthHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+
+/// <summary>
+/// Options for use with <see cref="DelegatedAuthHandler"/>
+/// </summary>
+public class DelegatedAuthenticationOptions : AuthenticationSchemeOptions
+{
+    /// <summary>
+    /// Get or set the Realm for use in auth challenges.
+    /// </summary>
+    public string Realm { get; set; }
```
JackLewis-digirati, line 53: is it worth fixing this GitHub comment?

src/IIIFPresentation/API/Settings/DlcsSettings.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace API.Settings;
+
+public class DlcsSettings
+{
+    public const string SettingsName = "DLCS";
+    
+    /// <summary>
+    /// URL root of DLCS API 
+    /// </summary>
+    public Uri ApiUri { get; set; }
```
JackLewis-digirati, line 10: is it worth fixing this GitHub comment?

JackLewis-digirati (approved):

src/IIIFPresentation/API/Auth/DelegatedAuthenticator.cs
```diff
@@ -0,0 +1,88 @@
...
+        appCache.Add(cacheKey, list, cacheSettings.CurrentValue.GetMemoryCacheOptions());
+        return true;
+    }
+
+    private async Task<bool> IsValidUserDlcs(AuthenticationHeaderValue authenticationHeaderValue, string customerId)
+    {
+        // Parse the CustomerId out of this.
+        var delegatePath = $"/customers/{customerId}";
+
+        // make a request to DLCS and verify the result received - if it's 200 we're good
```
griffri, line 74: Would it be worth updating this comment to have a capital latter at the start?

griffri (approved):

src/IIIFPresentation/API/Auth/DelegatedAuthHandler.cs
donaldgray: Marked as `= null!` to fix warning. This is passed to `AuthenticationHandler<T>`, which has generic constraint of `new()` so can't use `required`.

src/IIIFPresentation/API/Settings/DlcsSettings.cs
donaldgray: Made `required`

PR #96, opened 2024-10-15
Create and implement new hierarchy table

### LINKED ISSUE #59: Move hierarchy to a separate table
Maybe consider this before starting on manifests and refactor.
It's simpler for one table when it's just collections (both storage and IIIF colls are the same table), but once we have a hierarchy of mixed colls and manifests, this might be better:

```
create table if not exists public.hierarchy
(
    id                    text,    -- pk in collections OR manifests table
    type                  int,     -- storage coll=0, iiif coll=1, manifest=2
    slug                  text,
    parent                text,
    items_order           int,
    public                boolean,
    canonical             boolean
);
```

slug, parent and items_order would be dropped from the collections table and the manifests table

---

I think that only manifests and collections managed by the platform appear in the manifests and collections tables.
A IIIF Collection we store could reference external manifests and/or collections in the JSON we save for it and that never appears in any of these tables.

For those we do manage, that table is doing multiple jobs:
- resolving incoming hierarchical paths to specific resources (hierarchical query to determine which IIIF thing to load off disk, decorate and serve)
- generating paths to emit in generated id properties we decorate our outgoing JSON with
driving the explorer-style view of your IIIF repository in the portal (which is driven by the API providing navigable collections) - this only shows managed iiif resources; you can only tell that your collection's items property includes a BL manifest by looking that the JSON
- recording membership information, so that you can include a particular managed manifest in multiple places, and see that reflected in the portal UI

I didn't have that canonical property in before.
In this table, id is not always unique.
But for any id, there must be one and only one row for which canonical=true.
Most of the time that means one occurrence of id which therefore must have canonical=true
Where id appears multiple times, the rows that are canonical=false are like symlinks. They appear in the tree as you browse so you can see the membership relationship, but if you navigated into it you'd get to its canonical path.

And we could also use this information to redirect HTTP requests from "symlink" to canonical URLs - although really the non-canonical path shouldn't ever be published anywhere in a IIIF resource

JackLewis-digirati:
Resolves #59 

This PR creates the `hierarchy` table and updates the `collections` table to use it for hierarchical operations.  Additionally, a placeholder `manifests` table has been introduced to make sure that the `hierarchy` table can work with both `collections` and `manifests`.

Additionally, there are multiple updates to helper methods in the API to take advantage of this new table, while allowing all current functionality to work the same as previously.


src/IIIFPresentation/Repository/PresentationContext.cs
```diff
@@ -30,9 +35,43 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+            entity.HasKey(e => new {e.Id, e.CustomerId});
+            
+            entity.HasMany(e => e.Hierarchy)
+                .WithOne(e => e.Manifest)
+                .HasForeignKey(e => new { e.ManifestId, e.CustomerId })
+                .HasPrincipalKey(e => new { e.Id, e.CustomerId })
+                .OnDelete(DeleteBehavior.Cascade);
+        });
+        
+        modelBuilder.Entity<Hierarchy>(entity =>
```
JackLewis-digirati, line 60: currently the API assumes that there will always be a canonical path on retrieval, but this isn't enforced by the database - I did leave this out as it might complicate updates, but is this worth adding?

src/IIIFPresentation/API.Tests/Helpers/CollectionHelperXTests.cs
```diff
@@ -90,7 +108,13 @@ public void GenerateFlatCollectionViewId_CreatesViewId()
         var collection = new Collection()
         {
             Id = "test",
-            Slug = "slug"
+            Hierarchy =
+            [
+                new Hierarchy()
+                {
+                    Slug = "slug"
```
griffri, line 115: Would it be worth turning "slug" into a const somewhere, like `defaultSlugName`?

src/IIIFPresentation/API/Features/Storage/Requests/CreateCollection.cs
```diff
@@ -75,20 +76,50 @@ public async Task<ModifyEntityResult<PresentationCollection, ModifyCollectionTyp
...
+
+        if (!request.Collection.Behavior.IsStorageCollection())
+        {
+            try
+            {
+                convertedIIIFCollection = ConvertToIIIFCollection(request, collection);
+            }
+            catch (Exception ex)
+            {
+                logger.LogError(ex, "Error attempting to validate collection is IIIF");
```
griffri, line 109: Similar to a comment I left on #68:
> This message feels a little unclear to me - would something like "An error occurred while attempting to validate the collection as IIIF" be easier to read?

src/IIIFPresentation/Models/Database/General/Hierarchy.cs
```diff
@@ -0,0 +1,59 @@
...
+    
+    /// <summary>
+    /// Used to determine the order of the item when viewed in a collection
+    /// </summary>
+    public int? ItemsOrder { get; set; }
+    
+    /// <summary>
+    /// Whether the item is publicly available or not
+    /// </summary>
+    public bool Public { get; set; }
```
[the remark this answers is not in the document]
JackLewis-digirati: decision is to remove from `hierarchy` table

src/IIIFPresentation/Repository/PresentationContext.cs
JackLewis-digirati: decision is to default `canonical` to `true` for now

src/IIIFPresentation/API/Features/Storage/Requests/UpsertCollection.cs
```diff
@@ -77,14 +81,24 @@ public async Task<ModifyEntityResult<PresentationCollection, ModifyCollectionTyp
...
                 Label = request.Collection.Label,
-                Parent = parentCollection.Id,
-                Slug = request.Collection.Slug,
-                Thumbnail = request.Collection.Thumbnail,
+                Thumbnail = request.Collection.PresentationThumbnail,
                 Tags = request.Collection.Tags,
-                ItemsOrder = request.Collection.ItemsOrder
+            };
+            
+            hierarchy = new Hierarchy
```
donaldgray, line 88: fwiw, I think you should be able to add this hierarchy to the `databaseCollection.Hierarchies` collection and it should get `CollectionId` and `CustomerId` assigned automatically, granted not much of a saving as it's only a couple of fields

src/IIIFPresentation/API/Features/Storage/Requests/UpsertCollection.cs
```diff
@@ -132,21 +152,19 @@ await dbContext.TrySaveCollection<PresentationCollection>(request.CustomerId, lo
         {
             return saveErrors;
         }
-
-        var total = await dbContext.Collections.CountAsync(
+            
+        var total = await dbContext.Hierarchy.CountAsync(
```
donaldgray, line 156: Could we make the round trip to get total optional if we get a full `PageSize` back? If we ask for `PageSize` 20 and get 3 records we know the total is 3.

```cs
var items = dbContext
  .RetrieveCollectionItems(request.CustomerId, databaseCollection.Id)
  .Take(settings.PageSize);
if (items.Count < settings.PageSize)
{
  // there can't be more as we've asked for PageSize and got less 
  total = items.Count;
}
else
{
  // if we get PageSize back then there may be more in db
  total = await dbContext.Hierarchy.CountAsync(....
}
```

src/IIIFPresentation/API/Features/Storage/StorageController.cs
```diff
@@ -72,14 +72,14 @@ public async Task<IActionResult> Get(int customerId, string id, int? page = 1, i
         var storageRoot =
             await Mediator.Send(new GetCollection(customerId, id, page.Value, pageSize.Value, orderByField, descending));
 
-        if (storageRoot.Collection == null) return this.PresentationNotFound();
+        if (storageRoot.Collection == null || storageRoot.Hierarchy == null) return this.PresentationNotFound();
```
donaldgray, line 75: Would `storageRoot.Hierarchy == null` be possible?

src/IIIFPresentation/Models/API/Collection/Upsert/UpsertFlatCollection.cs
```diff
@@ -14,7 +16,19 @@ public class UpsertFlatCollection
     
     public string? Tags { get; set; }
     
-    public string? Thumbnail { get; set; }
+    public string? PresentationThumbnail { get; set; }
+    
+    [JsonProperty("thumbnail")]
+    public object? ThumbnailFromRequest
```
donaldgray, line 22: Was this added as part of this PR? I thought this was added in another one?

src/IIIFPresentation/Models/Database/General/Hierarchy.cs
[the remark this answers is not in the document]
donaldgray: Summary of what was discussed should we need to look back: for now the `Collection` will determine whether it's public or not. In the future we can add complexity around a `Collection` being publicly available on 1 path but not another.

src/IIIFPresentation/Repository/PresentationContext.cs
donaldgray: For future reference - ticket created to handle non-canonical paths. This isn't required initially.

src/IIIFPresentation/API/Features/Storage/Requests/CreateCollection.cs
```diff
@@ -138,9 +144,24 @@ await bucketWriter.WriteToBucket(
...
 
-    private static string GenerateCollectionFromRawRequest(string rawRequestBody)
+    private static string ConvertToIIIFCollection(CreateCollection request, Collection collection)
     {
-        var collection = rawRequestBody.FromJson<IIIF.Presentation.V3.Collection>();
-        return collection.ToJson();
+        var collectionAsIIIF = request.RawRequestBody.FromJson<IIIF.Presentation.V3.Collection>();
+        var convertedIIIFCollection = collectionAsIIIF.AsJson();
+        var thumbnails = collectionAsIIIF.Thumbnail?.Select(c => c as Image).ToList(); 
+        collection.Thumbnail = thumbnails!?.GetThumbnailPath();
```
donaldgray, line 152: Does the warning from Github need to be resolved?

> Argument of type 'List<Image?>' cannot be used for parameter 'thumbnails' of type 'List<Image>' in 'string? ThumbnailX.GetThumbnailPath(List<Image> thumbnails)' due to differences in the nullability of reference types

src/IIIFPresentation/API/Features/Storage/Requests/CreateCollection.cs
```diff
@@ -75,20 +76,50 @@ public async Task<ModifyEntityResult<PresentationCollection, ModifyCollectionTyp
...
+            CustomerId = request.CustomerId,
+            Canonical = true,
+            ItemsOrder = request.Collection.ItemsOrder,
+            Parent = request.Collection.Parent,
+            Public = request.Collection.Behavior.IsPublic()
         };
         
+        string? convertedIIIFCollection = null;
+
+        if (!request.Collection.Behavior.IsStorageCollection())
```
donaldgray, line 101: `request.Collection.Behavior.IsStorageCollection()` isn't particularly expensive but it's called 4 times in this method. Can it be simplified - either the result stored or `collection.IsStorageCollection` used as that's set first?

src/IIIFPresentation/API/Features/Storage/Requests/GetCollection.cs
```diff
@@ -33,33 +34,37 @@ public async Task<CollectionWithItems> Handle(GetCollection request,
...
+                cancellationToken);
+            
+            total = await dbContext.Hierarchy.CountAsync(
                 c => c.CustomerId == request.CustomerId && c.Parent == collection.Id,
                 cancellationToken: cancellationToken);
-            items = await dbContext.Collections.AsNoTracking()
-                .Where(c => c.CustomerId == request.CustomerId && c.Parent == collection.Id)
+            items = await dbContext.RetrieveCollectionItems(request.CustomerId, collection.Id)
                 .AsOrderedCollectionQuery(request.OrderBy, request.Descending)
+                .Include(c => c.Hierarchy)
```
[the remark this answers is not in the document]
JackLewis-digirati: it gets added to the SQL, but it's essentially ignored

src/IIIFPresentation/API/Features/Storage/Requests/GetCollection.cs
```diff
@@ -33,33 +34,37 @@ public async Task<CollectionWithItems> Handle(GetCollection request,
...
                 .Skip((request.Page - 1) * request.PageSize)
                 .Take(request.PageSize)
                 .ToListAsync(cancellationToken: cancellationToken);
             
             foreach (var item in items)
             { 
-                item.FullPath = collection.GenerateFullPath(item.Slug);
+                item.FullPath = hierarchy.GenerateFullPath(item.Hierarchy!.Single(h => h.Canonical).Slug);
```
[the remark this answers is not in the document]
JackLewis-digirati: I did have this at one point with 
```
public static Hierarchy RetrieveCanonical(this IEnumerable<collection> collection)
{
   return collection.Hierarchy!.Single(h => h.Canonical);
}
```

But this breaks whenever it attempts to retrieve items from the database (with cannot convert to SQL) - this is most often where it was used, so it felt slightly dangerous to have a method that only works sometimes - there might be another way to do this though?

src/IIIFPresentation/Models/API/Collection/Upsert/UpsertFlatCollection.cs
JackLewis-digirati: yeah, it was added in #68, which I (wrongly) assumed would be merged by the time we got to this one, so pulled the branch off that - might need to reverse these changes, or merge #68

PR #100, opened 2024-10-16
Process JSON read from S3

p-kaczynski:
- **Added StreamingJsonProcessor and hooked to S3 JSON read**
- **Update StreamingJsonProcessor.cs**
- **Split the StreamingJsonProcessor into general processor and pluggable implementations**
- **Add tests for S3StoredJsonProcessor**

---

Note: I have very strong suspicion that the actual "thing" that is being changed by the processor is probably not exactly what we want, but primarily focused on the speed and ability to do /any/ processing.

Hence I'll be very gracious, if I'm indeed wrong as to what should happen to JSON that we read from S3, if you could e.g. suggest a test that will verify it is what it should be, and I can amend the code :D

### DIFF
--- a/src/IIIFPresentation/API/Converters/Streaming/StreamingJsonProcessor.cs
+++ b/src/IIIFPresentation/API/Converters/Streaming/StreamingJsonProcessor.cs
@@ -20,10 +20,11 @@ public static class StreamingJsonProcessor
     {
         // Initial buffer size - will auto expand if token/whitespace sequence is bigger than that
         const int bufferSize = 1024;
-        var buffer = new byte[bufferSize];
+        var initialSize = inputLength.HasValue ? Math.Min(inputLength.Value, bufferSize) : bufferSize;
+        var buffer = new byte[initialSize];
 
         // First read - might be the only, might be nothing (if so, short circuit return)
-        var bytesRead = input.Read(buffer, 0, bufferSize);
+        var bytesRead = input.Read(buffer, 0, buffer.Length);
         if (bytesRead == 0) return;
 
         // Note using - without it the data will be lost due to lack of dispose call
--- a/src/IIIFPresentation/API/Converters/Streaming/StreamingJsonProcessor.cs
+++ b/src/IIIFPresentation/API/Converters/Streaming/StreamingJsonProcessor.cs
@@ -1,9 +1,8 @@
-﻿using System.Diagnostics;
-using System.Text.Json;
+﻿using System.Text.Json;
 
-namespace API.Converters;
+namespace API.Converters.Streaming;
 
-public class StreamingJsonProcessor
+public static class StreamingJsonProcessor
 {
     /// <summary>
     ///     Reads <paramref name="input" /> UTF-8 JSON stream, token by token, and writes it to the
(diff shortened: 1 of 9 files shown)


src/IIIFPresentation/API/Converters/Streaming/StreamingJsonProcessor.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// <param name="output">Processed UTF-8 JSON target</param>
+    /// <param name="inputLength">If known, helps to read the stream correctly</param>
+    /// <param name="implementation">Handles actual </param>
+    /// <remarks>
+    ///     In C#13 it can be made async, but currently ref structs don't work with async/await.
+    /// </remarks>
+    public static void ProcessJson(Stream input, Stream output, long? inputLength, IProcessJson implementation)
+    {
+        // Initial buffer size - will auto expand if token/whitespace sequence is bigger than that
+        const int bufferSize = 1024;
```
garyttierney, line 22: If you can use Newtonsoft.Json instead, the async version is a lot simpler:

```cs
interface JsonProcessor {
    bool PathMatches(string path);

    object? Process(object? value);
}

async void StreamingJsonMap(JsonTextReader reader, JsonTextWriter writer, JsonProcessor processor) {
    while (await reader.ReadAsync()) {
        if (processor.PathMatches(reader.Path) && reader.TokenType != JsonToken.PropertyName) {
            await writer.WriteTokenAsync(reader.TokenType, processor.Process(reader.Value));
        }
        else {
            await writer.WriteTokenAsync(reader.TokenType, reader.Value);
        }
    }
}
```

donaldgray (commented):
I think this is fine for now, all we need to rewrite is the top level `id` initially from one form to the other. 

Moving forward we might need to rewrite a few other values, like [`"partOf"`](https://iiif.io/api/presentation/3.0/#partof) ids for Manifests if it belongs to a collection - in this instance we might need to know the context of the current token (ie this `id` is child token of `partOf`) - would that just be an expansion of `S3ProcessorCustomState`?: 

```json
{
  "id": "flat-id-will-need-updated",
  "type": "Manifest",
  "partOf": [{
    "id": "flat-id-of-parent-collection-will-need-updated-too",
    "type": "Collection"
  }]
}
```

[`"items"`](https://iiif.io/api/presentation/3.0/#items) property is also a place where `id` would be rewritten but, again, tracking parent should be fine.

src/IIIFPresentation/API/Converters/Streaming/StreamingJsonProcessor.cs
p-kaczynski: It is simpler, but I ran some benchmarks and it seems the Newtonsoft version takes 5x the time STJ does. As performance was stated to be key factor, the "less simple but much faster" seems to be preferable.

p-kaczynski:
> I think this is fine for now, all we need to rewrite is the top level `id` initially from one form to the other.

Is my assumption that we "replace" the top-level id with the provided slug right? Or do we need the stored id to "take part" in the conversion? :D

> Moving forward we might need to rewrite a few other values, like [`"partOf"`](https://iiif.io/api/presentation/3.0/#partof) ids for Manifests if it belongs to a collection - in this instance we might need to know the context of the current token (ie this `id` is child token of `partOf`) - would that just be an expansion of `S3ProcessorCustomState`?:

yes, the idea is that we can expand the `S3ProcessorCustomState` to keep track of whatever we need (but until we need it, it's trying to be relatively light). Naively we can just keep track of StartObject and EndObject, which allows us to keep e.g. a Stack<string> that will tell us the path (or do whatever else).

Also in case we want to do different processing elsewhere this provides the "scaffold" of reading/writing tokens, and we can "plug in" however simple or complex operations to it.

donaldgray:
@p-kaczynski are you able to fix the broken test for this?

p-kaczynski:
Tests now clear

donaldgray (approved):

PR #101, opened 2024-10-18
Update SQL script to use the new hierarchy table

JackLewis-digirati:
This PR is a simple fix to update the `create_root` SQL script to use the new hierarchy table


donaldgray (approved):
LGTM - although I think `hierarchy.id` should be omitted as it's auto-incrementing and will make creating further root values simpler.

Issue #66, opened 2024-10-04
URL query parameters don't carry over on hierarchal -> presentation redirect


griffri:
When a request for a storage collection via its hierarchal URL is redirected to its presentation form (e.g, when valid `Authorization` and and `X-IIIF-CS-Show-Extras` headers have been supplied), the query parameters specified in the URL aren't used.
```
...this will return the presentation form of root with the specified page params
/1/collections/root?page=2&pageSize=2

...this returns it in the same form, but page + pageSize won't be applied to the view
/1?page=2&pageSize=2
```
This causes the returned view to be fixed at the default values.

### Acceptance Criteria
- Query parameters appended to a hierarchal URL are used for paginating the presentation form if redirected.

PR #105, opened 2024-10-21
Fix #95, don't redirect to non-public collections

### LINKED ISSUE #95: Do not redirect to non-public collections
When I request a collection on it's flat URI it will return a 303 / SeeAlso to the hierarchical view if I don't provide auth and `iiif-cs-*` header. However, if the collection that is _not_ public then this shouldn't return a 303 as it results in a 404 when followed.

p-kaczynski:
After #104


JackLewis-digirati (approved):
LGTM, but I think this might need rebasing of `main` as I can see the changes from #66

PR #116, opened 2024-10-22
Create stub manifest

### LINKED ISSUE #61: Create stub Manifest with POST to (flat) `/manifests` endpoint; no `items`
Acceptance criteria

- [ ] A Manifest can be created as a child of a storage collection by POSTing the Manifest JSON to `/<cust>/manifests`, including `parent` and `slug` properties.
~A Manifest can be created as a child of a IIIF Collection~ (stretch)
- [ ] The JSON is stored in S3
~Public can see the empty Manifest at the public (hierarchical) URL~ - removed as covered in #85 
- [ ] Manifest appears in parent Storage Collection items automatically
- [ ] You can't create a Manifest with the same slug and parent as another Collection or Manifest (see #59)

donaldgray:
Resolves #61

New API endpoint, `POST /manifests` where a stub manifest can be created (ie no special handling of `"items"` - this will come later). Manifest can contain any IIIF, which is persisted to S3 - nothing special is stored in DB other than it's place in the hierarchy. 

Manifest can _only_ be created in a storage-collection - attempt to create elsewhere will result in 409.

Collection handling was updated as storage-collections need to return both `Manifest` and `Collection` children, the following was done to handle this:
* Introduced common base for `Manifest` and `Collection` : `IHierarchyResource` - this is used in various places (ie to generate ids etc).
* `CollectionWithItems` class now stores a list of `Hierarchy`, rather than `Collection` items.
* `FullPath` is set on parent collections - this will either be taken from the Path, or calculated from DB depending on request. This serves as the base for building the `FullPath` for each child `Hierarchy` - this fixes a bug where child paths would only reflect 2 levels of parents.

Other changes:
* Refactored some helpers out of `HttpResponseMessageX` and moved to `IIIFResponseX` - this will help deserialise IIIF resources to iiif-net resources and our custom impls.
  * _Added a `catch (JsonReaderException)`, to handle non-JSON response - it didn't look to have any affect but calling out_
* Move `TryRunMigrations` outside of check for dev environment - this still requires `"RunMigrations"=true`
* `GetRawRequestBodyAsync()` extension to aid getting body as string
* Added `Forbidden` helper to shorten `this.PresentationProblem(statusCode: (int)HttpStatusCode.Forbidden);`
* Added `PresentationCreated` helper, this will use `.AsJson()` to pretty-print IIIF responses.
* Added `IIIFS3Service` as a higher level class for saving IIIF resources to S3 - will ensure flatId is set and context is present.
  * _Collections should update to use this for consistency_

The following changes should be temporary and applied, these were left as-is to minimise disruption in other classes.
* Overload of `ToFlatCollection` and `ToHierarchicalCollection` that take `List<Hierarchy>` rather than `List<Collection>` and uses this to generate items.
  * Check if any methods in `CollectionHelperX` can go after this is resolved
* `RetrieveCollectionAsync()` is a pass through to `Retrieve()` - this can be replaced when merged.


src/IIIFPresentation/API.Tests/Integration/ModifyManifestTests.cs
```diff
@@ -0,0 +1,378 @@
...
+        var response = await httpClient.AsCustomer(1).SendAsync(requestMessage);
+
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.Conflict);
+    }
+    
+    [Fact]
+    public async Task CreateManifest_Conflict_IfParentAndSlugExist_ForCollection()
+    {
+        var collectionId = nameof(CreateManifest_Conflict_IfParentAndSlugExist_ForCollection);
```
griffri, line 173: This test is missing an "arrange" comment

src/IIIFPresentation/API.Tests/Integration/ModifyManifestTests.cs
```diff
@@ -0,0 +1,378 @@
...
+        var response = await httpClient.AsCustomer(1).SendAsync(requestMessage);
+
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.Forbidden);
+    }
+
+    [Fact]
+    public async Task CreateManifest_BadRequest_IfUnableToDeserialize()
+    {
+        var requestMessage =
```
griffri, line 83: This and several following tests seem to be missing an "Arrange" comment

src/IIIFPresentation/API/Infrastructure/ControllerBaseX.cs
```diff
@@ -157,4 +159,37 @@ public static ObjectResult PresentationNotFound(this ControllerBase controller,
...
+                return new ContentResult
+                {
+                    Content = jsonLdBase.AsJson(),
+                    ContentType = IIIF.Presentation.ContentTypes.V3,
+                    StatusCode = 201,
+                };
+            }
+            catch (Exception ex)
+            {
+                var asdas = ex.Message;
```
griffri, line 189: This doesn't seem to be used anywhere - should it be logging the message instead?

donaldgray: My bad - was troubleshooting and missed this. Will remove.

griffri (approved):

src/IIIFPresentation/Models/Database/Collections/Collection.cs
```diff
@@ -39,7 +39,7 @@ public class Collection
     public DateTime Modified { get; set; }
 
     /// <summary>
-    /// Who last committed a change to this Collection
+    /// Who last created this Collection
```
JackLewis-digirati, line 42: maybe removing `last` would be good on this as well, as I don't think this can be changed

src/IIIFPresentation/API/Features/Manifest/Helpers/ErrorHelper.cs
```diff
@@ -0,0 +1,13 @@
...
+using Models.API.General;
+
+namespace API.Features.Manifest.Helpers;
+
+public class ManifestErrorHelper
+{
+    public static ModifyEntityResult<TCollection, ModifyCollectionType> ParentMustBeStorageCollection<TCollection>()
+        where TCollection : class
+        => ModifyEntityResult<TCollection, ModifyCollectionType>.Failure("The parent must be a storage collection",
+            ModifyCollectionType.ParentMustBeStorageCollection, WriteResult.Conflict);
```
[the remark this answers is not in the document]
donaldgray: Yeah it should, I didn't touch this here as there's >40 usages so didn't want to add to already large PR.

There are a few areas that use `"Collection"` that are now shared - will revisit in later change.

PR #117, opened 2024-10-23
Validate hierarchical manifest parent

donaldgray:
Follows on from #116 - which was rebased onto `main` after #110 had been merged. The latter introduced logic for validating hierarchical parent, this change adds that to `Manifest` handling and refactors to a common helper.

Introduces new `IPresentation.IsUriParentInvalid()` helper which encapsulates this logic.

Also resolves issue where `Manifest.Parent` was returning Id only, rather than Flat path.

### DIFF
--- a/src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
+++ b/src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
@@ -1,5 +1,6 @@
 ﻿using API.Converters;
 using Models.API.Manifest;
+using Models.Database.General;
 using DBManifest = Models.Database.Collections.Manifest;
 
 namespace API.Tests.Converters;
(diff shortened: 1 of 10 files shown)


src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
```diff
@@ -16,7 +17,8 @@ public void SetGeneratedFields_AddsCustomContext()
             CustomerId = 123,
             Created = DateTime.UtcNow,
             Modified = DateTime.UtcNow,
-            Id = "id"
+            Id = "id",
+            Hierarchy = [new Hierarchy { Slug = "slug" }],
```
JackLewis-digirati, line 21: this is repeated - is it worth creating a method to generate this

JackLewis-digirati (approved):

Issue #62, opened 2024-10-03
Save an "external" Manifest


tomcrane:
POST and PUT

Take any existing manifest (not referencing DLCS assets) and save it using the RFC techniques

This needs to make rows in canvas table - so needs to mint IDs

tomcrane:
Tom - write code to turn `items` into CP rows
Tom - write code to turn CP rows into `paintedResources`

PR #126, opened 2024-10-25
Initial PUT manifest

donaldgray:
This is for ticket #62 but is only the first part - there's been a fair amount of refactoring so creating PR now to avoid the full PR being too large.

This adds `PUT /manifests/{manifest-id}` which will create a new manifest with specified Id OR update existing manifest with specified Id.

Create logic was already implemented for POST so refactored this out of mediatr handler and into `ManifestService` - this is realyl the majority of the PR. This also gained an additional logic for handling Update logic. I expect this class will change so I've not been overly precious about how it's laid out, main point was to get reusable stuff together.

Also renamed `ModifyManifestTests` to `ModifyManifestCreateTests` to avoid it getting overly large. I added `PUT` create tests to this and also created `ModifyManifestUpdateTests` (left 'modify' prefix to keep these classes together in sln explorer).

Started some helper methods for creating Manifest and Collection entities to remove noise from test classes.


src/IIIFPresentation/API.Tests/Integration/ModifyManifestUpdateTests.cs
```diff
@@ -0,0 +1,361 @@
...
+                manifest.AsJson());
+        SetCorrectEtag(requestMessage, dbManifest);
+
+        // Act
+        var response = await httpClient.AsCustomer(Customer).SendAsync(requestMessage);
+
+        // Assert
+        var error = await response.ReadAsPresentationResponseAsync<Error>();
+
+        // Assert
```
griffri, line 194: There's two `assert` comments here

griffri (approved):

src/IIIFPresentation/API.Tests/Integration/ModifyManifestUpdateTests.cs
donaldgray: This is first part of fix for #62 so will remove in follow up change

p-kaczynski (approved):
LGTM

PR #138, opened 2024-11-08
Handle customer created

donaldgray:
Add SQS listener for 'customer created' notification from Protagonist, on receipt ensure a Root collection exists for customer. Implementation is basically a `BackgroundWorker` in the API project, handling logic is all in `CustomerCreatedMessageHandler`.

Lots of this PR is boilerplate for SQS listening, based on previous project.

Note that if the SQS queue name setting isn't set the background worker won't run - this can be default in dedicated instances or for local dev.

Related (publishing side) - https://github.com/dlcs/protagonist/pull/914

### DIFF
--- a/src/IIIFPresentation/API/KnownCollections.cs
+++ b/src/IIIFPresentation/API/KnownCollections.cs
@@ -1,0 +1,9 @@
+﻿namespace API;
+
+public static class KnownCollections
+{
+    /// <summary>
+    /// The id of root collection
+    /// </summary>
+    public const string RootCollection = "root";
+}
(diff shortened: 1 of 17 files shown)


p-kaczynski (approved):

src/IIIFPresentation/API/KnownCollections.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace API;
+
+public static class KnownCollections
```
JackLewis-digirati, line 3: I think this might be (correctly) duplicating functionality of the `Test.Helpers.Helpers.RootCollection` class - worth removing that class now this has been added?

JackLewis-digirati (approved):

Issue #140, opened 2024-11-12
Cache-Control headers causes browser to cache incorrectly after updates


JackLewis-digirati:
From Gary:

> we have `Cache-Control` headers in addition to etags in iiif-presentation and it seems to be breaking browser caching (the browser isn't sending an `If-None-Match`)
> I had to clear my browser cache to get an updated response from /collections/root?page=1&type=Manifest

WE need to stop the browser caching responses after an update to a resource - might be complicated by how the hierarchical resources live on a different URL to create/update endpoints meaning the caching isn't invalidated

**NOTE:** look into `Content-Location` headers which should cause the browser to remove items from other URL's - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Location

alternatively, `no-cache` will work, with a draft implementation here - https://github.com/dlcs/iiif-presentation/pull/141/files

PR #141, opened 2024-11-12
Set Cache-Control: no-cache for responses with an ETag

garyttierney:
Background: right now browsers will cache IIIF resources to disk for the caching period before checking if new content is available. That means we're never sending up an `If-None-Match` and getting a corresponding 304, so we end up with stale data in the browser. This change makes clients always send an `If-None-Match` before using the local cache.

Related to https://github.com/dlcs/iiif-presentation/issues/140

### DIFF
--- a/src/IIIFPresentation/API.Tests/Attributes/ETagCachingAttributeTests.cs
+++ b/src/IIIFPresentation/API.Tests/Attributes/ETagCachingAttributeTests.cs
@@ -111,7 +111,7 @@ public class ETagCachingAttributeTests
         });
 
         // Assert
-        Assert.Single(context.HttpContext.Response.Headers.CacheControl.ToList(), "max-age=10");
+        Assert.Single(context.HttpContext.Response.Headers.CacheControl, "max-age=10");
     }
     
     [Fact]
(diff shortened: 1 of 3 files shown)


donaldgray (approved):
LGTM - brackets on same line is inconsistent with rest of code base (nothing enforcing this).

Worth adding a test, a regression here would cause FE issues?

garyttierney:
> Worth adding a test, a regression here would cause FE issues?

What test suite would that go in (i.e. do we have an integration suite in the .NET code, or do you mean e2e)?

garyttierney:
Actually, it was the auto-formatter that put the braces on the same line. Can someone check the correct code style into the shared DotSettings in Rider? 
image

donaldgray:
> What test suite would that go in (i.e. do we have an integration suite in the .NET code, or do you mean e2e)?

Integration tests are in https://github.com/dlcs/iiif-presentation/tree/main/src/IIIFPresentation/API.Tests/Integration - I think a test for this could be added to GET Manifest and/or Collection tests.

Using existing integration test class will bootstrap db (and localstack if required) and configure `WebApplicationFactory` for making http requests.

donaldgray (approved):

PR #149, opened 2024-11-14
PresentationManifest and Collection gain FlatId

donaldgray:
`"flatId"` is now output on `PresentationManifest` and `PresentationCollection` objects. 

This saves parsing the `"id"` to get the last slug, for now they match but could differ in the future.


p-kaczynski (approved):
well, that's pretty straightforward :D

JackLewis-digirati (approved):
LGTM, but it does feel a little redundant when we have the `id` field

PR #179, opened 2024-11-21
When no items returned, set totalItems from db

p-kaczynski:
...because it means the page/pageSize could be out of bounds


donaldgray (approved):
LGTM - but agree w/ comment on ticket re: 400

PR #182, opened 2024-11-21
Fix #29 - hierarchy.slug use citext

### LINKED ISSUE #29: Make storage collection `slug` case insensitive
Storage collection slugs are currently case sensitive - e.g, you can `POST` two child storage collections with "mycollection" and "myCollection" as slugs, and you'd end up with two different collections, available from `(parent)/mycollection` and `(parent)/myCollection`.

They should be made case insensitive to avoid any potential confusion.

**Acceptance Criteria:**
- 409 Conflict is returned if an attempt is made to create a storage collection using a slug that is already in use, regardless of casing.

p-kaczynski:
- **Tests to cover #29**
- **Fix #29 - Use citext for hierarchy.slug**

From what I found online I believe it is supported in AWS RDS.
`citext` only differs from `text` that it uses `TOLOWER()` but only during comparison operations.

Alternative solutions are available, but seem either more work intensive or more error-prone.


src/IIIFPresentation/API.Tests/Integration/ModifyManifestUpdateTests.cs
```diff
@@ -151,12 +151,34 @@ public async Task PutFlatId_Update_BadRequest_IfParentAndSlugAlreadyExist_ForCol
         // Assert
         response.StatusCode.Should().Be(HttpStatusCode.Conflict);
     }
+
+    [Fact]
+    public async Task PutFlatId_Update_BadRequest_IfParentAndSlug_VaryCase_ForCollection()
```
donaldgray, line 156: Tiny thing but `_BadRequest_` for these 2 tests should be `_Conflict_`

donaldgray (approved):
Not familiar with `citext` but read up on it and looks like it should do the job.

What SQL is generated for the migration? Is it `CREATE EXTENSION IF NOT EXISTS citext;`? Previously I've ran into issues with needing to be an `rds_superuser` to install extensions but that looks like it's changed in RDS with Postgres13+ (see https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.PostgreSQL.CommonDBATasks.Extensions.html).

The user has `create` privs but just flagging in case migrations act up! <- we should log _somewhere_ (readme?) that the user running migrations requires `create` permission on DB for it to work.

src/IIIFPresentation/API.Tests/Integration/ModifyManifestUpdateTests.cs
p-kaczynski: More than 2 actually, checked all over and fixed naming

PR #188, opened 2024-11-22
Moving path location to a single location

### LINKED ISSUE #170: move URL generation to a single place
Currently paths can be retrieved off `Collection` and `PostHierarchicalCollection` - it would be best to move this to a single class so that only 1 location needs to be changed if paths change.

Acceptance Criteria
* Paths are all created in 1 single place

Stretch: Move away from extension methods in favour of something injected as this will make custom path generation easier in the future.

JackLewis-digirati:
Resolves #170

This PR removes the ability for `PostHierarchicalCollection` to generate a collection and uses the centralized methods for URL generation

Additionally, this PR moves the extension methods used to generate paths to an injected class which should provide more extensability


src/IIIFPresentation/API.Tests/Converters/Streaming/S3StoredJsonProcessorTests.cs
```diff
@@ -3,51 +3,56 @@
...
 using API.Helpers;
 using LateApexEarlySpeed.Xunit.Assertion.Json;
+using Microsoft.AspNetCore.Http;
 using Models.Database.General;
 
 namespace API.Tests.Converters.Streaming;
 
 public class S3StoredJsonProcessorTests
 {
+    private readonly IPathGenerator pathGenerator = new PathGenerator(new HttpContextAccessor()
```
donaldgray, line 13: These 10 lines are duplicated in a lot of test classes - worth having something shared?

src/IIIFPresentation/API/Converters/CollectionConverter.cs
```diff
@@ -167,53 +169,55 @@ public static List<string> GenerateContext()
     /// </summary>
     /// <param name="collection">The collection to use in generation</param>
     /// <param name="urlRoots">The URL</param>
```
donaldgray, line 171: Can go

src/IIIFPresentation/API/Converters/CollectionConverter.cs
```diff
@@ -147,15 +147,17 @@ public static List<string> GenerateContext()
     /// <param name="hierarchy">The hierarchy to use to generate</param>
     /// <param name="collection">The collection required</param>
     /// <param name="urlRoots">The URL</param>
```
donaldgray, line 149: Can go

src/IIIFPresentation/API/Converters/ManifestConverter.cs
```diff
@@ -17,25 +17,26 @@ public static class ManifestConverter
     /// <param name="iiifManifest">Presentation Manifest to update</param>
     /// <param name="dbManifest">Database Manifest</param>
     /// <param name="urlRoots">Current UrlRoots instance</param>
```
donaldgray, line 19: `urlRoots`

src/IIIFPresentation/API/Helpers/IPathGenerator.cs
```diff
@@ -0,0 +1,58 @@
+﻿using API.Converters;
+using Models.Database.Collections;
+using Models.Database.General;
+
+namespace API.Helpers;
+
+public interface IPathGenerator
```
donaldgray, line 7: Worth adding xml comments to all for consistency (or removing if obvious what's happening)?

src/IIIFPresentation/API/Program.cs
```diff
@@ -54,6 +55,8 @@
 builder.Services.AddCaching(cacheSettings);
 builder.Services
     .AddSingleton<IETagManager, ETagManager>()
+    .AddSingleton<IPathGenerator, PathGenerator>()
+    .AddSingleton<IHttpContextAccessor, HttpContextAccessor>()
```
donaldgray, line 59: There's a framework method for this `.AddHttpContextAccessor()`. I suspect it does the same as this under the hood but might be worth changing so that if `AddHttpContextAccessor` in dotnet10 does something different we get that behaviour

donaldgray (commented):
Looks like `UrlRoots` is still passed to all of the mediatr requests but I suspect those can be removed now? If, after removal, `UrlRoots` class isn't required can it go?

donaldgray (approved):

PR #191, opened 2024-11-27
Create a new project for running background services

### LINKED ISSUE #172: Create job-runner project
This is preparatory work for `paintedResource` handling.

Move the `CustomerCreatedListenerService : BackgroundWorker` from the API project to it's own separate project. This project "job-runner" (name me better) project will hold a couple of listener services, customer-created being the first.

Behaviour should continue to work as-is. 

There will be multiple `BackgroundWorkers` in "job-runner" so have a method to enable/disable based on appSettings.

JackLewis-digirati:
Resolves #172

This Pr adds a new project for running background services called `BackgroundServices` and moves the `CustomerCreatedMessageHandler` into it. 

By itself, this work does not provide any new functionality, but is in preparation of handling asset creation hooks in the DLCS


src/IIIFPresentation/AWS/SQS/Models/SubscribedToQueue.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Diagnostics;
+
+namespace AWS.SQS.Models;
+
+/// <summary>
```
donaldgray, line 5: Double opening + closing `<summary>`

src/IIIFPresentation/AWS/SQS/Models/SubscribedToQueue.cs
```diff
@@ -0,0 +1,19 @@
...
+/// <summary>
+/// Model representing a queue that has been subscribed to.
+/// </summary>
+/// </summary>
+/// <param name="Name">Name of queue</param>
+/// <param name="MessageType">Type of message handled by queue</param>
+/// <param name="Url">Url of queue</param>
+/// <typeparam name="TMessageType"></typeparam>
```
donaldgray, line 13: Empty element, update or remove

src/IIIFPresentation/BackgroundHandler.Tests/CustomerCreation/CustomerCreationMessageHandlerTests.cs
```diff
@@ -1,16 +1,17 @@
...
 using Repository;
 using Test.Helpers.Integration;
 
-namespace API.Tests.Features.CustomerCreation;
+namespace BackgroundHandler.Tests.CustomerCreation;
 
 [Trait("Category", "Database")]
 [Collection(CollectionDefinitions.DatabaseCollection.CollectionName)]
-public class CustomerCreationMessageHandlerTests
+public class CustomerCreationMessageHandlerTests : IClassFixture<PresentationContextFixture>
```
donaldgray, line 14: Class fixture not required?

src/IIIFPresentation/BackgroundHandler/BackgroundHandler.csproj
```diff
@@ -0,0 +1,26 @@
...
+      <PackageReference Include="Serilog" Version="4.0.0" />
+      <PackageReference Include="Serilog.AspNetCore" Version="8.0.3" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <ProjectReference Include="..\AWS\AWS.csproj" />
+      <ProjectReference Include="..\Repository\Repository.csproj" />
+    </ItemGroup>
+
+    <ItemGroup>
```
donaldgray, line 20: Remove this if it's not checked in?

src/IIIFPresentation/BackgroundHandler/Infrastructure/ServiceCollectionX.cs
```diff
@@ -0,0 +1,50 @@
...
+        return services;
+    }
+    
+    public static IServiceCollection AddBackgroundServices(this IServiceCollection services, AWSSettings aws)
+    {
+        
+        if (!string.IsNullOrEmpty(aws.SQS.CustomerCreatedQueueName))
+        {
+            services
+                .AddHostedService(serviceProvider => new CreateBackgroundListenerService<CustomerCreatedMessageHandler>(
```
donaldgray, line 32: You don't need to `new` this up entirely, you can use `ActivatorUtilities.CreateInstance<T>` to do the same thing. By using that you only need to provide some of the parameters, the rest are fetched from the `ServiceProvider`. End result is the same but means if any additional ctor params are added you don't need to change this call.

```cs
services
 .AddHostedService(serviceProvider => new CreateBackgroundListenerService<CustomerCreatedMessageHandler>(
    serviceProvider.GetRequiredService<SqsListener>(), 
    aws.SQS.CustomerCreatedQueueName, 
    serviceProvider.GetRequiredService<ILogger<CustomerCreatedMessageHandler>>()))`
```

could be 
```cs
services
  .AddHostedService(sp => 
     ActivatorUtilities.CreateInstance<BackgroundListenerService>(sp, aws.SQS.CustomerCreatedQueueName));
```

src/IIIFPresentation/BackgroundHandler/appsettings.json
```diff
@@ -0,0 +1,23 @@
...
+      "Override": {
+        "Microsoft": "Warning",
+        "System": "Warning"
+      }
+    },
+    "WriteTo": [
+      {
+        "Name": "Console",
+        "Args": {
+          "outputTemplate": "[{Timestamp:HH:mm:ss} {Level:u3} {SourceContext} {CorrelationId}] {Message:lj}{NewLine}{Exception}"
```
donaldgray, line 17: Will we need `{CorrelationId}` here? If so, how will it be set? 

An approach I've found useful elsewhere for projects like this (multiple background services) is to use an alternative property name, say, `{ServiceName}`. Then in each listener/handler can push a `ServiceName` property to the current log-context, that way all of the logs contain which service context it's running as. To aid setting `ServiceName` I've used an static method, e.g.

```cs
public static class LogContextHelpers
{
    /// <summary>
    /// Add "ServiceName" property to log context, which is then output as part of default log template.
    /// This is useful to filter logs as there can multiple processes running at the same time 
    /// </summary>
    public static IDisposable SetServiceName(string serviceName) =>
        LogContext.PushProperty("ServiceName", serviceName, false);
}
```

Then `CustomerCreatedMessageHandler` can start with

```cs
public async Task<bool> HandleMessage(QueueMessage message, CancellationToken cancellationToken)
{
    try
    {
        using (LogContextHelpers.SetServiceName(typeof(CustomerCreatedMessageHandler)))
        {
            var customerCreatedMessage = DeserializeMessage(message);
            .. snip ..
```

donaldgray (approved):

PR #192, opened 2024-11-27
Fix #164 - Redirect trailing slash

### LINKED ISSUE #164: 404 if trailing slash when fetching Hierarchical
Doesn't happen for Root element but does for children. e.g. 

* GET `/{customer}` - returns 200 + root
* GET `/{customer}/` - returns 200 + root
* GET `/{customer}/child-collection` - returns 200 + collection
* GET `/{customer}/child-collection/` - returns 404
* GET `/{customer}/child-manifest` - returns 200 + manifest
* GET `/{customer}/child-manifest/` - returns 404

p-kaczynski:
- **#164 trailing slash tests**
- **Set global redirect rule to remove trailing slash**

Note, that if we happen to require more granular controls there's 1001 ways to do this, but I didn't want to overengineer sth that might never be needed.

### DIFF
--- a/src/IIIFPresentation/API/Program.cs
+++ b/src/IIIFPresentation/API/Program.cs
@@ -1,7 +1,6 @@
 using System.Text.Json.Serialization;
 using API.Auth;
 using API.Features.Manifest;
-using API.Features.Storage.Validators;
 using API.Helpers;
 using API.Infrastructure;
 using API.Infrastructure.Helpers;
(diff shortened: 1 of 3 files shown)


src/IIIFPresentation/API/Program.cs
```diff
@@ -88,7 +88,11 @@
     app.UseSwaggerUI();
 }
 
+var rewriteOptions = new RewriteOptions()
+    .AddRedirect("(.*)/$", "$1");
```
JackLewis-digirati, line 92: would it better to use `.AddRewrite` over `.AddRedirect`? Would save the client from having to make another call

p-kaczynski: One is telling customer to call properly, the other opens a pandoras box :D

PR #193, opened 2024-11-28
Create `paintedResources` for external items

### LINKED ISSUE #62: Save an "external" Manifest

donaldgray:
Handle `"items"` property in POST/PUT manifest and reflect changes in `canvas_paintings` DB table.

Resolves #62 (note that PR #126 was a prefactor for #62)

### Summary of main implementation

New class `ManifestItemsParser` parses incoming `Manifest` and generates a list of `CanvasPainting` objects. These are then handled via the `CanvasPaintingResolver` to work out what create/update/delete operations are required to reconcile incoming payload with changes in database.

Note on db context changes
```cs
entity.Property(cp => cp.ChoiceOrder)
  .IsRequired()
  .HasField("internalChoiceOrder")
  .UsePropertyAccessMode(PropertyAccessMode.Field);
```

This means `ChoiceOrder` in POCO is nullable (as it's `int?`) but the DB value is not nullable (`.IsRequired()`). When EF accesses this it'll use the private `internalChoiceOrder` field (`.HasField("internalChoiceOrder")`) which is set to `-1` in place of `null`. This allows us to match "null" values in unique constraint.

### Other changes / Notes

Added new test file specifically for testing external `"items"` - left full chunks of JSON inline as it felt a little awkward to need to flit between a separate test class to see what was being posted.

* Introduce `IdentityManager` as a wrapper around `IIdGenerator` - this gives helpers for create single + multiple ids. Extension method marked as `[Obsolete]` - will follow up with PR removing this shortly.
* Added `.gitattributes` (haven't reprocessed anything - we should do at some point).


src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
```diff
@@ -143,5 +143,43 @@ public void SetGeneratedFields_SetsParentAndSlug_FromHierarchyUsingFactory()
...
+                    CanvasOrder = 100
+                }
+            ]
+        };
+        
+        // Act
+        var result = iiifManifest.SetGeneratedFields(dbManifest, pathGenerator);
+
+        // Assert
+        var cp = result.PaintedResources.Single().CanvasPainting;
```
griffri, line 179: Would it be worth changing `cp` to something more descriptive like `itemCanvasPainting`? (This probably doesn't matter too much, but thought I'd add as it seems like shorthand variable names like these are usually reserved for loops and lambda expressions)

src/IIIFPresentation/API.Tests/Integration/ModifyManifestExternalItemsTests.cs
```diff
@@ -0,0 +1,1257 @@
...
+            ]
+        }}
+    ]
+}}";
+        
+        var requestMessage =
+            HttpRequestMessageBuilder.GetPrivateRequest(HttpMethod.Put, $"{Customer}/manifests/{id}", manifest);
+        
+        // Act
+        var response = await httpClient.AsCustomer(1).SendAsync(requestMessage);
```
griffri, line 451: Would it be worth defining the customer's ID as a constant value shared between the tests?

donaldgray: Default of `.AsCustomer()` is 1 so will remove.

src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
donaldgray: Feels okay in the context of a single test?

JackLewis-digirati (approved):

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -0,0 +1,164 @@
...
+            if (matching == null)
+            {
+                if (incoming.ChoiceOrder.HasValue)
+                {
+                    // This is a choice. If there are other, existing items for the same canvas then seed canvas_id
+                    incoming.Id = candidates.SingleOrDefault()?.Id;
+                }
+
+                // Store it in a list for processing later (e.g. for bulk generation of UniqueIds)
+                toInsert.Add(incoming);
```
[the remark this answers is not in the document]
donaldgray: What would be a valuable message to log that actually tells something useful? `canvasOriginalId`?  `canvas_order`, `choice_order`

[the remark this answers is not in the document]
JackLewis-digirati: potentially the `CanvasPaintingId`? - I was thinking more in lines that every other path in this method is logged, so it might be worth it to get this too - I'm pretty happy for the answer to that question to be "not useful" though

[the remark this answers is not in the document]
donaldgray: I think it would be useful, I just wasn't sure what would be a succinct enough message. `CanvasPaintingId` isn't known until record is saved in DB so wont' be known here.

Maybe the fact that a canvas will be added at index X is enough (for now anyway). `logger.LogTrace("Adding canvas {CanvasIndex}, choice {ChoiceIndex}", matching.CanvasIndex, matching.ChoiceIndex);`

src/IIIFPresentation/API/Infrastructure/IdGenerator/IdentityManager.cs
```diff
@@ -0,0 +1,76 @@
...
+        logger.LogTrace("Generating {IdCount} uniqueIds for {Type}", count, typeof(T).Name);
+        
+        var currentAttempt = 0;
+        var random = new Random();
+        var nonMatching = new List<string>(count);
+        while (currentAttempt < MaxAttempts)
+        {
+            var requiredCount = count - nonMatching.Count;
+            logger.LogTrace("Generating {IdCount} uniqueIds for {Type}, attempt {Attempt}", requiredCount,
+                typeof(T).Name, currentAttempt);
```
griffri, line 51: For a shorter form of `typeof(T).Name`, I think `nameof(T)` can be used

PR #198, opened 2024-12-02
Allow space creation via header

### LINKED ISSUE #139: Create a Space for a Manifest with HTTP Header
Allow a space to be created on creation of a Manifest (note this is create _only_).

See new sections in docs. This issue is only for the creation of a Space via the header, not to implement the `assets` alias property for a Space.

https://deploy-preview-2--dlcs-docs.netlify.app/api-doc/iiif#assets-property

HTTP Header is 
```
Link: <https://dlcs.io/vocab#Space>;rel="DCTERMS.requires"
```

Acceptance criteria:

- [x] A newly created Manifest made without the HTTP header does not have a `space` property (as-is)
- [x] A newly created Manifest where the HTTP Header was included has a newly created Space which is the API URL of that Space
  - `Manifest.Space` is stored in DB as integer value (URL expose in payload can be inferred from this and customer)
  - The space functions exactly as normal
- [x] POSTing a Manifest body with a `space` property has no effect - the `space` property is ignored on input (users can't assign a different space to the manifest)
- [x] If creation of Space fails in DLCS API - return a 500 and appropriate problem+json (no record / s3 content should be saved)
- [x] Default name of space to be `"For manifest {manifest-id} - {ISO-8601 format date}"` (verify field length for space name)

### LINKED ISSUE #197: Add CorrelationId to requests

donaldgray:
Resolves #139. Resolves #197

Overall change is to allow an http header to be specified on manifest _creation_, when specified this results in a DLCS space being created. The Id of this space is stored against the Manifest and it's full URL is returned when manifest is requested with show-extras header.

Introduced new project, DLCS, which holds `IDlcsApiClient`. Refactored `DelegatedAuthenticator` to use this, rather than calling DLCS directly. Implemented auth w/ a `DelegatingHandler`, `AmbientAuthHandler`, - this keeps the `IDlcsApiClient` implementation cleaner but downside is it's not immediately obvious how the auth is being handled. There are some helpers that are similar to those in `Core` project, e.g. for reading responses but kept these internal rather than trying to be too clever and share - allows DLCS to use STJ without worrying about conflicts - the impl can change depending on DLCS requirements too.

Introduced CorrelationId handling as we are now calling a downstream DLCS


src/IIIFPresentation/API.Tests/Integration/Infrastructure/TestAuthenticator.cs
```diff
@@ -11,7 +11,7 @@ public class TestAuthenticator : IAuthenticator
 {
     private const string AuthHeader = "Authorization";
     
-    public async Task<AuthResult> ValidateRequest(HttpRequest request)
+    public async Task<AuthResult> ValidateRequest(HttpRequest request, CancellationToken cancellationToken = default)
```
JackLewis-digirati, line 14: is this cancellation token being used?

src/IIIFPresentation/API/Program.cs
```diff
@@ -45,10 +51,11 @@
 var cacheSettings = builder.Configuration.GetSection(nameof(CacheSettings)).Get<CacheSettings>() ?? new CacheSettings();
 var dlcs = dlcsSettings.Get<DlcsSettings>()!;
 
-builder.Services.AddDelegatedAuthHandler(dlcs, opts =>
-{
-    opts.Realm = "DLCS-API";
-});
+var aws = builder.Configuration.GetSection("AWS").Get<AWSSettings>() ?? new AWSSettings();
```
JackLewis-digirati, line 54: I don't think this is used

src/IIIFPresentation/DLCS.Tests/API/DlcsApiClientTests.cs
```diff
@@ -0,0 +1,95 @@
+﻿using System.Net;
```
JackLewis-digirati, line 1: this file is CRLF - is that correct, or do we want LF?

src/IIIFPresentation/API.Tests/Integration/Infrastructure/TestAuthenticator.cs
donaldgray: Updated actual implementation of `IAuthenticator` to use token but `TestAuthenticator` has no need to use

src/IIIFPresentation/DLCS/Models/JsonLDBase.cs
```diff
@@ -0,0 +1,21 @@
...
+    [JsonPropertyOrder(0)]
+    public string? Context { get; set; } = null;
+
+    [JsonPropertyName("@id")]
+    [JsonPropertyOrder(1)]
+    public string? ResourceId { get; set; }
+
+    [JsonPropertyName("@type")]
+    [JsonPropertyOrder(2)]
+    public string Type { get; set; }
```
JackLewis-digirati, line 20: github warning on this line

src/IIIFPresentation/DLCS/Models/Space.cs
```diff
@@ -0,0 +1,7 @@
+namespace DLCS.Models;
+
+public class Space : JsonldBase
+{
+    public int? Id { get; set; }
+    public string Name { get; set; }
```
JackLewis-digirati, line 6: github warning

src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
```diff
@@ -17,8 +17,8 @@ public class PresentationManifest : IIIF.Presentation.V3.Manifest, IPresentation
     /// <summary>
     /// Represents, in a compact non-IIIF form, the relationship between a Content Resource and a Manifest.
     /// </summary>
-    [JsonProperty(Order = 12)]
-    public List<PaintedResource>? PaintedResources { get; set; }
+    [JsonProperty(Order = 12)] public List<PaintedResource>? PaintedResources { get; set; }
+    [JsonProperty(Order = 13)] public string? Space { get; set; }
```
JackLewis-digirati, line 21: think this might need formatting

JackLewis-digirati (commented):
LGTM, but I think most of the new files use CRLF, which means the project has a mixture of CRLF and LF

src/IIIFPresentation/DLCS.Tests/API/DlcsApiClientTests.cs
donaldgray: See https://github.com/dlcs/iiif-presentation/pull/200

src/IIIFPresentation/API/Program.cs
donaldgray: Removes

src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
donaldgray: Reformatted - let Rider auto-format.

JackLewis-digirati (approved):

Issue #167, opened 2024-11-19
Hierarchical collection contain non-public Items


donaldgray:
Only collections that are public should be contained in the `"items"` property - non public items shouldn't be included.

PR #206, opened 2024-12-09
Bugfixes - correct partOf and hide non-public items

### LINKED ISSUE #166: Storage-Collections showing incorrect root "partOf"
This issue relates to the `"partOf"` property for a storage-collection that is a first child of `root` when fetched via flat uri.

* `"id"` is the hierarchical path (ie `/{customer}` but it should be `/{customer}/collections/root`
* `"label"` is the label of the current item (ie the first-child) but it should be the label of the parent.

Verify whether this happens to all `"partOf"` (ie is it just for first-child of root or any level of hierarchy)

donaldgray:
Fixes #166 and #167

## #166 - incorrect `"partOf"`

Introduced a couple of navigation properties; `Hierarchy` gains `ParentCollection` and `Collection` gains `Children`. It's unlikely that the latter will be used but it's there should we need it. The former allows us to load the parent of the current collection when rendering collection to set the `"partOf"` - previously it was using current collections label and redirecting to parent hierarchical path - now it's using parent label and parent flat-id.

Renamed `ToFlatCollection` to `ToPresentationCollection` to reflect the name of the return type.

## #167 - exclude non-public from `"items"` in hierarchical collection

This was adding a bool to `RetrieveCollectionItems` in `GetHierarchicalCollection` but also slightly refactored this:
* Rather than checking if collection is public or not in calling controller, encapsulate that logic in the request handler, this shortcuts the request and moves the decision making into the handler.
* Use multiple returns; there are 2 distinct paths for iiif-collection or storage-collection so have a return in each. These fall through to a single `empty` return for any cases that can't be fulfulled

### DIFF
--- a/src/IIIFPresentation/API/Converters/CollectionConverter.cs
+++ b/src/IIIFPresentation/API/Converters/CollectionConverter.cs
@@ -4,19 +4,18 @@ using Core.IIIF;
 using IIIF.Presentation;
 using IIIF.Presentation.V3;
 using IIIF.Presentation.V3.Content;
-using IIIF.Presentation.V3.Strings;
 using Models.API.Collection;
-using Models.Database.Collections;
 using Models.Database.General;
 using Models.Infrastucture;
 using Collection = IIIF.Presentation.V3.Collection;
 using Manifest = IIIF.Presentation.V3.Manifest;
+using DbCollection = Models.Database.Collections.Collection;
 
 namespace API.Converters;
 
 public static class CollectionConverter
 {
-    public static Collection ToHierarchicalCollection(this Models.Database.Collections.Collection dbAsset, 
+    public static Collection ToHierarchicalCollection(this DbCollection dbAsset,
         IPathGenerator pathGenerator, List<Hierarchy>? items)
     {
         var collection = new Collection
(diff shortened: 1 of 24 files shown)


src/IIIFPresentation/API/Converters/CollectionConverter.cs
```diff
@@ -132,36 +132,33 @@ public static List<string> GenerateContext()
...
+    public static List<ResourceBase>? GeneratePartOf(DbCollection? collection,
+        IPathGenerator pathGenerator) =>
+        collection != null
+            ?
+            [
+                new ExternalResource(nameof(PresentationType.Collection))
                 {
-                    Id = pathGenerator.GenerateHierarchicalCollectionParent(collection, hierarchy),
-                    Label = collection.Label
+                    Id = pathGenerator.GenerateFlatCollectionId(collection),
```
JackLewis-digirati, line 157: Should this be changed to `GeneratePresentationCollectionId` given the rename

donaldgray: I don't think so, renaming `ToFlatCollection()` -> `ToPresentationCollection()` is to match the return type (`FlatCollection` was renamed to `PresentationCollection` in #47).

I think the flat methods on `IPathGenerator` still make sense as the paths are flat (or hierarchical). If we want to revisit we can do and rename all methods in `IPathGenerator` that refer to 'flat' paths.

PR #211, opened 2024-12-13
Fix null behavior returning internal server error

### LINKED ISSUE #165: Creating IIIF collection without "behaviour" results in 500
Happens on PUT or POST endpoint, encountered when creating (not sure if it's just creation or create and update).

Not sure if it will happen for just iiif-collection or storage-collection too.

If payload doesn't contain `"behaviour"` property then 500 returned. Checking logs only showed 500 in request-logging which makes me wonder if it's happening in deserialization? If not we should capture logs (do we need an unhandled exception handler?)

JackLewis-digirati:
resolves #165 

This PR creates a fallback value of `false` for `behavior` extension methods, causing them to not throw an error when `behavior` is null


donaldgray (approved):
LGTM - worth having unit tests for `BehaviorX` ?

PR #216, opened 2024-12-19
Ignore painted resource when items are also present

### LINKED ISSUE #215: ignore painted resource block when items is filled out
As discussed in https://digirati.atlassian.net/browse/ICS-361 this ticket is to make it so that painted resources are ignored when items are filled out, instead of producing a bad request.

JackLewis-digirati:
Resolves #215

This PR changes manifests with painted resources and items present to ignore the painted resource section of a request, instead of responding with a bad request

### DIFF
--- a/src/IIIFPresentation/API/Features/Manifest/ManifestService.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/ManifestService.cs
@@ -10,20 +10,15 @@ using API.Infrastructure.Validation;
 using Core;
 using Core.Auth;
 using Core.Helpers;
-using DLCS;
 using DLCS.API;
 using DLCS.Exceptions;
 using IIIF.Serialisation;
-using Microsoft.Extensions.Options;
 using Models.API.General;
 using Models.API.Manifest;
 using Models.Database.Collections;
 using Models.Database.General;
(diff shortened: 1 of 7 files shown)


src/IIIFPresentation/API/Features/Manifest/ManifestService.cs
```diff
@@ -123,6 +118,13 @@ private async Task<PresUpdateResult> CreateInternal(WriteManifestRequest request
         var (parentErrors, parentCollection) = await TryGetParent(request, cancellationToken);
         if (parentErrors != null) return parentErrors;
 
+        // can't have both items and painted resources, so items takes precedence
```
tomcrane, line 121: Do we need to have this behaviour controlled by an appsetting or feature flag, or is it so simple that it can be stripped out again easily? I guess the updated tests when fully implemented will make it obvious if it's still doing this.

tomcrane (approved):

src/IIIFPresentation/API/Features/Manifest/ManifestService.cs
JackLewis-digirati: I've added a setting to control this

Issue #203, opened 2024-12-06
Returning full `Asset` in Manifest response


donaldgray:
Follow on from #173 (which is returning the asset-id only) this ticket is to flesh out the returned asset result to include the full DLCS asset.

To do this call `/allImages`, specifying the ids of the assets contained in the manifest. Note that `/allImages` has a maximum number of ids so we may need to make multiple requests.

These should be returned in the same order as they are in the `canvas_paintings` table.

If an asset could not be found, return a "placeholder" asset in the result set, with just an `"id"` and `"error"`

donaldgray:
Note - this ticket mentions using `/allImages` to get details of all assets. https://github.com/dlcs/protagonist/pull/932 introduced a new endpoint where all assets could be fetched for a batch, which is a possibility.

Consider trade off of number of calls to be made, size of payloads etc.

PR #240, opened 2025-01-20
Impl #203 - retrieve asset details from DLCS when responding to Create Manifest

p-kaczynski:
#203 
- **Add ReadAsJson helper to handle not-deserialised responses from DLCS**
- **Add GetBatchAssets call to DLCS client**
- **Add ability to use provided asset dictionary during canvas painting supplementation in manifest create**
- **Code cleanup**
- **Test + fix for asset detail in manifest creation response**

### DIFF
--- a/src/IIIFPresentation/API.Tests/Integration/ModifyManifestAssetCreationTests.cs
+++ b/src/IIIFPresentation/API.Tests/Integration/ModifyManifestAssetCreationTests.cs
@@ -52,6 +52,12 @@ public class ModifyManifestAssetCreationTests : IClassFixture<PresentationAppFac
             A<List<JObject>>.That.Matches(o => o.First().GetValue("id").ToString() == "returnError"),
             A<CancellationToken>._)).Throws(new DlcsException("DLCS exception"));
 
+        A.CallTo(() => dlcsApiClient.GetBatchAssets(Customer, 404, A<CancellationToken>._))
+            .ReturnsLazily(() => []);
+
+        A.CallTo(() => dlcsApiClient.GetBatchAssets(Customer, 500, A<CancellationToken>._))
+            .Throws(new DlcsException("DLCS exception"));
+        
         A.CallTo(() => dlcsApiClient.GetBatchAssets(Customer, 2137, A<CancellationToken>._))
             .ReturnsLazily(() =>
             [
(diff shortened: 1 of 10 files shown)


src/IIIFPresentation/API/Features/Manifest/ManifestService.cs
```diff
@@ -273,7 +272,43 @@ await canvasPaintingResolver.GenerateCanvasPaintings(request.CustomerId, request
...
+        return (saveErrors, dbManifest, await fetchBatchAssets);
+    }
+
+    private async Task<Dictionary<string, JObject>> GetBatchAssets(int customerId, IList<int>? batchIds,
+        CancellationToken cancellationToken)
+    {
+        if (batchIds == null) return new();
+        List<JObject> assets = [];
+        foreach (var batchId in batchIds)
+            assets.AddRange(await dlcsApiClient.GetBatchAssets(customerId, batchId, cancellationToken));
```
JackLewis-digirati, line 284: is this potentially more work than using `allImages`?

src/IIIFPresentation/API.Tests/Integration/ModifyManifestAssetCreationTests.cs
```diff
@@ -212,6 +263,64 @@ await amazonS3.GetObjectAsync(LocalStackFixture.StorageBucketName,
         s3Manifest.Id.Should().EndWith(dbManifest.Id);
         s3Manifest.Items.Should().BeNull();
     }
+
+    [Fact]
+    public async Task CreateManifest_ReturnsAssetDetails_FromGetBatchAssets()
```
JackLewis-digirati, line 268: Is it worth adding some negative tests? I.E.: what happens when the DLCS responds with an error

src/IIIFPresentation/API/Features/Manifest/ManifestService.cs
p-kaczynski: `allImages` _might_ be paged, so you are mostly forced into sequential calls. Here I also went with sequential calls for safety (not DDoSing DLCS, not expending http pool etc.), but can be parallelised if needed, so I feel like it's more flexible. No particularly strong opinions on it tbh

src/IIIFPresentation/DLCS/API/DlcsApiClient.cs
```diff
@@ -84,13 +88,45 @@ public async Task<List<Batch>> IngestAssets<T>(int customerId, List<T> assets, C
         return batches.ToList();
     }
 
-    private async Task<T?> CallDlcsApi<T>(HttpMethod httpMethod, string path, object payload,
+    public async Task<List<JObject>> GetBatchAssets(int customerId, int batchId,
```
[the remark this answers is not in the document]
p-kaczynski: Good spot, slipped my mind. Added tests for the expected scenarios.

src/IIIFPresentation/API/Features/Manifest/ManifestService.cs
p-kaczynski: Groan.

You might be right, that is, not even about more/less work, but about the specific requirement for turning `AssetId` into either asset or error. All fine when it works, but in scenario:

1. Manifest has enough assets to trigger multiple batches
2. `GetBatchAssets` receives e.g. `500` from two batches, but `200` from third.
3. When populating response manifest we try to match `AssetId` from `CanvasPainting` to data retrieved previously, but we only know which BATCH failed, not which `Batch` contained which assets.

So if the `AssetId` it might be because it is missing (`{"id": <asset_id>, "error": "Asset details missing"}`) or because the batch failed (`{"id": <asset_id>, "error": "Some DLCS Exception message"}`), and we don't have a way of figuring it out.

But if I go and request `allImages`, all pages etc. sth similar happens: what if page e.g. `3` fails? Well, instant thought is "just retry the call" - but we're getting dangerously into territory of "fancy DLCS client".

What I can do "easily" is sth like:

```
"asset": {
  "id": <asset_id>,
  "errors": [
    "Retrieving assets from batch 123 failed",
    "Retrieving assets from batch 125 failed"
  ]
}
```

@donaldgray FYI + any ideas?

donaldgray: The `/allImages` request won't return pages results (it's not really `/allImages`, more `/imagesIAskFor`) - you POST a list of AssetId and it'll return details of those assets 

```json
{ 
  "member": [ 
    { "id": "1/2/foo"  }, 
    { "id": "1/2/bar"  } 
  ]
}
```

There's an upper limit to the number of `"member"` values you can supply so for larger manifests you could make multiple requests. If you ask for 10 you might only get 8 results if the asset has been subsequently deleted.

Using the batch `/assets` would result in multiple requests if there were more than 1 batch. This could happen because the number of image is huge, or because there's only a small number of images but they've been updated 8 times (so 8 batches in total). I think you'd still need to make 8 requests as we don't store which assets are in which batch.

It seems like `/allImages` is a larger request but might be easier to identify 'missing' requests? If it fails you can have `{"id": "1/2/foo", "error": "Unable to fetch details of this"}` or if it's not returned `{"id": "1/2/foo", "error": "Asset not found"}`

p-kaczynski:
Switched to `allImages`, added tests for it as well.

src/IIIFPresentation/DLCS/API/DlcsApiClient.cs
```diff
@@ -84,13 +91,63 @@ public async Task<List<Batch>> IngestAssets<T>(int customerId, List<T> assets, C
...
+    public async Task<IList<JObject>> GetCustomerImages(int customerId, ICollection<string> assetIds,
+        CancellationToken cancellationToken = default)
+    {
+        logger.LogTrace("Requesting images for customer {CustomerId}: {AssetIds}", customerId,
+            string.Join(",", assetIds));
+
+        if (assetIds.Count == 0)
+            return [];
+
+        var endpoint = $"/customers/{customerId}/allImages";
```
donaldgray, line 120: There's an upper limit to the number of images that can be requested in 1 request from DLCS. It defaults to 500 but can differ https://github.com/dlcs/protagonist/blob/main/src/protagonist/API/Settings/ApiSettings.cs#L38. We should have a similar config value and chunk up into POSTs of max XXX image ids

donaldgray (approved):

src/IIIFPresentation/DLCS/API/DlcsApiClient.cs
p-kaczynski: Cheers, I haven't noticed that!

Added analogous config value with same 500 default and turned the call into batched call. Again, could be at some point "improved", but that definitely gets in scope of "proper client".

PR #243, opened 2025-01-24
Include count of child items for storage-collection

### LINKED ISSUE #234: ChildStorage Counts
See parent for links.

Add new `"totals"` property on StorageCollection API responses that contains 3 new properties:
* `"childStorageCollections"`
* `"childIIIFCollections"`
* `"childManifests"`

These are the immediate children of the current object. E.g.:
```json
   "totals": {
      "childStorageCollections": 4,
      "childIIIFCollections": 0,
      "childManifests": 0
   }
```

donaldgray:
Add `"totals"` property to API Storage Collection results which returns the counts of immediate children by type.

Refactored `ToPresentationCollection` (used for GET) and `EnrichPresentationCollection` (used for POST and PUT) with the former now calling the latter. They both contained almost identical code, the difference being that the latter:
* Didn't set `Label` property. It will now always set `Label`
* Used `??=` for a few properties, meaning that the value returned is what was passed in body of request. However, the value saved to the DB could differ, meaning that a subsequent GET request would return different values. In practice I think the only property this could effect is `Behavior` (which I think isn't quite right as it prevents saving custom behaviors). The decision was to always set the value (so `=` rather than `??=`) to be consistent with GET responses.

Resolves #234

### DIFF
--- a/src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
+++ b/src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
@@ -97,29 +97,29 @@ public class CollectionConverterTests
         var expectedCounts = new DescendantCounts(1, 0, 0);
 
         // Act
-        var flatCollection =
+        var presentationCollection =
             collection.ToPresentationCollection(PageSize, 1, 1, CreateTestItems(), null, pathGenerator);
 
         // Assert
-        flatCollection.Id.Should().Be("http://base/1/collections/some-id");
-        flatCollection.FlatId.Should().Be("some-id");
-        flatCollection.PublicId.Should().Be("http://base/1");
-        flatCollection.Label!.Count.Should().Be(1);
-        flatCollection.Label["en"].Should().Contain("repository root");
-        flatCollection.Slug.Should().Be("root");
-        flatCollection.SeeAlso.Should().HaveCount(2);
-        flatCollection.SeeAlso![0].Id.Should().Be("http://base/1");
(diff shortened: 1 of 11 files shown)


src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
```diff
@@ -219,7 +228,116 @@ public void ToPresentationCollection_ConvertsStorageCollection_IncludingPartOfFo
...
+                    Slug = "root",
+                    CustomerId = 1,
+                    Type = ResourceType.StorageCollection,
+                    Canonical = true
+                }
+            ]
+        };
+
+        // Act
+        var flatCollection =
```
JackLewis-digirati, line 260: small, but should this be `presentationCollection`?

src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
```diff
@@ -219,7 +228,116 @@ public void ToPresentationCollection_ConvertsStorageCollection_IncludingPartOfFo
...
```
JackLewis-digirati, line 292: small, but should this be `presentationCollection`

JackLewis-digirati (approved):

src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
donaldgray: Renamed all usages of `flatCollection` -> `presentationCollection` in class

PR #244, opened 2025-01-24
setting status code to Accepted when ingesting

### LINKED ISSUE #230: Tighten responses for GET manifests w/ assets
This ticket is to improve the responses returned when an API GET is made for a manifest that contains assets as the assets are processed asynchronously so may or may not be complete:

For API/flat requests:
* [x] Return 202 if the manifest is still being processed, returning what's stored in S3 hydrated with API-specific properties (as-is but different status code)
* [x] Return 200 if the manifest is finished with fully fleshed out IIIF from S3 (this is as-is)

For public/hierarchical requests:
* [x] Return 404 if the manifest is still being processed AND it _has never_ before completed processing (as we won't have a full manifest in S3 to return).
* [x] Return 200 if the manifest is still being processed AND it _has_ completed processing in the past (as we will have a full manifest in S3 to return). Better to serve stale data than nothing. Do we need to set a shorter cache header here as we know it's stale? Or will etags avoid the need to do that?
* [x] Return a 200 if the manifest is completed

JackLewis-digirati:
Resolves #230 

This PR does the following

- modifies `GET` manifest to return an `Accepted` status code when assets are ingested
- adds an additional `Ingested` property to the database, denoting that the manifest has been ingested at some point in the past
- Modifies the GET hierarchical call to return `NotFound` if the manifest is ingesting, unless it has previously been ingested

### DIFF
--- a/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
+++ b/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
@@ -203,10 +203,8 @@ public class GetManifestTests : IClassFixture<PresentationAppFactory<Program>>
     [Fact]
     public async Task Get_IiifManifest_Hierarchical_ReturnsNotFoundWhenIngesting()
     {
-        // Arrange and Act
+        // Arrange
         var id = nameof(Get_IiifManifest_Hierarchical_ReturnsNotFoundWhenIngesting);
-        
-        // Arrange and Act
         await dbContext.Manifests.AddAsync(GenerateManifestRecord(id, 2));
         
         await amazonS3.PutObjectAsync(new()
@@ -218,7 +216,9 @@ public class GetManifestTests : IClassFixture<PresentationAppFactory<Program>>
         
         await dbContext.SaveChangesAsync();
         
+        // Act
         var response = await httpClient.GetAsync($"1/{id}");
+        
         // Assert
         response.StatusCode.Should().Be(HttpStatusCode.NotFound);
         response.Headers.Vary.Should().HaveCount(2);
@@ -227,10 +227,9 @@ public class GetManifestTests : IClassFixture<PresentationAppFactory<Program>>
     [Fact]
     public async Task Get_IiifManifest_Hierarchical_ReturnsOkWhenIngestingButHasIngestedBefore()
     {
-        // Arrange and Act
+        // Arrange
         var id = nameof(Get_IiifManifest_Hierarchical_ReturnsOkWhenIngestingButHasIngestedBefore);
         
-        // Arrange and Act
         await dbContext.Manifests.AddAsync(GenerateManifestRecord(id, 3, true));
         
         await amazonS3.PutObjectAsync(new()
@@ -242,6 +241,7 @@ public class GetManifestTests : IClassFixture<PresentationAppFactory<Program>>
         
         await dbContext.SaveChangesAsync();
         
+        // Act
         var response = await httpClient.GetAsync($"1/{id}");
         
         var manifest = await response.ReadAsPresentationJsonAsync<PresentationManifest>();
--- a/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
+++ b/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
@@ -168,47 +168,22 @@ public class GetManifestTests : IClassFixture<PresentationAppFactory<Program>>
     [Fact]
     public async Task Get_IiifManifest_Flat_ReturnsAccepted_WhenIngesting()
     {
+        var id = nameof(Get_IiifManifest_Flat_ReturnsAccepted_WhenIngesting);
+        
         // Arrange and Act
-        await dbContext.Manifests.AddAsync(new Models.Database.Collections.Manifest()
-        {
-            Id = "ManifestWithBatch",
-            CustomerId = 1,
-            Created = DateTime.UtcNow,
-            Modified = DateTime.UtcNow,
-            CreatedBy = "admin",
-            Hierarchy =
-            [
-                new Hierarchy
-                {
-                    Slug = "manifest-with-batch",
-                    Parent = RootCollection.Id,
-                    Type = ResourceType.IIIFManifest,
-                    Canonical = true
-                }
-            ],
-            Batches =
-            [
-                new Batch
-                {
-                    Id = 1,
-                    Submitted = DateTime.UtcNow,
-                    Status = BatchStatus.Ingesting,
-                    ManifestId = "ManifestWithBatch"
-                }
-            ]
-        });
+        await dbContext.Manifests.AddAsync(GenerateManifestRecord(id, 1));
         
         await amazonS3.PutObjectAsync(new()
         {
             BucketName = LocalStackFixture.StorageBucketName,
-            Key = "1/manifests/ManifestWithBatch",
+            Key = $"1/manifests/{id}",
             ContentBody = TestContent.ManifestJson
         });
         
         await dbContext.SaveChangesAsync();
         
         var requestMessage =
-            HttpRequestMessageBuilder.GetPrivateRequest(HttpMethod.Get, "1/manifests/ManifestWithBatch");
+            HttpRequestMessageBuilder.GetPrivateRequest(HttpMethod.Get, $"1/manifests/{id}");
         var response = await httpClient.AsCustomer().SendAsync(requestMessage);
 
         var manifest = await response.ReadAsPresentationJsonAsync<PresentationManifest>();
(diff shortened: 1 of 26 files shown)


src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
```diff
@@ -88,4 +100,128 @@ public async Task Get_IiifManifest_Hierarchical_ReturnsManifestFromS3()
...
+        manifest.Id.Should().Be($"http://localhost/1/manifests/{id}", "requested by flat URI");
+        manifest.Items.Should().HaveCount(3, "the test content contains 3 children");
+        manifest.FlatId.Should().Be(id);
+        manifest.PublicId.Should().Be($"http://localhost/1/{id}", "iiif-manifest is slug and under root");
+    }
+    
+    [Fact]
+    public async Task Get_IiifManifest_Hierarchical_ReturnsNotFoundWhenIngesting()
+    {
+        // Arrange and Act
```
griffri, line 142: This test has two "Arrange and Act" comments. These comments just help split up the tests into chunks so they're not crucial, but would it be worth removing the duplicate for consistency?

(@donaldgray had a chat about this a while back - arrange/act/assert is a useful pattern but not one we need to strictly adhere to - comments that explain what the test is doing might be more useful for more complex ones)

src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
```diff
@@ -88,4 +100,128 @@ public async Task Get_IiifManifest_Hierarchical_ReturnsManifestFromS3()
...
+        var response = await httpClient.GetAsync($"1/{id}");
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.NotFound);
+        response.Headers.Vary.Should().HaveCount(2);
+    }
+
+    [Fact]
+    public async Task Get_IiifManifest_Hierarchical_ReturnsOkWhenIngestingButHasIngestedBefore()
+    {
+        // Arrange and Act
```
griffri, line 166: This test also has two "Arrange and Act" comments

src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
```diff
@@ -24,6 +24,9 @@ public class PresentationManifest : IIIF.Presentation.V3.Manifest, IPresentation
     [JsonProperty(Order = 13)] public string? Space { get; set; }
 
     [JsonIgnore] public string? FullPath { get; set; }
+    
+    [JsonIgnore] 
+    public bool Ingesting { get; set; }
```
griffri, line 29: Should this be `[JsonIgnore] public bool Ingesting { get; set; }` for consistency? (`PaintedResources` also has its annotation on the previous line, so I'm not sure)

griffri (approved):
LGTM, just a few minor things

src/IIIFPresentation/API/Infrastructure/ControllerBaseX.cs
```diff
@@ -173,7 +174,7 @@ public static ObjectResult Forbidden(this ControllerBase controller)
     /// Create a 201 result with standard serialized value or custom serialized IIIF object if value is
     /// <see cref="JsonLdBase"/> 
     /// </summary>
-    public static ActionResult PresentationCreated(this ControllerBase controller, string? uri, object? value)
+    public static ActionResult PresentationWithBodyResponse(this ControllerBase controller, string? uri, object? value, int statusCode)
```
donaldgray, line 177: Update XML comment if this can return 201 + non-201?

src/IIIFPresentation/Repository/Helpers/CollectionRetrieval.cs
```diff
@@ -164,6 +164,7 @@ INNER JOIN
             .FromSqlRaw(query)
             .Include(h => h.Collection)
             .Include(h => h.Manifest)
+            .ThenInclude(m => m.Batches)
```
donaldgray, line 167: Should this be added to the similar query above in the `if (slug.Equals(string.Empty))` clause?

src/IIIFPresentation/Models/Database/Collections/Manifest.cs
```diff
@@ -47,6 +47,11 @@ public class Manifest : IHierarchyResource
     public List<CanvasPainting>? CanvasPaintings { get; set; }
     
     public List<Batch>? Batches { get; set; }
+    
+    /// <summary>
+    /// Whether the manifest has been ingested with assets at some point
+    /// </summary>
+    public bool Ingested { get; set; }
```
donaldgray, line 54: Would there be any use in this being a `DateTime? LastProcessed` or `DateTime? LastFinished` or something? Would provide both an easy way of identifying both whether it was and when it was.

Alternatively a `bool?` as some Manifest will never Ingest with assets (ie if they've only ever been vanilla IIIF). With this `false` means "not done yet" but also "won't ever be done"

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -181,7 +177,7 @@ public class CanvasPaintingResolver(
     }
     
     private async Task<(PresUpdateResult? updateResult, List<CanvasPainting>? canvasPaintings)> CreateCanvasPaintingsFromAssets(
-        int customerId, PresentationManifest presentationManifest, int manifestSpace, CancellationToken cancellationToken)
+        int customerId, PresentationManifest presentationManifest, int? manifestSpace, CancellationToken cancellationToken)
```
donaldgray, line 180: Has something in this PR changed nullability of this?

src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
```diff
@@ -88,4 +100,128 @@ public async Task Get_IiifManifest_Hierarchical_ReturnsManifestFromS3()
...
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+        response.Headers.Should().ContainKey(HeaderNames.ETag);
+        response.Headers.Vary.Should().HaveCount(2);
+        manifest.Should().NotBeNull();
+        manifest!.Type.Should().Be("Manifest");
+        manifest.Id.Should().Be($"http://localhost/1/{id}", "requested by hierarchical URI");
+        manifest.Items.Should().HaveCount(3, "the test content contains 3 children");
+    }
+
+    private static Models.Database.Collections.Manifest GenerateManifestRecord(string id, int batchId, bool ingested = false)
```
donaldgray, line 195: Is it worth extending the `AddTestManifest` method, or creating an overload, so that it's widely usable moving forward?

src/IIIFPresentation/API/Features/Manifest/Requests/GetManifestHierarchical.cs
```diff
@@ -41,8 +42,12 @@ public class GetManifestHierarchicalHandler(
             new(settings.S3.StorageBucket, BucketHelperX.GetManifestBucketKey(request.Hierarchy.CustomerId, flatId)),
             cancellationToken);
 
-        if (objectFromS3.Stream.IsNull())
+        if (objectFromS3.Stream.IsNull() ||
+            (request.Hierarchy.Manifest.IsIngesting() &&
```
donaldgray, line 46: This method should be shortcut, requests for public manifests should be as quick as possible. There's no need to get the fullpath from DB or hit S3 etc to know that it definitely won't be served due to processing state.

Do we need both `request.Hierarchy.Manifest.IsIngesting() && !request.Hierarchy.Manifest.HasIngestedPreviously()`? I _think_ the latter would suffice? If it's ever processed then it's serveable, regardless of whether it's ingesting now or not.

src/IIIFPresentation/Repository/Helpers/CollectionRetrieval.cs
JackLewis-digirati: I don't think this can be hit as it would mean it has no `parent` record, which is only possible with the `root` collection

donaldgray: Worth putting something to that effect in the code as at a glance they're very similar

src/IIIFPresentation/API/Features/Manifest/Requests/GetManifestHierarchical.cs
JackLewis-digirati: I think just using `!request.Hierarchy.Manifest.HasIngestedPreviously()` doesn't quite work as it comes back `false` for manifests which have never been ingested

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -279,7 +280,8 @@ await canvasPaintingResolver.GenerateCanvasPaintings(request.CustomerId, request
                 }
             ],
             CanvasPaintings = canvasPaintings,
-            SpaceId = spaceId
+            SpaceId = spaceId,
+            LastProcessed = canvasPaintings?.Any(cp => cp.AssetId != null) ?? false ? null : DateTime.UtcNow
```
donaldgray, line 284: If modified and created are both set to the exact same timestamp should this use that too? (I can't recall the reason for c + m using exact same value but I'm assuming it's for ease of identifying that it happened at the same timE)

src/IIIFPresentation/API/Infrastructure/ControllerBaseX.cs
```diff
@@ -170,10 +171,11 @@ public static ObjectResult Forbidden(this ControllerBase controller)
         => controller.PresentationProblem(statusCode: (int)HttpStatusCode.Forbidden);
     
     /// <summary>
-    /// Create a 201 result with standard serialized value or custom serialized IIIF object if value is
+    /// Creates either a 201 or 202 result with standard serialized value or custom serialized IIIF object if value is
```
donaldgray, line 174: Tiny thing but there's nothing limiting this to 201 or 202, it creates with whatever status code is specified.

src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
```diff
@@ -131,9 +131,12 @@ await iiifS3.SaveIIIFToS3(mergedManifest, batch.Manifest!, $"{dlcsSettings.Prese
 
     private void CompleteBatch(Batch batch, DateTime finished)
     {
-        batch.Processed = DateTime.UtcNow;
+        var processed = DateTime.UtcNow;
+        
+        batch.Processed = processed;
         batch.Finished = finished;
         batch.Status = BatchStatus.Completed;
+        batch.Manifest!.LastProcessed = processed;
```
donaldgray, line 139: It looks like batch 1 of n could mark the manifest as `LastProcessed` here? Rather than only the final batch doing so?

src/IIIFPresentation/Models/Database/Collections/Manifest.cs
```diff
@@ -47,6 +47,11 @@ public class Manifest : IHierarchyResource
     public List<CanvasPainting>? CanvasPaintings { get; set; }
     
     public List<Batch>? Batches { get; set; }
+    
+    /// <summary>
+    /// Whether the manifest has been ingested with assets at some point
```
donaldgray, line 52: Update comment

donaldgray (approved):

PR #246, opened 2025-01-28
Fix SeeAlso links for api storage collections

### LINKED ISSUE #168: Fix SeeAlso links for flat path GET
`"SeeAlso"` links are added but hardcoded, they need to correctly render relevant links. 

From [documentation](https://deploy-preview-3--dlcs-docs.netlify.app/api-doc/iiif#seealso-links)
> The API Storage Collection has one or two seeAlso links, that point at the canonical public equivalent collection (but only if the Storage Collection has the public-iiif behavior), and also at the hierarchical form of the API version

So, we should render:
* canonical public collection, only if collection has `public-iiif`
* hierarchical form, with `"profile": [ "api-hierarchical" ]`

donaldgray:
Fixes #168 

Add `"seeAlso"` to storage collections. Which values are shown is dependant on configuration, as noted in docs there is redundancy in these links until canonical paths are implemented.

### DIFF
--- a/src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
+++ b/src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
@@ -66,6 +66,38 @@ public class CollectionConverterTests
         hierarchicalCollection.Items!.Count.Should().Be(1);
         hierarchicalCollection.Context!.Should().Be("http://iiif.io/api/presentation/3/context.json");
     }
+
+    [Fact]
+    public void ToPresentationCollection_NoSeeAlso_IfIIIFCollection()
+    {
+        // Arrange
+        var collection = CreateTestHierarchicalCollection(false, true);
+
+        // Act
+        var presentationCollection =
+            collection.ToPresentationCollection(PageSize, 1, 1, CreateTestItems(), null, pathGenerator);
+
+        // Assert
+        presentationCollection.SeeAlso.Should().BeNull();
+    }
+    
+    [Fact]
+    public void ToPresentationCollection_PublicAndApiSeeAlso_IfPublicStorageCollection()
+    {
+        // Arrange
+        var collection = CreateTestHierarchicalCollection(true, true);
+
+        // Act
+        var presentationCollection =
+            collection.ToPresentationCollection(PageSize, 1, 1, CreateTestItems(), null, pathGenerator);
+
+        // Assert
+        presentationCollection.SeeAlso.Should().HaveCount(2);
+        presentationCollection.SeeAlso![0].Id.Should().Be("http://base/1/top/some-id");
+        presentationCollection.SeeAlso[0].Profile.Should().Be("public-iiif");
+        presentationCollection.SeeAlso[1].Id.Should().Be("http://base/1/top/some-id");
+        presentationCollection.SeeAlso[1].Profile.Should().Be("api-hierarchical");
+    }
     
     [Fact]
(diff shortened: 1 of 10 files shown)


src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
```diff
@@ -107,11 +139,9 @@ public void ToPresentationCollection_ConvertsStorageCollection()
         presentationCollection.Label!.Count.Should().Be(1);
         presentationCollection.Label["en"].Should().Contain("repository root");
         presentationCollection.Slug.Should().Be("root");
-        presentationCollection.SeeAlso.Should().HaveCount(2);
+        presentationCollection.SeeAlso.Should().HaveCount(1);
```
griffri, line 142: Would it be worth using `...Should().ContainsSingle()` instead of `...Should().HaveCount(1)` here?

src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
```diff
@@ -107,11 +139,9 @@ public void ToPresentationCollection_ConvertsStorageCollection()
...
         presentationCollection.SeeAlso![0].Id.Should().Be("http://base/1");
-        presentationCollection.SeeAlso![0].Profile.Should().Contain("Public");
-        presentationCollection.SeeAlso![1].Id.Should().Be("http://base/1/iiif");
-        presentationCollection.SeeAlso[1].Profile.Should().Contain("api-hierarchical");
+        presentationCollection.SeeAlso[0].Profile.Should().Contain("api-hierarchical");
         presentationCollection.Created.Should().Be(DateTime.MinValue);
         presentationCollection.Parent.Should().BeNull();
         presentationCollection.Items!.Count.Should().Be(1);
```
griffri, line 147: I noticed that this uses `Items!.Count.Should().Be(1);` instead of `Items!.Should().HaveCount(1);` like some of the other assertions - is there a reason behind this?

donaldgray: Not sure - oversight I assume?

src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
donaldgray: The code is arguably cleaner with `.ContainsSingle()` but I think failing tests are better described with the existing checks as there are multiple conditions in the predicate

E.g. if "Profile" was `"foo"` the code + error would be:

```cs
presentationCollection.SeeAlso.Should()
            .ContainSingle(sa => sa.Id == "http://base/1" && sa.Profile.Contains("api-hierarchical"));
```

> Expected presentationCollection.SeeAlso to contain a single item matching (sa.Id == "http://base/1") AndAlso sa.Profile.Contains("api-hierarchical"), but no such item was found

```cs
presentationCollection.SeeAlso.Should().HaveCount(1);
presentationCollection.SeeAlso![0].Id.Should().Be("http://base/1");
presentationCollection.SeeAlso[0].Profile.Should().Contain("api-hierarchical");
```

> Expected presentationCollection.SeeAlso[0].Profile "foo" to contain "api-hierarchical"

src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
donaldgray: Replaced usages of `.Count.Should().Be()` with `.Should().HaveCount()`

Issue #279, opened 2025-02-18
Consistent approach to `CanvasPainting`


donaldgray:
Noticed when implementing #231

There are slight inconsistencies in approach to generating `CanvasPaintings` in the `CanvasPaintingResolver` class depending on whether the incoming manifest contains `"items"` or `"paintedResources"`. 

_Note the following flow may be out of date now. Work done in #316 and #340 will have tidied this code somewhat_

For `"items"`: The `"items"` are passed to `ManifestItemsParser` class which iterates through all of them and returns an `IEnumerable<CanvasPainting>`, where the `CanvasPaintings` are "partial" (they only have what we could derive from the manifest, no customer or manifest id etc). These are then returned to `CanvasPaintingResolver` which then iterates over them, generates and assigns canvasIds etc. All `CanvasPainting` object have canvas + choice order etc so identifying those sharing canvasId is simplified.

For `"paintedResources"`: all handling happens in `CanvasPaintingResolver`. The `"paintedResources"` property is iterated over and 'full' `CanvasPainting` objects are generated in 1 pass. The canvas + choice orders are only set if they've been provided so the logic to identify those sharing canvasId is a little more complex (and lead to a bug where `canvasPainting` objects on the same manifest with the same canvasOrder could have differing canvasId).

We should have a single approach to processing `CanvasPaintings`. This will be of more importance when we can handle both `"items"` _and_ `"paintedResources"`, rather than _or_. Initial suggestion is that both should pass to `ICanvasPaintingBuilder` (e.g. `ManifestItemsParser : ICanvasPaintingBuilder`) which can return "partial" classes. The logic for iterating over and assigning ids etc should hopefully then be simplified.

JackLewis-digirati:
**NOTE:** When PUTting assets, splitting a choice apart is not supported, this needs to be looked at as part of this ticket to enable

`UpdateCanvasPaintingRecords` currently searches for asset based candidates using the `assetId`, which will only have a match of 1 item, it should possibly use `canvasOrder` instead which can then correctly match choice constructs for setting values

donaldgray:
note on above comment - should be resolved by #340

Issue #227, opened 2025-01-16
Expand payload returned on create


donaldgray:
When a manifest is created with `"asset"` property we should return a default `"items"` prop as part of the 201 response, as we know the overall shape of what will be returned (if not the exact details). This will enable an API consumer to get immediate feedback on what items will be created, rather than needing to wait until the asset processing is complete.

Example of a minimal `"items"` property for a single "image per canvas" object:

```json
"items": [{
  "id": "{pres-uri}/{cust}/canvas/{canvas_id}",
  "type": "Canvas",
  "items": [{
    "id": "{pres-uri}/{cust}/canvas/{canvas_id}/annopages/{canvas_order}",
    "type": "AnnotationPage",
    "items": [{
      "id": "{pres-uri}/{cust}/canvas/{canvas_id}/annotations/{canvas_order}"
      "type": "Annotation",
      "motivation": "painting",
      "target": "{the-canvas-id}",
      "behavior": ["processing"],
      "body": null // or missing if iiif-net rolls that way
      }]
    }]
}]
```

Note the painting annotation introduces a new `"processing"` behavior.

The above is the minimal response, and could contain any of the fields that can be provided for `"canvasPainting"` (e.g. `thumbnail`, `canvas` and `canvas_label`) as well as choices etc. It's essentially a IIIF projection of what we have in the `canvasPaintings` db table.

This default `"items"` property is persisted to S3.

The `Canvas`, `AnnotationPage` and `Annotation` properties have Ids new minted:
* `Canvas` - use `"{pres-uri}/{cust}/canvases/{canvas_id}"` where canvas-id is the value from DB (this should match the `"id"` of the `canvasPainting` object)
* `AnnotationPage` - use `"{pres-uri}/{cust}/canvases/{canvas_id}/annopages/{canvas_order}"` (this will change in subsequent tickets)
* `Annotation` - use `"{pres-uri}/{cust}/canvases/{canvas_id}/annotations/{canvas_order}"` (this will change in subsequent tickets)

donaldgray:
For reference, the DLCS `"items"` generation happens here: https://github.com/dlcs/protagonist/blob/main/src/protagonist/Orchestrator/Infrastructure/IIIF/IIIFCanvasFactory.cs#L51

This contains more than we need for this but the overall shape will be the same.

PR #280, opened 2025-02-18
Output presentation ids on generated manifests

### LINKED ISSUE #231: Merge incoming NQ canvases into template `"items"`
> [!Warning]
> Blocked by #227 

Ticket #227 saves a template `"items"` property to S3. This contains `"id"` properties for `Canvas`, `AnnotationPage` and painting `Annotation`. 

These identifiers should be maintained in the final manifest - when we receive the output of the Named Query we want to persist these identifiers as they may be tracked and/or we may make them dereferenceable at a later date.

donaldgray:
Resolves #231 - ultimately outputting iiif-presentation ids for Canvases, AnnotationPages and PaintingAnnotations on Manifests generated from paintedResources/assets.

The main change is in `BatchCompletionMessageHandler`. When building `Dictionary<AssetId, Canvas>` from NamedQuery the `Canvas`, `AnnotationPages` and `PaintingAnnotations` ids are updated to IIIF-Presentation values. This seemed the simplest place, rather than peppering the manifest-merger with that logic .

> [!NOTE]
> Some of the changes here might look too wide-ranging for this ticket. This is because this ticket, #231, followed on from #227. The expectation was that the latter was writing the provisional canvases to S3 and this ticket was to persist identifiers in the final generated manifests. However, implementation of #227 was generating the provisional manifests when requested and not persisting to S3. Initially I updated the code to write these to S3 but ran into issues in background-handler when merging and reverted a load of changes but left some in where they'll help later when we come to save provisional items to S3.

Altered `ManifestConverter` to generate both `PaintedResources` and `Items` properties from `CanvasPaintings`. This removed the non-obvious precedence order between these 2 properties being set and will hopefully make them easier to tease apart in the future.

`ManifestConverter` was generating the required `id` properties manually but moved out to `IPathGenerator` so that these can be shared by background-handler.

`CanvasPaintingResolver.CreateCanvasPaintingsFromAssets` changes:
* It will now throw if it can't find `id` or `space` for an asset. Previously it would have thrown if it couldn't find `id` but attempted to update `space` if not found. Removed as `ManifestWriteService` is already doing this and changing the assets here was non-obvious.
* There was a bug in `CreateCanvasPaintingsFromAssets` where `CanvasPainting` records with the same `CanvasOrder` could end up with the wrong `CanvasId`. I fixed in that method but this logic is elsewhere in the class we should come back and have a single approach (#279 tracks this).

Refactored `ManifestWriteService` and extracted the DLCS interactions from it as it was doing too much. These are now in `DlcsManifestCoordinator`. 

Added `AssetProperties` to encapsulate known asset properties that are accessed.


src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -0,0 +1,92 @@
...
+    IDlcsApiClient dlcsApiClient,
+    PresentationContext dbContext,
+    ILogger<DlcsManifestCoordinator> logger)
+{
+    /// <summary>
+    /// Carry out any required interactions with DLCS for given <see cref="WriteManifestRequest"/>, this can include
+    /// creating a space and/or creating DLCS batches
+    /// </summary>
+    /// <returns>Tuple of any errors encountered and new Manifest SpaceId</returns>
+    public async Task<(ModifyEntityResult<PresentationManifest, ModifyCollectionType>? error, int? space)> HandleDlcsInteractions(WriteManifestRequest request,
```
p-kaczynski, line 26: On one hand I like and use tuples, but this is getting... unwieldy. Perhaps aliasing the type could be used? Not sure if we are on/can jump to C#12, but then you can do `using Point = (int x, int y);` :D

p-kaczynski (approved):
Love the `AssetProperties`, good call on that. Slightly worried we have now about seven classes that "do sth with manifest" but it's probably better than having a "god class" that does everything (i.e. it is what it is :D)

Minor comment on some tuple creep but that's purely aesthetic really. Well, that and I don't have any large screens to sometimes see even the signature of a method :D

src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
```diff
@@ -104,11 +94,48 @@ await dlcsOrchestratorClient.RetrieveAssetsForManifest(batch.CustomerId, batches
...
+                var canvasPaintingForAsset = canvasPaintings.FirstOrDefault(cp => cp.AssetId == assetId);
+                if (canvasPaintingForAsset == null)
+                {
+                    logger.LogWarning("Unable to find CanvasPainting record for manifest {CustomerId}:{ManifestId}",
+                        dbManifest.CustomerId, dbManifest.Id);
+                    continue;
+                }
+
+                // Set all NQ Canvas ids to those that we use in iiif-presentation
+                logger.LogWarning("Rewriting NQ ids for {CanvasId}, from manifest {CustomerId}:{ManifestId}",
```
JackLewis-digirati, line 124: should this be a warning? Seems like it should be info

JackLewis-digirati (approved):

src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
donaldgray: Yeah, nice catch. Copy/paste error. Will make debug as it could get really verbose for larger items.

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
donaldgray: Hmm, think I just used refactoring tools and left it as is but agree it's messy. Added `DlcsInteractionResult` to encapsulate result, eased creating and tidied up.

PR #287, opened 2025-02-20
Allow status code to be passed in a DLCS exception

### LINKED ISSUE #257: Presentation incorrectly returns `500` for manifests containing duplicate asset/canvas IDs
Spotted by @donaldgray in iiif-presentation-tests PR [#32](https://github.com/dlcs/iiif-presentation-tests/pull/32), regarding a test that ensures that Presentation will reject manifests containing duplicate IDs (both canvas and asset):
> This seems like a bug, looking at `isDlcsExceptionError` it checks for status code 500 but that feels incorrect - it should be a 400 BadRequest as the user can resolve by de-duplicating the Ids.

Presentation should be updated to return `400` instead in these situations.

JackLewis-digirati:
Resolves #257

This PR allows a status code to be passed out from a DLCS exception, which can then be used by various callers to correctly set a response status code

### DIFF
--- a/src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
@@ -103,7 +103,18 @@ public class DlcsManifestCoordinator(
             logger.LogError(exception, "Error creating batch request for customer {CustomerId}, manifest {ManifestId}",
                 customerId, manifestId);
             return EntityResult.Failure(exception.Message, ModifyCollectionType.DlcsException,
-                exception.StatusCode == HttpStatusCode.BadRequest ? WriteResult.BadRequest : WriteResult.Error);
+                ErrorStatusCodeToWriteResult(exception.StatusCode));
         }
+    }
+
+    private WriteResult ErrorStatusCodeToWriteResult(HttpStatusCode statusCode)
+    {
+        return statusCode switch
+        {
+            HttpStatusCode.BadRequest => WriteResult.BadRequest,
+            HttpStatusCode.NotFound => WriteResult.NotFound,
+            HttpStatusCode.Conflict => WriteResult.Conflict,
+            _ => WriteResult.Error
+        };
     }
 }
--- a/src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
@@ -1,4 +1,5 @@
-﻿using API.Features.Storage.Helpers;
+﻿using System.Net;
+using API.Features.Storage.Helpers;
 using API.Helpers;
 using Core;
 using DLCS.API;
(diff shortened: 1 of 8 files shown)


src/IIIFPresentation/DLCS/Exceptions/DlcsException.cs
```diff
@@ -1,15 +1,22 @@
...
+
 namespace DLCS.Exceptions;
 
 /// <summary>
 /// Represents an error interacting with DLCS
 /// </summary>
 public class DlcsException : Exception
 {
-    public DlcsException(string? message) : base(message)
+    public HttpStatusCode? StatusCode { get; }
```
donaldgray, line 11: Can this + the 2 ctor params be non-nullable? Looks like it's always set and would save null check/coalescing when verifying response

src/IIIFPresentation/DLCS/API/DlcsHttpContent.cs
```diff
@@ -58,7 +58,7 @@ public static StringContent GenerateJsonContent<T>(T body)
         }
         catch (JsonReaderException jre)
         {
-            throw new DlcsException("Error reading DLCS response", jre);
+            throw new DlcsException("Error reading DLCS response", jre, response.StatusCode);
```
donaldgray, line 61: This method starts

```cs
if (!response.IsSuccessStatusCode)
  throw await CheckAndThrowResponseError(response, cancellationToken);
```

so to hit here `response.IsSuccessStatusCode` so `response.StatusCode` would be a 2xx, which feels odd to be in an exception. Should we set a StatusCode (500?) or leave it `null` (which conflicts with comment I added about `StatusCode?` in `DlcsException` but maybe this is why we should allow null).

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -102,7 +103,7 @@ private static List<JObject> GetAssetJObjectList(WriteManifestRequest request) =
             logger.LogError(exception, "Error creating batch request for customer {CustomerId}, manifest {ManifestId}",
                 customerId, manifestId);
             return EntityResult.Failure(exception.Message, ModifyCollectionType.DlcsException,
-                WriteResult.Error);
+                exception.StatusCode == HttpStatusCode.BadRequest ? WriteResult.BadRequest : WriteResult.Error);
```
donaldgray, line 106: Maybe not required here but wonder if we need an `HttpStatusCode` -> `WriteResult` converter? We have an approximation of the opposite way in `ControllerBaseX.ModifyResultToHttpResult` but it converts to a full `IActionResult`.

e.g. can batch creation return a 409? If so it'd be good to return that rather than a 500

donaldgray (approved):

griffri (approved):

src/IIIFPresentation/DLCS/API/DlcsHttpContent.cs
JackLewis-digirati: yeah, would like be best to set it to 500 - good catch!

PR #296, opened 2025-02-24
Set the thumbnail correctly in a choice order

### LINKED ISSUE #258: Canvases that contain a choice do not contain a `thumbnail` object
(I'm unsure what the correct behaviour should be, so I'm flagging this)

If you POST a IIIF manifest containing painted resources, these will be returned as items in the processed manifest.
Standalone painted resources will be translated into items with a "thumbnail" object, like so:

```
   ... 
   "items": [
        {
            "id": "https://dlcs.digirati.io/iiif-img/52/988/my-image-0.tiff_NQ3Dhi/canvas/c/1",
            "type": "Canvas",
            "label": {
                "en": [
                    "My Image"
                ]
            },
            "width": 3903,
            "height": 3786,
            "thumbnail": [
                {
                    "id": "https://dlcs.digirati.io/thumbs/52/988/my-image-0.tiff_NQ3Dhi/full/200,194/0/default.jpg",
                    "type": "Image",
                    "format": "image/jpeg",
                    ...
   ... 
```
But painted resources that are part of a choice don't have this object in their equivalent item.

JackLewis-digirati:
Resolves #258 

This PR allows the canvas thumbnail to be set in a `choice` construct from a named query

### DIFF
--- a/src/IIIFPresentation/BackgroundHandler.Tests/Helpers/ManifestMergerTests.cs
+++ b/src/IIIFPresentation/BackgroundHandler.Tests/Helpers/ManifestMergerTests.cs
@@ -193,11 +193,14 @@ public class ManifestMergerTests
         // should be 1 + 3 then 4 + 2 then 5
         mergedManifest.Items[0].Id.Should().Be(canvas0Choice0);
         mergedManifest.Items[0].Label.Keys.Should().Contain("canvasPaintingCanvasLabel");
+        mergedManifest.Items[0].Thumbnail[0].Id.Should().Be($"{canvas0Choice0}_CanvasThumbnail");
         mergedManifest.Items[1].Id.Should().Be(canvas1Choice0);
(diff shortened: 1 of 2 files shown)


src/IIIFPresentation/BackgroundHandler.Tests/Helpers/ManifestMergerTests.cs
```diff
@@ -370,6 +377,14 @@ private static Canvas GenerateCanvas(string id, bool paintingChoice = false)
             Width = 110,
             Height = 110,
             Metadata = GenerateMetadata(),
+            Thumbnail = [
```
donaldgray, line 380: It looks like this `GenerateCanvas()` is used for every test. We should have some tests that use a canvas without a Thumbnail as that's a valid NQ manifest, will definitely be rarer but is possible.

Issue #232, opened 2025-01-16
Handle canvasPainting `static_width` and `static_height`


donaldgray:
If a `paintedResource` is specified alongside an `"asset"` the `CanvasPainting.StaticWidth` and `CanvasPainting.StaticHeight` properties are always populated by reading the w + h of the Canvas from the NQ response. 

We should update this to read the w + h of the `ImageService` (`"body"`) property, rather than the full image. The `ImageService` contains w + h of a known full image, which isn't necessarily the full size of the image - it allows for a smaller derivative to be used for a large image (e.g. a 200x200 for a source image of 5000x5000).

In addition to changing where the value is read from we should allow the `static_width` and `static_height` property to be specified. If specified we store the value in the db and update the [IIIF ImageApi size parameter](https://iiif.io/api/image/3.0/#42-size) for the image body.

Example, given the following (invalid) canvas from a NQ response:

```json
{
  "type": "Canvas",
  "width": 3648,
  "height": 5472,
  "items": [{
    "type": "AnnotationPage",
    "items": [{
      "type": "Annotation",
      "body": {
        "id": "https://dlcs.digirati.io/iiif-img/1/2/foo/full/683,1024/0/default.jpg",
        "type": "Image",
        "width": 683,
        "height": 1024,
        "format": "image/jpeg",
        "service": [{
         "id": "https://dlcs.digirati.io/iiif-img/1/2/foo",
         "type": "ImageService3",
```

We would
* If no `static_width` or `static_height` specified in `"canvasPainting"` block:
  * Store  `static_width = 683`, `static_height = 1024` in the database (read from body)
  * Store body with `{"id": "https://dlcs.digirati.io/iiif-img/1/2/foo/full/683,1024/0/default.jpg", "width": 683, "height": 1024 }`
* If `static_width = 67` and `static_height = 100` specified in `"canvasPainting"` block:
  * Store  `static_width = 67`, `static_height = 100` in the database (ignore values in NQ)
  * Store body with `{"id": "https://dlcs.digirati.io/iiif-img/1/2/foo/full/67,100/0/default.jpg", "width": 67, "height": 100 }`
* If only `static_width` OR `static_height` specified in `"canvasPainting"` block, reject with 400 response.

PR #299, opened 2025-02-24
Impl #232 - static image width/height handling

p-kaczynski:
Implements changes described in #232 

---

- **Use `ImageService` w+h instead of `Canvas` w+h**
- **Validate that `static_width` is paired with `static_height` for `canvasPainting`**
- **Add `GetModifiedImageRequest` to `PathGenerator`**
- **In `BatchCompletionMessageHandler` handle static image sizes**
- **Add tests to cover #232 changes**
- **Handle fully qualified uris in `GetModifiedImageRequest`**

---

Code includes explanatory comments because it is neither obvious nor trivial.

To modify the `ImageRequest` size one has to know customer id and space id. While customer id is provided, space id is in theory optional.

While possible to manually deduce the space id from the existing image id, it feels like it is a bad idea to me, hence the image resizing will only happen if we have `SpaceId` on the manifest.

### DIFF
--- a/src/IIIFPresentation/API/Features/Manifest/Validators/PresentationManifestValidator.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/Validators/PresentationManifestValidator.cs
@@ -48,5 +48,15 @@ public class PresentationManifestValidator : AbstractValidator<PresentationManif
                 }))
             .When(f => !f.PaintedResources.IsNullOrEmpty() && !f.PaintedResources.Any(pr => pr.CanvasPainting == null))
             .WithMessage("'choiceOrder' cannot be a duplicate within a 'canvasOrder'");
+
+        RuleFor(f => f.PaintedResources)
+            .Must(lpr =>
+                lpr!.All(
+                    // either both have value, or none
+                    pr => pr.CanvasPainting!.StaticHeight.HasValue == pr.CanvasPainting.StaticWidth.HasValue))
+            .When(f => !f.PaintedResources.IsNullOrEmpty() && f.PaintedResources.All(pr => pr.CanvasPainting != null))
+            .WithMessage(
+                "'static_width' and 'static_height' have to be both set or both absent within a 'canvasPainting'");
+
     }
 }
--- a/src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
+++ b/src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
@@ -5,6 +5,7 @@ using BackgroundHandler.Helpers;
 using Core.Helpers;
 using Core.IIIF;
 using DLCS.API;
+using IIIF.ImageApi.V3;
 using IIIF.Presentation.V3;
 using IIIF.Presentation.V3.Content;
 using Microsoft.EntityFrameworkCore;
@@ -184,8 +185,14 @@ public class BatchCompletionMessageHandler(
             canvasPainting.Thumbnail = thumbnailPath != null ? new Uri(thumbnailPath) : null;
             canvasPainting.Ingesting = false;
             canvasPainting.Modified = DateTime.UtcNow;
-            canvasPainting.StaticHeight = item.Height;
-            canvasPainting.StaticWidth = item.Width;
+
+            //#232 Use ImageService - leaving previous values as fallback, seems safer - PK
+            var (height, width) = item.Service?.OfType<ImageService3>().FirstOrDefault()
+                is { } imageService
+                ? (imageService.Height, imageService.Width)
+                : (item.Height, item.Width);
+            canvasPainting.StaticHeight = height;
+            canvasPainting.StaticWidth = width;
         }
     }
 
--- a/src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
+++ b/src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
@@ -196,7 +196,7 @@ public class BatchCompletionMessageHandler(
                 // if canvas painting comes with statics, use them
                 // otherwise, if we can find dimensions within, use those
                 // ReSharper disable once InvertIf
-                if (GetCanvasDimensions(item) is var (width, height))
+                if (item.GetCanvasDimensions() is var (width, height))
                 {
                     canvasPainting.StaticWidth = width;
                     canvasPainting.StaticHeight = height;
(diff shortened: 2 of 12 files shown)


src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
```diff
@@ -184,11 +189,87 @@ private void UpdateCanvasPaintings(Batch batch, Dictionary<AssetId, Canvas> item
...
+                // ReSharper disable once InvertIf
+                if (GetCanvasDimensions(item) is var (width, height))
+                {
+                    canvasPainting.StaticWidth = width;
+                    canvasPainting.StaticHeight = height;
+                }
+            }
+            else if (canvasPainting is {StaticWidth: { } staticWidth, StaticHeight: { } staticHeight}
+                     && item.Items?.GetFirstPaintingAnnotation()?.Body is Image image
+                     && manifest.SpaceId.HasValue) // required for ImageRequest parsing and modification
```
donaldgray, line 207: I don't think `&& manifest.SpaceId.HasValue` is correct. 

I could submit X assets to a manifest and specify the space for all of them - in that case the manifest wouldn't have a space. 
OR I could submit X assets and all but 1 have a space so the manifest would have a space but it would only be relevant for the 1 without.

`canvasPainting.AssetId` is the best place to read space from, it's an `AssetId` construct which has `Space` as part of it - this class mimics what's used internally in Protagonist.

src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
```diff
@@ -184,11 +189,87 @@ private void UpdateCanvasPaintings(Batch batch, Dictionary<AssetId, Canvas> item
...
+            {
+                // #232: if static_width/height provided in canvasPainting
+                // then don't use ones from NQ
+                // and set the body to those dimensions + resized id (image request uri)
+
+                image.Width = staticWidth;
+                image.Height = staticHeight;
+
+                image.Id = pathGenerator.GetModifiedImageRequest(image.Id, manifest.CustomerId,
+                    manifest.SpaceId.Value, staticWidth, staticHeight);
```
donaldgray, line 217: As mentioned above, `manifest.SpaceId` could result in incorrect path if the asset is in a different space from the manifests assigned space.

src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
```diff
@@ -184,11 +189,87 @@ private void UpdateCanvasPaintings(Batch batch, Dictionary<AssetId, Canvas> item
...
+            }
         }
     }
 
+    private static (int width, int height)? GetCanvasDimensions(Canvas canvas)
```
donaldgray, line 222: Worth moving these to a reusable utility function, there is a `ManifestX` somewhere (where `GetFirstPaintingAnnotation()` etc are). Would allow for testing in isolation.

src/IIIFPresentation/Repository/Paths/IPathGenerator.cs
```diff
@@ -99,4 +99,6 @@ Uri GenerateFlatCollectionViewLast(Collection collection, int lastPage, int page
     /// Generate the hierarchical id for specified customer and path slugs 
     /// </summary>
     string GenerateHierarchicalFromFullPath(int customerId, string? fullPath);
+
+    string? GetModifiedImageRequest(string? existing, int customerId, int spaceId, int width, int height);
```
donaldgray, line 103: Missing XML comment, everything else in class has brief one explaining what's being returned

src/IIIFPresentation/Repository/Paths/PathGeneratorBase.cs
```diff
@@ -99,6 +100,22 @@ public string GeneratePaintingAnnotationId(CanvasPainting canvasPainting)
         return uriBuilder.Uri;
     }
 
+    public string? GetModifiedImageRequest(string? existing, int customerId, int spaceId, int width, int height)
```
donaldgray, line 103: Missing tests in `PathGeneratorTests`

src/IIIFPresentation/API/Features/Manifest/Validators/PresentationManifestValidator.cs
```diff
@@ -48,5 +48,15 @@ public PresentationManifestValidator(IOptions<ApiSettings> options)
                 }))
             .When(f => !f.PaintedResources.IsNullOrEmpty() && !f.PaintedResources.Any(pr => pr.CanvasPainting == null))
             .WithMessage("'choiceOrder' cannot be a duplicate within a 'canvasOrder'");
+
+        RuleFor(f => f.PaintedResources)
```
donaldgray, line 52: Missing tests, `PresentationManifestValidatorTests`

donaldgray (changes requested):
I think part of the implementation is missing (or I'm missing it):

> In addition to changing where the value is read from we should allow the static_width and static_height property to be specified

Locally I tested by including the following in body but no `static_width` or `static_height` is stored:

```json
"canvasPainting": {
    "canvasOrder": 1,
    "staticWidth": 100,
    "staticHeight": 200
}
```

src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
p-kaczynski: Aaah there it is. Thanks - I'd have asked about that anyway today :D

src/IIIFPresentation/API/Features/Manifest/Validators/PresentationManifestValidator.cs
p-kaczynski: Added new tests

src/IIIFPresentation/Repository/Paths/PathGeneratorBase.cs
p-kaczynski: Added new tests

src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
p-kaczynski: Switched to using `AssetId` values`

p-kaczynski:
Found that indeed in one place the static width and height were not carried over to the db stored entity. Added along with the rest of feedback changes.

donaldgray (approved):

PR #303, opened 2025-02-26
Add migration to set last_processed where it shouldn't be null

### LINKED ISSUE #300: `last_processed` being null causes a manifest to return as 404
This is primarily an issue with old manifests that were created before the `last_processed` column in the manifest table.  This causes them to return as a 404 due to this check - https://github.com/dlcs/iiif-presentation/blob/main/src/IIIFPresentation/API/Features/Manifest/Requests/GetManifestHierarchical.cs#L32

This should be a fix to the migration that creates the `last_processed` column to default the value of `last_processed`, rather than allow it to be null

JackLewis-digirati:
Resolves #300 

This PR adds a migration to set `last_processed` where it shouldn't be null


donaldgray (approved):
Is it worth adding a comment to add some context to the script?

Issue #297, opened 2025-02-24
Log deserialisation errors


donaldgray:
There are a couple of helpers methods for deserializing IIIF payloads. These swallow exceptions but we should log the exception to aid debugging.

griffri:
Deserialisation errors are now logged as expected - moving to "done".

PR #304, opened 2025-03-02
#297 - error logging and code clean up

p-kaczynski:
Split into thematic commits:

- **Add extra analyzers to find incorrect exception handling**
- **Log exceptions before returning**
- **Observe the result**
- **Clean up usings**
- **misc code cleanup**

Set analyzer that checks few common issues and emits warnings; due to large number of CSC warnings it was hard to see, so I cleaned up the "numerous but not requiring functional changes". Also applied fixes for warnings from the analyzer (observing the result)


donaldgray (approved):
TIL `ErrorProne.NET.CoreAnalyzers` 👍

PR #305, opened 2025-03-03
Allow canvas id to be set via the canvas id, rather than auto-generate

JackLewis-digirati:


src/IIIFPresentation/API.Tests/Helpers/PaintedResourceXTest.cs
```diff
@@ -86,9 +86,9 @@ public void GetRequiredNumberOfCanvases_0_IfPaintedResourcesEmpty()
     public void GetRequiredNumberOfCanvases_Correct_IfAllPaintedResourcesNullCanvasOrder()
     {
         List<PaintedResource> paintedResources = [
-            new() { CanvasPainting = new CanvasPainting { CanvasId = "1" } },
-            new() { CanvasPainting = new CanvasPainting { CanvasId = "2" } },
-            new() { CanvasPainting = new CanvasPainting { CanvasId = "3" } },
+            new() { CanvasPainting = new CanvasPainting() },
```
donaldgray, line 89: There are now no longer any tests for `CanvasPainting` records that have `CanvasId` - seems like we should still have some?

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -235,6 +258,21 @@ string GetCanvasId(Models.API.Manifest.CanvasPainting? cp)
         }
     }
 
+    private string? GetCanvasId(Models.API.Manifest.CanvasPainting? canvasPainting, int customerId)
```
donaldgray, line 261: This seems suited to `PathParser` class? Keeps all path-parsing utils together and makes it easier to test etc

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -211,13 +223,24 @@ public class CanvasPaintingResolver(
...
+                    null);
+            };
 
             count++;
             canvasPaintings.Add(cp);
         }
         return (null, canvasPaintings);
 
-        string GetCanvasId(Models.API.Manifest.CanvasPainting? cp)
+        string GetGeneratedCanvasId(Models.API.Manifest.CanvasPainting? cp)
```
donaldgray, line 243: Should the method of getting CanvasIds be unified? There's currently 2 different methods to getting canvasIds - `GetGeneratedCanvasId` which returns canvasId that we've minted and `GetCanvasId` which returns canvasId that we've been provided. 

Seems like it would be preferably to combine these methods into 1 that can return the appropriate canvasId regardless of whether it was provided or minted?

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -211,13 +223,24 @@ public class CanvasPaintingResolver(
                 StaticWidth = paintedResource.CanvasPainting?.StaticWidth,
                 StaticHeight = paintedResource.CanvasPainting?.StaticHeight
             };
+            
+            if (canvasPaintings.Where(c => c.CanvasOrder != cp.CanvasOrder).Any(c => c.Id == cp.Id))
```
donaldgray, line 227: Seems odd to do this validation per iteration? Could this be caught up front? If so would it simplify the different `GetCanvasId` methods (previous PR comment)?

src/IIIFPresentation/API/Helpers/PaintedResourceX.cs
```diff
@@ -16,7 +16,8 @@ public static bool HasAsset(this List<PaintedResource>? paintedResources)
     /// </summary>
     public static int GetRequiredNumberOfCanvases(this List<PaintedResource>? paintedResources)
     {
-        if (paintedResources.IsNullOrEmpty()) return 0;
+        // no need to generate a canvas id for canvases that have an id, as we'll use that
```
donaldgray, line 19: The code doesn't line up with this comment. This comment sounds like if we had 10 paintedResources and 4 had and id we'd still return 6 but that isn't happening.

Question is - _should_ it happen? The "all or nothing" for canvasId is MVP only and caught in validator - would it be simpler to implement as in the comment? Would mean less places need changed when we allow specifying canvasIds for some but not all?

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -235,6 +258,21 @@ string GetCanvasId(Models.API.Manifest.CanvasPainting? cp)
         }
     }
 
+    private string? GetCanvasId(Models.API.Manifest.CanvasPainting? canvasPainting, int customerId)
+    {
+        var canvasId = canvasPainting?.CanvasId;
+        
+        if (canvasId == null || !Uri.IsWellFormedUriString(canvasId, UriKind.Absolute)) return canvasId;
```
donaldgray, line 265: Would `canvasId` be returned here if it was `"foo/bar/baz"` ?

src/IIIFPresentation/API/Features/Common/Helpers/ErrorHelper.cs
```diff
@@ -74,6 +74,21 @@ public static ModifyEntityResult<TCollection, ModifyCollectionType> ParentMustBe
...
+            ModifyCollectionType.InvalidCanvasId, WriteResult.BadRequest);
+
+    public static ModifyEntityResult<TCollection, ModifyCollectionType>? DuplicateCanvasId<TCollection>(string? canvasId)
+        where TCollection : JsonLdBase
+        => ModifyEntityResult<TCollection, ModifyCollectionType>.Failure($"The canvas ID {canvasId} cannot be a duplicate",
+            ModifyCollectionType.DuplicateCanvasId, WriteResult.BadRequest);
+
+    public static ModifyEntityResult<TCollection, ModifyCollectionType>? CanvasOrderDifferentCanvasId<TCollection>(string? canvasId)
+        where TCollection : JsonLdBase
+        => ModifyEntityResult<TCollection, ModifyCollectionType>.Failure($"The canvas ID {canvasId} must be the same within a canvas order",
```
donaldgray, line 90: Not sure how but I think this error message could be worded to explain what's happening better

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
donaldgray: I think we should reject? Only allow a single, url-friendly value `3bhgfsdkl`. CanvasId will be derefenceable eventually so having additional `/` values might make that difficult and removes ability to just use the last slug as id

JackLewis-digirati: ah ok, sure - makes sense - I've modified the `PathParser` to reject these now

src/IIIFPresentation/Repository.Tests/Paths/PathParserTests.cs
```diff
@@ -38,4 +39,62 @@ public void GetAssetIdFromNamedQueryCanvasId_Throws_IfInvalidFormat(string incor
         act.Should().Throw<FormatException>()
             .WithMessage($"Unable to extract AssetId from {incorrectId}");
     }
+
+    [Theory]
+    [InlineData("https://dlcs.example/1/canvases/someId")]
+    [InlineData("someId")]
```
donaldgray, line 45: `[InlineData("https://dlcs.example/1/canvases/someId?foo=bar")]` fails this test, results in id of `someId?foo=bar`

src/IIIFPresentation/Repository/Paths/PathParser.cs
```diff
@@ -25,5 +25,31 @@ public static AssetId GetAssetIdFromNamedQueryCanvasId(this Canvas canvas, ILogg
             throw new FormatException($"Unable to extract AssetId from {canvas.Id}");
         }
     }
+    
+    public static string GetCanvasId(Models.API.Manifest.CanvasPainting? canvasPainting, int customerId)
+    {
+        var canvasId = canvasPainting?.CanvasId.ThrowIfNull(nameof(canvasPainting));
```
donaldgray, line 31: Looking at this, param should be `Models.API.Manifest.CanvasPainting canvasPainting`? Currently `Models.API.Manifest.CanvasPainting? canvasPainting` makes it look like it can accept a null `CanvasPainting` but that would result in an error being thrown

donaldgray (changes requested):

src/IIIFPresentation/Repository/Paths/PathParser.cs
```diff
@@ -25,5 +25,40 @@ public static AssetId GetAssetIdFromNamedQueryCanvasId(this Canvas canvas, ILogg
...
+
+        var convertedCanvasId = new Uri(canvasId).PathAndQuery;
+        var startsWith = $"/{customerId}/canvases/";
+
+        if (!convertedCanvasId.StartsWith(startsWith) || convertedCanvasId.Length == startsWith.Length)
+            throw new ArgumentException($"Canvas Id {convertedCanvasId} is not valid");
+
+        var actualCanvasId =
+            convertedCanvasId.Substring(startsWith.Length, convertedCanvasId.Length - startsWith.Length);
+        if (prohibitedCharacters.Any(pc => actualCanvasId.Contains(pc)))
```
donaldgray, line 51: Duplicate of lines 35-38 (but different as no braces). Create a method? Single place to manage this and the error message (e.g. if we switch from a black to a white list would involve updating in 1 place rather than 2).

donaldgray (approved):

Issue #271, opened 2025-02-13
Background Handler uses staged manifest


donaldgray:
Update background handler batch completed notification handler to read the base manifest from the staging location.

If the 'staging' location doesn't contain a manifest - log an error and stop processing. Don't delete the message so that it ends up in DLQ for investigation.

Write the final manifest to 'real' location.

Delete manifest from staging.

Issue #269, opened 2025-02-13
Manifest flat GET may read staging manifest


donaldgray:
If manifest is ingesting, read the manifest from the staging location.

If ingesting and manifest not in staging location, log an error and return from normal s3 location (something has went wrong if we end up in this scenario).

Else, return manifest from standard s3 location.

Ingesting is determined by `IsIngesting()`

PR #312, opened 2025-03-07
Resolves #268 #269 #271 - staging location for manifests

### LINKED ISSUE #268: Manifests written to staging location
For flat/api PUT and POST operations.

If manifest contains `"items"` the logic stays as-is and manifest is written to current s3 location.

If manifest has `"paintedResources"` then, the manifest is written to the 'staging' s3 location.

p-kaczynski:
- **#268 Manifests written to staging location**
- **#271 Background Handler manifest staging**
- **#269 Manifest ingesting read staging**

### DIFF
--- a/src/IIIFPresentation/API/Features/Manifest/ManifestReadService.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/ManifestReadService.cs
@@ -94,7 +94,8 @@ public class ManifestReadService(
                 logger.LogError("Manifest {DbManifestId} IsIngesting but can't read from staging", dbManifest.Id);
         }
 
-        // else or if null
+        // if is not ingesting read from "real" location
+        // or if not found in "staging", an error was logged and we fall back to "real"
         manifest ??= await iiifS3.ReadIIIFFromS3<PresentationManifest>(dbManifest, false, cancellationToken);
 
 
--- a/src/IIIFPresentation/API/Features/Manifest/ManifestReadService.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/ManifestReadService.cs
@@ -39,7 +39,7 @@ public class ManifestReadService(
     IPathGenerator pathGenerator,
     ILogger<ManifestReadService> logger) : IManifestRead
 {
-    public async Task<Dictionary<string, JObject>?> GetAssets(int customerId, Models.Database.Collections.Manifest? dbManifest,
+    public async Task<Dictionary<string, JObject>?> GetAssets(int customerId, DbManifest? dbManifest,
         CancellationToken cancellationToken)
     {
         var assetIds = dbManifest?.CanvasPaintings?.Select(cp => cp.AssetId?.ToString())
@@ -86,10 +86,9 @@ public class ManifestReadService(
         }
 
         var getAssets = GetAssets(customerId, dbManifest, cancellationToken);
-        var manifest = await iiifS3.ReadIIIFFromS3<PresentationManifest>(dbManifest, cancellationToken);
+        var manifest = await iiifS3.ReadIIIFFromS3<PresentationManifest>(dbManifest, false, cancellationToken);
         dbManifest.Hierarchy.Single().FullPath = await fetchFullPath;
 
-        // PK: Will this even happen? Should we log or even throw here?
         if (manifest == null)
             return FetchEntityResult<PresentationManifest>.Failure(
                 "Unable to read and deserialize manifest from storage");
--- a/src/IIIFPresentation/API/Features/Manifest/ManifestReadService.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/ManifestReadService.cs
@@ -86,7 +86,18 @@ public class ManifestReadService(
         }
 
         var getAssets = GetAssets(customerId, dbManifest, cancellationToken);
-        var manifest = await iiifS3.ReadIIIFFromS3<PresentationManifest>(dbManifest, false, cancellationToken);
+        PresentationManifest? manifest = null;
+        if (dbManifest.IsIngesting())
+        {
+            manifest = await iiifS3.ReadIIIFFromS3<PresentationManifest>(dbManifest, true, cancellationToken);
+            if (manifest == null)
+                logger.LogError("Manifest {DbManifestId} IsIngesting but can't read from staging", dbManifest.Id);
+        }
+
+        // else or if null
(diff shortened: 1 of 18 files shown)


src/IIIFPresentation/API/Features/Manifest/ManifestReadService.cs
```diff
@@ -86,10 +86,20 @@ public async Task<FetchEntityResult<PresentationManifest>> GetManifest(int custo
         }
 
         var getAssets = GetAssets(customerId, dbManifest, cancellationToken);
-        var manifest = await iiifS3.ReadIIIFFromS3<PresentationManifest>(dbManifest, cancellationToken);
+        PresentationManifest? manifest = null;
+        if (dbManifest.IsIngesting())
+        {
+            manifest = await iiifS3.ReadIIIFFromS3<PresentationManifest>(dbManifest, true, cancellationToken);
+            if (manifest == null)
+                logger.LogError("Manifest {DbManifestId} IsIngesting but can't read from staging", dbManifest.Id);
```
JackLewis-digirati, line 94: should `IsIngesting` be is ingesting?

src/IIIFPresentation/API/Features/Manifest/ManifestReadService.cs
```diff
@@ -86,10 +86,20 @@ public async Task<FetchEntityResult<PresentationManifest>> GetManifest(int custo
...
+        }
+
+        // else or if null
```
JackLewis-digirati, line 97: took me little bit to understand that the difference is that this one is reading from the "live" bucket - maybe this comment could be improved to say that?

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -202,7 +205,9 @@ private async Task<PresUpdateResult> UpdateInternal(UpsertManifestRequest reques
                 await UpdateDatabaseRecord(request, parentCollection!, existingManifest, cancellationToken);
             if (error != null) return error;
 
-            await SaveToS3(dbManifest!, request, cancellationToken);
+            var saveToStaging = request.PresentationManifest.Items.IsNullOrEmpty();
```
JackLewis-digirati, line 208: this might work now, but will it break when we start mixing items and assets?

src/IIIFPresentation/AWS/Settings/S3Settings.cs
```diff
@@ -9,9 +9,14 @@ public class S3Settings
     /// Name of bucket for storing ingested web-friendly assets
     /// </summary>
     public string StorageBucket { get; set; }
+
+    /// <summary>
+    ///     Name of the bucket used as a staging location
+    /// </summary>
+    public string StagingStorageBucket => $"staging-{StorageBucket}";
```
JackLewis-digirati, line 16: this looks like a new bucket is being used? Might be wrong, but I thought we were using a sub-path within the same bucket for the staging location? (Though I can see this isn't mentioned in the ticket)

If we are using a new setting, the `appsettings.Development.json` needs updating

JackLewis-digirati (changes requested):
I think there's some broken tests that need to be looked at

Additionally, the name of the PR should be meaningful as it's pulled through into release notes, and if you add `Resolves #<ticket number>` into the description, it will automatically close the ticket for you when the PR is merged

src/IIIFPresentation/AWS/Settings/S3Settings.cs
donaldgray: Not in ticket, should be added. 

`/staging/` prefix is suggested in https://github.com/dlcs/iiif-presentation/blob/develop/docs/notes/storage-keys.md#storage-keys. It's shown in [associated RFC](https://github.com/dlcs/iiif-presentation/blob/develop/docs/rfcs/0002-manifest-write-mvp.md) but only in sequence diagram.

p-kaczynski: True, was gonna have a chat about a need for a new bucket. I might be misremembering, as I think I asked about it (i.e. whether "staging" is a new bucket). I think it might be easier in the longer term, for stuff like removing orphans and such?

src/IIIFPresentation/API/Features/Manifest/ManifestReadService.cs
p-kaczynski: kinda purposeful to indicate how this was determined

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
p-kaczynski: Best what I can think of rn is adding a comment with a `TODO`, or for explicit method that returns bool, so the method can be later modified (and primarily located easier)

JackLewis-digirati: think a method would be a good idea, but would it make more sense to use something like `request.PresentationManifest.PaintedResources.Any(x => x.Asset != null);`?  As the assets are what we're explicitly checking for?

src/IIIFPresentation/AWS/Helpers/IIIFS3Service.cs
```diff
@@ -36,14 +36,16 @@ public class IIIFS3Service(
...
+        ReadIIIFFromS3<T>(dbResource.GetResourceBucketKey(fromStaging), cancellationToken);
 
     public async Task<T?> ReadIIIFFromS3<T>(string bucketKey,
         CancellationToken cancellationToken) where T : ResourceBase, new()
     {
-        var objectFromBucket = await bucketReader.GetObjectFromBucket(
-            new(options.CurrentValue.S3.StorageBucket, bucketKey), cancellationToken);
+        var item = new ObjectInBucket(options.CurrentValue.S3.StorageBucket, bucketKey);
+        var objectFromBucket = await bucketReader.GetObjectFromBucket(item
+            , cancellationToken);
```
JackLewis-digirati, line 48: feels strange to have a comma start the line

JackLewis-digirati (approved):

src/IIIFPresentation/AWS/Helpers/IIIFS3Service.cs
p-kaczynski: autoformatted tbh, will ignore this one in the name of getting the PR in, sorry :D

Issue #169, opened 2024-11-20
Cleanup DB entities


donaldgray:
There are a couple of columns in `Collections` table (e.g. `use_path` and `locked_by`) that were in the RFC but aren't actively used in the project. We should remove these - we may add them back in later but we should do it as they are required, rather than adding in up front. EF migrations will make this trivial.

Tidy namespaces for Entities, e.g. `Models.Database.Collections` contains both `Collection` and `Manifest`.

Consider removing any current unused fields (e.g. `hierarchy.canonical`).

Acceptance Criteria
* There are no functional changes (nothing should be removed that's exposed in a payload)
* There are no unused/redundant fields in entitities
* Entities are in correct namespaces

donaldgray:
due to wide ranging namespace changes we'll need to be careful when applying this to avoid merge issues

PR #314, opened 2025-03-10
Allow PUT to update root collection

### LINKED ISSUE #313: Allow root collection to be updated via PUT
Allow consumers to update root storage collection via PUT. Label, thumb and private/public status can be updated.

Validation rules for PUT to root that differ from other storage-collections:

* `slug` and `parent` cannot be changed (ie must be empty in payload. Treat `null` and empty as same)

donaldgray:
Resolves #313

Allow 'root' collection to be updated - only public/private status; label and thumbnail can be updated.

Validation rules differ from "normal" collections so added a separate ruleSet to fluentValidator to make this simple.

Simplified `PresentationCollection` by removing `PresentationThumbnail` property and overriding of `Thumbnail`. This resulted in a few broken test hence changes in `ModifyCollectionTests`

## Edit after merge to `develop`

Had to refactor `IParentSlugParser` to accommodate root collection validation


src/IIIFPresentation/API/Features/Storage/Validators/RootCollectionValidator.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        RuleFor(pc => pc.Parent)
+            .Empty()
+            .WithMessage("Cannot set 'parent' for root collection");
+
+        RuleFor(pc => pc.Slug)
+            .Empty()
+            .WithMessage("Cannot set 'slug' for root collection");
+
+        RuleFor(pc => pc.Behavior)
```
JackLewis-digirati, line 19: should there also be rules for `ItemsOrder` and `Type` as these also aren't allowed to be changed for the root collection?

src/IIIFPresentation/API/Helpers/ParentSlugParser.cs
```diff
@@ -19,121 +21,147 @@ namespace API.Helpers;
...
+            CancellationToken cancellationToken) where T : JsonLdBase
+    {
+        // Try and get a parent from publicId 
+        var publicIdParent= await GetParentFromPublicId(presentation, customerId, cancellationToken);
+
+        // If we don't have parent, return what we could parse from publicId 
+        if (presentation.Parent == null) return (null, publicIdParent);
+        
+        // We have Parent property - find Collection for that 
+        var parent= await RetrieveParentFromPresentation(presentation, customerId, cancellationToken);
```
JackLewis-digirati, line 127: missing space before `=`

src/IIIFPresentation/Models.Tests/Database/Collections/CollectionXTests.cs
```diff
@@ -0,0 +1,20 @@
...
+
+public class CollectionXTests
+{
+    [Theory]
+    [InlineData(null)]
+    [InlineData("")]
+    [InlineData(" ")]
+    [InlineData("not-root")]
+    public void IsRoot_False_IfNotRoot(string? id)
+        => new Collection { Id = id }.IsRoot().Should().BeFalse();
```
JackLewis-digirati, line 13: missed warning

JackLewis-digirati (approved):

src/IIIFPresentation/API/Features/Storage/Validators/RootCollectionValidator.cs
donaldgray: `Type` can't be set on it's own, it's value is determined by `Behavior`.

`ItemsOrder` - it looks like that's an erroneous field and isn't used for anything? Seems unnecessary to reject payloads that contain a field that isn't used. #169 suggests removing the field so any change would be temporary _or_ if we decide to keep it we can add rules around it at that point.

I opted to wrap things that change hierarchy values in [`if (!databaseCollection.IsRoot())`](https://github.com/dlcs/iiif-presentation/pull/314/files/d3d8f109c78428a94e5f5d8c31e0d0bbd9f2c6ec#diff-d1234e91efe969cbe7bf3d138b0dc01d03e8024f927e6d17412e9da6cd69eba2R128) within `UpsertCollection` so any values are ignored too

PR #321, opened 2025-03-17
Set body to be of type choice or null when creating provisional items

JackLewis-digirati:


src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
```diff
@@ -236,6 +248,54 @@ public void SetGeneratedFields_SetsCanvasPaintings_InCanvasThenChoiceOrder()
         result.PaintedResources.Should()
             .BeInAscendingOrder(pr => pr.CanvasPainting.CanvasOrder)
             .And.ThenBeInAscendingOrder(pr => pr.CanvasPainting.ChoiceOrder);
+
+        var items = JsonConvert.SerializeObject(result.Items, serializerSettings);
```
donaldgray, line 252: Should this be in a different test? 

The test is `SetGeneratedFields_SetsCanvasPaintings_InCanvasThenChoiceOrder`. The assertion at line 248 asserts what's in the title of this test (`InCanvasThenChoiceOrder`), which could pass but then the test could fail because of a serialisation change.

Or should we test this at all? Is this testing how a nuget package deserialises (`iiif-net`)?

donaldgray (approved):

PR #322, opened 2025-03-17
RFC for identity rewrites

donaldgray:
RFC for how we can output and parse rewritten ids. This is _not_ for identifying assets, only Manifest, Collection and Canvas.


docs/rfcs/0003-identity-rewrites.md
```diff
@@ -0,0 +1,274 @@
+# Identity Rewrites
+
+IIIF Presentation sets a number of identifiers on generated resources. Manifests and Collections can have different identifiers depending on whether they are access publicly, or via API paths.
```
JackLewis-digirati, line 3: `accessed`?

docs/rfcs/0003-identity-rewrites.md
```diff
@@ -0,0 +1,274 @@
...
+These are:
+
+* Manifests: `id` property. Public and api variations.
+  * Canonical public: `{customerId}/{hierarchyPath}` (e.g. `https://dlcs.example/123/path/to/item`)
+  * Canonical private: `{customerId}/manifests/{resourceId}` (e.g. `https://dlcs.example/123/manifests/internal-id`)
+* Collections: `id` property. Public and api variations
+  * Canonical public: `{customerId}/{hierarchyPath}` (e.g. `https://dlcs.example/123/path/to/item`)
+  * Canonical private: `{customerId}/collections/{resourceId}` (e.g. `https://dlcs.example/123/collections/internal-id`)
+* Canvases: `id` property. Single variation only for both public and api requests.
+  * Canonical `{customerId}/canvases/{canvasId}` (e.g. `https://dlcs.example/123/canvases/internal-id`)
```
JackLewis-digirati, line 20: while it's not direct, are we going to leave the identifiers in the asset as is? Assuming so given we're pointing to the DLCS for this

JackLewis-digirati (approved):

docs/rfcs/0003-identity-rewrites.md
donaldgray: Yeah, as mentioned in the PR description asset rewrites weren't included when I was writing this for this - it was only Manifest, Collection and Canvas.

However, it does raise an interesting question I hadn't considered - if DLCS is configured with rewrites rules for `canonical.dlcs` (`https://canonical.dlcs/iiif-img/2/1/foo`) and `alternative.dlcs` (`https://alternative.dlcs/images/foo`) - the generated Manifest will have been created using (as of now) the canonical version only, regardless of the path it's requested on.

Assuming IIIF Presentation has rewrite rules for `canonical.dlcs` and `alternative.dlcs` too, whether the manifest is requested on either of these paths we'd only be outputting assets for `canonical.dlcs`. Would this need to be addressed? Or is it okay for the assets to always reference one version? If so, should that version be controllable via config? `"AssetHost"` or something. Note that if we want to use something non-canonical then it'd required changes to Orchestrator as path templates are only applied for `info.json` ids (not NQ or single-item manifests).

cc @tomcrane for thoughts. Might be best to chat through when you're back.

tomcrane: I think we have to assume that, like Wellcome, customers who want rewrite rules are going to want pretty URLs for both IIIF-P resources and assets, and that there is a connection between the two - you're either going to want a Manifest that has the "native" URIs for both, or the rewritten URIs for both, because the one-of-each isn't much use.

If the IIIF-P has an AssetHost it somehow has to use that when asking IIIF-Asset (?) for asset details so that it gets assets back with their rewritten URL forms.

donaldgray: Added update in new commit to address asset ids. 

Discussed with @tomcrane and decision was to have a configurable `OrchestratorUrl` per customer

docs/rfcs/0003-identity-rewrites.md
```diff
@@ -0,0 +1,291 @@
...
+	var templates = new string[]{ altTemplate1, altTemplate2, altTemplate3 };
+	
+	// Sample incoming paths, 1 matching canonical and 1 per rewrite pattern
+	var canonical = "/99/manifests/123abc";
+	var alt1 = "/foo/just-manifest/123abc"; // no customerId
+	var alt2 = "/123abc/99"; // manifest + customer reversed
+	var alt3 = "/foo/99/diff/123abc"; // starts like alt1 but diff element and customerId
+	
+	// This would come from Route
+	const int customerId = 99;
```
tomcrane, line 207: ...but not present as "99" in the route: implied by variations of host name, e.g., at Leeds, 

https://dev-iiif.leeds.ac.uk/presentation/cc/abcdedfgh

"dev-iiif.leeds.ac.uk" => customerId = 2
"cc/abcdedfgh" => hierarchyPath = "cc/abcdefgh"  (happens to be a Manifest)


https://iiif.leeds.ac.uk/presentation/cc/mm44nn55

"iiif.leeds.ac.uk" => customerId = 9
"cc/mm44nn55" => hierarchyPath = "cc/mm44nn55"  (happens to be a Collection)

How does customerId get set as a route parameter?

This config is "at a customer level" but where does the global config live that tells us that iiif.leeds.ac.uk is customer 9 and what does that config look like? Does some overall thing scan all the customer path templates across the board and keep a global map of them?

Config that already belongs to Cust 9:
```
{
    "pathTemplate": {
        "iiif.leeds.ac.uk": {
          "ManifestPublic": "???",
          "ManifestPrivate": "???"
        },
        // etc
    }
}
```

JackLewis-digirati (approved):

docs/rfcs/0003-identity-rewrites.md
donaldgray: > How does customerId get set as a route parameter?

Will be rewritten in proxy sitting in front of IIIF-P. IIIF-P won't really know that it's been called on a different path, the proxy (lambda@edge in CloudFront) will have rewritten. The knowledge of how to do that sits there.

`dev-iiif.leeds.ac.uk/presentation/cc/abcdedfgh -> [CloudFront] -> iiif-p.leeds/99/cc/abcdedfgh`

CloudFront will also set the `x-forwarded-host`, which IIIF-P will use for pathTemplate lookup.

PR #334, opened 2025-03-21
Pass assets through to SetGeneratedFields

JackLewis-digirati:
Fixes an issue where assets weren't correctly passed through to the response.  This was due to assets not being correctly passed through to `SetGeneratedFields`


griffri (approved):
LGTM

PR #335, opened 2025-03-24
Etag RFC

### LINKED ISSUE #146: Improve ETag support RFC
The API currently handles ETags, with the process being:
* Resources that return etags are marked with `[ETagCaching]` attribute.
  * This is a `ActionFilterAttribute` that generates the ETag by hashing the response body.
  * After generation the ETag is cached in-memory by request path.
* On subsequent requests the above cache is checked:
  * If not found, the request is rejected
  * If non-matching, request is rejected
  * If matching, request is processed

This process needs to be revisited, salient points (some of these summarised from ticket comments):
* The ETag should be generated eagerly on PUT/POST/PATCH and stored against the resource.
* If ETags are cached in-memory, this must be invalidated when resource changes, bearing in mind this could happen on another node.
  * Should we consider using `IDistributedCache`? If so, what backend?
* How do we generate the checksum?
  * Is it a checksum, or just a random identifier we generate? 
  * Can we leverage S3 to generate [checksum](https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html) and store that?
    * StorageCollection doesn't have S3 representation so what would we store for this?
    * Does staging/non-staging areas come into it?
    * If a resource changes `parent` or `slug` the JSON representation wouldn't change as it contains the `flatId` only.
  * Do we need different checksum for API and Public view?
    * Database values could affect the resources (see `flatId` above, child counts etc too).
* #141 set `cache-control` header for ETag, does what we do here affect that?
* See [api-scratchpad](https://github.com/dlcs/iiif-presentation-tests/blob/04213f185bf4fb370855e7e37be27ee4587234bf/tests/apiscratch/t0071-update-managed-asset-manifest.spec.ts#L60-L62) tests that suggest only returning ETag if the resource is editable (ie if it's in-flight then not returning an ETag would mean it can't be changed).

This ticket is to write up an RFC addressing how we generate/store/process ETags.

----
_Original ticket below_

Following on from PR comment - https://github.com/dlcs/iiif-presentation/pull/1/files#r1723019033 - ETag handling needs to be revisited as we are now storing items in S3.

Also (will confirm) - do we need to improve how ETag content is generated and make it longer-lived than in a memory cache? Implementation looks like if the resource hasn't been fetched or created by current instance it won't be in memory-cache and so requests will fail. This ties up with code comments in tests

> TODO: remove this when better ETag support is implemented - this implementation requires GET to be called to retrieve the ETag

JackLewis-digirati:
Resolves #146

This PR adds an RFC around changes to the Etag


docs/rfcs/0003-etag-changes.md
```diff
@@ -24,54 +24,153 @@ While the current Etag implementation allows for easily adding and removing Etag
...
 
-Doing this would mean an additional value being added to the database called `etag` that allows for the Etag to be stored.  Given that there are multiple resource types that require updating (manifests and collections) with their own table, this column would likely be best going onto the shared `hierarchy` table to reduce duplication.  Another potential solution is to have another table containing the Etag, but this is probably overkill for thisd single resource.
+Doing this would mean an additional value being added to the database called `etag` that allows for the Etag to be stored.  Given that there are multiple resource types that require updating (manifests and collections) with their own table, while it would be possible to add a column onto the shared `hierarchy` table to reduce duplication, but due to the ability to create `canoncial` and `non-canonical` resources which would mean the Etag value is duplicated across the table.  Therefore, the preferered solution to this is to have another table containing the Etag, which would look like this:
 
-### Distrtibuted cache
+| name | type | constraints |
```
donaldgray, line 37: Why use a new table? Would `etag` not be better as a column on `Collection` and `Manifest`, rather than a separate table? `etag` is just another attribute of `Collection` or `Manifest` (like `Created` or `ModifiedBy`) rather than an entity (~table) in it's own right.

docs/rfcs/0003-etag-changes.md
```diff
@@ -24,54 +24,153 @@ While the current Etag implementation allows for easily adding and removing Etag
...
+
+It should be fairly obvious when the database begins to become resource starved and in this case, any form of caching would improve the situation.
+
+#### Distributed cache
+
+This change would require that a call is made to the database to retrieve the Etag value, which would be slower than the current solution when checking the Etag.  A way to mitigate this would be to use an `IDistributedCache` that can be shared between instances, which would avoid the scaling issues as it shares the cache across all instances.
+
+In order to use `IDistributedCache`, a backing cache needs to be selected, which could be Redis or a Postgres cache, though the specifc implementation of this is outside the scope of this RFC.
+
+If the distributed cache wasn't hit, there would be fallback to the database.  The "happy path" for this process would look like this:
```
donaldgray, line 59: read-through cache

docs/rfcs/0003-etag-changes.md
```diff
@@ -24,54 +24,153 @@ While the current Etag implementation allows for easily adding and removing Etag
...
+    Api ->> User: Return updated resource
+```
  
 #### Leveraging S3
 
-- S3 has the ability to generate a file hash we could take advantage of
+S3 automatically generates a file hash we could take advantage of for using as an Etag value
+
+  - This file hash is pre-generated and would be easy to retreive
 - This has issues with storage collections, as these aren't stored in S3 - we could use a different system for generating this hash, but then that means there is a different code journey in this specific scenario, which has caused issues with other parts of the presentation API
```
donaldgray, line 149: > a different code journey in this specific scenario

I don't think this will have any bearing on the outcome but this doesn't seem a different code journey, instead just a different implementation:

```cs
public class Something(IEtagGenerator etagGenerator)
{
  public void Processor(Collection collection) 
  {
    var etag = etagGenerator.GenerateEtag(collection);
  }
}

public class StorageCollectionGen : IEtagGenerator { }

public class NotStorageCollectionGen : IEtagGenerator { }
```

docs/rfcs/0003-etag-changes.md
```diff
@@ -24,54 +24,153 @@ While the current Etag implementation allows for easily adding and removing Etag
...
+
+## Recomendations
 
-### Other considerations
+Based on the information discussed above, these are what the new Etags process should do:
+
+- The Etag should be stored in the database under a separate `etag` table
+- While the distributed cache would be useful, as an initial implementation it would be ok to run using the direct database calls.  However, this should be monitored and if it seems that the database begins to become resource starved, then the distributed cache should be imlemented.  However, in the intervening time between deciding the cache is needed and deploying the implementation, additional resources can be dedicated to the [RDS instance](https://aws.amazon.com/rds/instance-types/) to mitigate issues with resource starvation.
+- The hash of the response body should be used to generate the Etag as the S3 object has too many instances where it's not updated
```
donaldgray, line 165: If background-handler can generate this, is it the response body? Or something else that would be used to generate it? (A hash of the `IPresentation` object? Are there any issues with size of response to generate this, like an upper limit (thinking Fargate tasks are running in memory-constrained env, is that likely to be an issue?).

docs/rfcs/0003-etag-changes.md
```diff
@@ -24,54 +24,153 @@ While the current Etag implementation allows for easily adding and removing Etag
...
+
+## Other considerations
 
 based on the [API scratchpad tests](https://github.com/dlcs/iiif-presentation-tests/blob/04213f185bf4fb370855e7e37be27ee4587234bf/tests/apiscratch/t0071-update-managed-asset-manifest.spec.ts#L60-L62) shows that editable resources (i.e.: PUT/POST) should not respond with Etags and instead would require a GET to retrieve the Etag.  This would have implications on the automated tests.
 
-The presentation API [sends Cache-Control: no-cache headers](https://github.com/dlcs/iiif-presentation/issues/140) to stop clients from caching values and then serving old Etags.  As it stands, this shouldn't have an impact on how the Etags work
\ No newline at end of file
+The presentation API [sends Cache-Control: no-cache headers](https://github.com/dlcs/iiif-presentation/issues/140) to stop clients from caching values and then serving old Etags.  As it stands, this shouldn't have an impact on how the Etags work
```
donaldgray: Does `no-cache` have any affect on how CDNs or reverse proxies (ie CloudFront) work? Or does including etag mean it'll make requests to see if content expired?

docs/rfcs/0003-etag-changes.md
```diff
@@ -24,54 +24,153 @@ While the current Etag implementation allows for easily adding and removing Etag
...
+
+Storage collections would need to change the id whenever a child collection is added - a child collection would cause parent etag to update
```
donaldgray, line 173: Is this a discarded consideration? Or does it need solved? 

Would we update the parent storage-collection while updating the child, or queue for later?

docs/rfcs/0003-etag-changes.md
```diff
@@ -24,54 +24,153 @@ While the current Etag implementation allows for easily adding and removing Etag
...
+
+Geenrating the Etag could also be combined with data to indicate versioning (i.e. create has v1 encoded, first update v2 etc.) but care must be taken to make sure that these Etags look truly random as it's possible users begin ascribing meaning to the Etag values if they are too similar or seemingly follow a convention.
+
+## Recomendations
 
-### Other considerations
+Based on the information discussed above, these are what the new Etags process should do:
+
+- The Etag should be stored in the database under a separate `etag` table
+- While the distributed cache would be useful, as an initial implementation it would be ok to run using the direct database calls.  However, this should be monitored and if it seems that the database begins to become resource starved, then the distributed cache should be imlemented.  However, in the intervening time between deciding the cache is needed and deploying the implementation, additional resources can be dedicated to the [RDS instance](https://aws.amazon.com/rds/instance-types/) to mitigate issues with resource starvation.
```
donaldgray, line 164: One option could be to split logic for read+write requests, assuming GET requests _vastly_ outnumber write requests we can optimise for that.

Read requests:
* On receipt of GET w/ etag, get value from cache (read-through cache so may originate from DB)
* If we match, return 304 (memory-lookup only, inexpensive)
* else, read content from DB + s3 and return w/ 200. On way out refresh cache with DB etag if it differs to ensure fresh.
  * If the given etag was for a later version than we had in-memory all we've done is fetch once from DB so minimal extra work.

Write requests:
* On receipt of PUT/POST w/ etag, get value from _db_ - ensures we're always look at the latest one.
* If eTag matches or not allow or not. Main thing is we got this fresh.

docs/rfcs/0003-etag-changes.md
JackLewis-digirati: while the child is updated, the parent collection Etag should also be updated - will make it more clear that this needs to happen

docs/rfcs/0003-etag-changes.md
JackLewis-digirati: I've left this in, but modified the recommendation to use an existing table

docs/rfcs/0003-etag-changes.md
```diff
@@ -0,0 +1,226 @@
...
+
+There's an issue with using the file to generate the hash in that PUT/POST can occur on the hierarchical path and the flat path which has different styles of request body.  This means that potentially updating a resource could have 2 different file hashes if using the request.  This can also happen with variations of `publicId`, `slug` and `parent`. Given these are ultimately the same request, it would be better for a single value to control the Etag on all of these requests and responses. This could be done using the response instead of the request, but there are issues with this around properties such as `ingesting` which will change based on assets. However, this would likely make for a better value when PATCH is integrated.
+
+Additionally, the PUT/POST response is modified at a later date byt the background handler if there are assets attached to the manifest.  This would mean that Etags would not be available for manifests with assets until after processing was completed.  The full process would look like this:
```
griffri, line 150: Typo ("byt")

docs/rfcs/0003-etag-changes.md
```diff
@@ -0,0 +1,226 @@
...
+- This is still a call to another external service
+- Not required if the usage of the presentation API is not large
+
+##### Benefits
+
+- Implementation is shared between nodes, meaning it's easy to scale
+- A failed match would not require calls to the database on items in the cache 
+- Ultimately means less calls on the database, which could be a performance bottleneck
+
+#### Publisher and subscriber with internal cache
```
donaldgray, line 101: fwiw I think, if done correctly and is justified, this is a really powerful pattern. I'm not sure it's warranted here though.

docs/rfcs/0003-etag-changes.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Recommendations
+
+Based on the information discussed above, these are what the new Etags process should do:
+
+- The Etag should be stored in the database under a column on the `manifests` and `collections` table
+- While the distributed cache would be useful, as an initial implementation it would be ok to run using the direct database calls.  However, this should be monitored and if it seems that the database begins to become resource starved, then the distributed cache should be implemented.  Although, in the intervening time between deciding the cache is needed and deploying the implementation, additional resources can be dedicated to the [RDS instance](https://aws.amazon.com/rds/instance-types/) to mitigate issues with resource starvation. However, in preparation for this, an in-memory cache should be implemented using `IDistributedCache` with a short caching duration (~ 5 seconds)
```
donaldgray, line 206: Can this recommendation be clarified? Starts by saying direct DB calls but finished by saving use in-memory cache.

docs/rfcs/0003-etag-changes.md
```diff
@@ -0,0 +1,226 @@
...
+- In order to help with potential resource starvation, the read and write logic should be split like mentioned above.
+- S3 should be used to retrieve the hash as this avoids issues with keeping the response body in memory
+- The cache key should be stored by path
+
+## Other considerations
+
+This section contains additional areas of concern where additional tickets should be generated to cover the below scenarios
+
+based on the [API scratchpad tests](https://github.com/dlcs/iiif-presentation-tests/blob/04213f185bf4fb370855e7e37be27ee4587234bf/tests/apiscratch/t0071-update-managed-asset-manifest.spec.ts#L60-L62) shows that editable resources (i.e.: PUT/POST) should not respond with Etags and instead would require a GET to retrieve the Etag.  This would have implications on the automated tests.
```
donaldgray, line 215: fyi, ticket for this - https://github.com/dlcs/iiif-presentation/issues/354

docs/rfcs/0003-etag-changes.md
```diff
@@ -0,0 +1,226 @@
...
+
+The presentation API [sends Cache-Control: no-cache](https://github.com/dlcs/iiif-presentation/issues/140) headers to stop clients from caching values and then serving old Etags.  As it stands, this shouldn't have an impact on how the Etags work, due to how Cloudfront caches with the `no-cache` header.  Essentially, if the `Minimum TTL` is larger than 0, the `no-cache` header will not be respected, and instead cached for the `Minimum TTL` value. Details [here](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html).
+
+**NOTE:** the dev environment currently has a `Minimum TTL` of 0, meaning nothing would be cached, it might be worth increasing this value.
```
donaldgray, line 219: fwiw, CF on dev isn't used, it's for testing rewrites. Doesn't seem worthwhile having dev traffic cached.

donaldgray (approved):
Couple of small points but overall think we should be good to create tickets from this.

I think some of the caching points here will need addressed more generally for the wider system.

Can you also update the index RFC readme.md

PR #372, opened 2025-05-01
Initial attempt adding new path rewrite rules for presentation URL's

JackLewis-digirati:


src/IIIFPresentation/Repository/Paths/IPresentationPathGenerator.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Repository.Paths;
+
+public interface IPresentationPathGenerator
+{
+    /// <summary>
+    /// Generate full path for IIIF Presentation resources
+    /// </summary>
+    public string GetPresentationPathForRequest(string presentationServiceType, int? customerId = null,
```
donaldgray, line 8: Should there be 2 overloads here? This takes 2 params: `string? hierarchyPath = null, string? resourceId = null` but every call will only provide `hierarchyPath` OR `resourceId`.

src/IIIFPresentation/Core/Web/TypedPathTemplateOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+public class TypedPathTemplateOptions
+{
+    public const string SettingsName = "PathRules";
+
+    internal static readonly Dictionary<string, string> DefaultFormats = new ()
+    {
+        ["ManifestPrivate"] = "{customerId}/manifests/{resourceId}",
+        ["CollectionPrivate"] = "{customerId}/collections/{resourceId}",
+        ["ResourcePublic"] = "{customerId}/{hierarchyPath}",
+        ["Canvas"] = "{customerId}/canvases/{resourceId}"
```
donaldgray, line 16: I think these should all start with a `/`? In `HttpRequestX` there's logic to check if the path starts with `/`, and it adds it if it doesn't. 

```cs
if (!string.IsNullOrEmpty(path) && !path.StartsWith(PathDelimiter))
{
  path = PathDelimiter + path;
}
```

If the defaults don't have `/` as a prefix then _every_ default request will do some string manipulation to add it when it seems easier to just add that to the template, particularly as this code is in hotpath and will be called a lot? Alternatively use `/` for all path values and avoid the need to ensure it's there in `GetDisplayUrl()`?

src/IIIFPresentation/Core/Web/TypedPathTemplateOptions.cs
```diff
@@ -0,0 +1,54 @@
+namespace Core.Web;
+
+/// <summary>
+/// A collection of options related to path generation. This differs from <see cref="PathTemplateOptions"/> as this
```
donaldgray, line 4: `This differs from <see cref="PathTemplateOptions"/> as this` comment isn't relevant here, looks like copy/paste from Protagonist?

src/IIIFPresentation/Core/Web/HttpRequestX.cs
```diff
@@ -24,6 +26,10 @@ public static string GetDisplayUrl(this HttpRequest request, string? path = null
         var queryString = includeQueryParams
             ? request.QueryString.Value ?? string.Empty
             : string.Empty;
+        if (!string.IsNullOrEmpty(path) && !path.StartsWith(PathDelimiter))
```
donaldgray, line 29: See comment in `TypedPathTemplateOptions` - if we want paths to start with `/` then the templates should.

src/IIIFPresentation/API/Paths/ConfigDrivenPresentationPathGenerator.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace API.Paths;
+
+public class ConfigDrivenPresentationPathGenerator(
+    IOptions<TypedPathTemplateOptions> settings,
+    IHttpContextAccessor httpContextAccessor)
+    : IPresentationPathGenerator
+{
+    private readonly TypedPathTemplateOptions settings = settings.Value;
+
+    public string GetPresentationPathForRequest(string presentationServiceType, int? customerId, string? hierarchyPath, string? resourceId)
```
donaldgray, line 14: Can `customerId` be null?

donaldgray (commented):
Overall looks good, once BackgroundHandler resolved can test `Canvas` is updated correctly.

Some of the examples use `ManifestPublic` and `CollectionPublic` which can be removed.

src/IIIFPresentation/API.Tests/Paths/ConfigDrivenPresentationPathGeneratorTests.cs
```diff
@@ -0,0 +1,145 @@
...
+                ["CollectionPrivate"] = "https://base/{customerId}/collections/{resourceId}",
+                ["ResourcePublic"] = "https://base/{customerId}/{hierarchyPath}",
+                ["Canvas"] = "https://base/{customerId}/canvases/{resourceId}",
+            }
+        }
+    };
+    
+    public ConfigDrivenPresentationPathGeneratorTests()
+    {
+        HttpContextAccessor.HttpContext = A.Fake<HttpContext>();
```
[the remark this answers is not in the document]
JackLewis-digirati: think updating this might take a bit more than using fakes as it's overridden in tests using `A.CallTo`

src/IIIFPresentation/Core/Core.csproj
```diff
@@ -8,6 +8,7 @@
 
     <ItemGroup>
       <PackageReference Include="iiif-net" Version="0.3.5" />
+      <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="2.2.0" />
```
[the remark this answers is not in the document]
JackLewis-digirati: think this was from an earlier incarnation from shared stuff between background handler and API - have moved the things using this to API and removed `Abstractions`

src/IIIFPresentation/API/Paths/ConfigDrivenPresentationPathGenerator.cs
JackLewis-digirati: if you don't have a `customerId` set in the template, then it can be null

donaldgray (approved):

PR #385, opened 2025-05-20
Modify the batch notification to use the standard format

JackLewis-digirati:


src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessage.cs
```diff
@@ -1,11 +1,65 @@
...
+    public bool Superseded { get; set; }
+    
+    [JsonRequired]
+    public DateTime Submitted { get; set; }
+    
+    [JsonRequired]
+    public DateTime Finished { get; set; }
+}
+
+public class OldBatchCompletionMessage
```
donaldgray, line 32: Comment outlining what makes it "old" and some details around when it can be deleted?

donaldgray (approved):

Issue #378, opened 2025-05-14
Multiple non-choice paintingAnnotations


donaldgray:
Allow multiple non-choice paintingAnnotations on the same canvas. Example manifest in cookbook - https://iiif.io/api/cookbook/recipe/0036-composition-from-multiple-images/

This needs to handle incoming via `"items"` or `"paintedResources"`. With the final `canvasPainting` DB record accurately reflecting the payload (ie matching `canvasId` but no `choiceOrder`, `target` populated etc) and the final manifest reflecting the payload too.

Resolving this will likely require changes in a few areas:
* Validation (both Manifest level and during processing). There may be validation for ensuring items with matching canvasId have differing choice.
* `choiceOrder` is configured to be non-nullable in db for inclusion in unique index as, until now, multiple items must be part of a choice.
* Generation of placeholder canvases for manifest stored in staging
* Final generated manifest, when created via `paintedResources`, must handle non-choice shared bodies + include target in `"target"`
  * When creating via `paintedResources` - allow `"target"` to be specified in payload.
* Generation of canvasId must include target, if used for `"target"` and canvasPainting has target value.

donaldgray:
See https://github.com/dlcs/docs/blob/wip-skeleton/public/manifest-builder/database.py#L90-L99

Issue #351, opened 2025-04-04
Conditionally create or update assets


donaldgray:
Note, much of this is detailed in [Protagonist RFC 019](https://github.com/dlcs/protagonist/blob/main/docs/rfcs/019-presentation-dlcs.md#example-of-calls-made)

On receipt of an upsert request for a manifest - check if the asset exists by looking in canvasPaintings table and `POST /allImages` as fallback. 

This lookup will need to use the full assetId (`customer/space/asset`)
* `id` from `paintedResource>asset>id`
* `space` from `paintedResource>asset>space` OR `manifest>space`
* `customer` from customer

After doing both of these we should be able to determine what assets exist already (don't need to be ingested) and which ones don't exist (do need to be ingested).

For those assets that exist already:
* `PATCH /allImages` to add manifest id to those assets in IIIF-CS
* Record asset in canvasPainting table (as-is).

For those assets that existed previously, but don't with the current iteration:
* `PATCH /allImages` to remove manifest id to those assets in IIIF-CS.
* Remove asset from canvasPainting table (as-is).

For those assets that _don't_ exist:
* Create batch by calling IIIF-CS (as-is), specifying "manifest" field.
* Record asset in canvasPainting table (as-is).
* Record batch in db (as-is).

> [!NOTE]
> This could result in some Manifests not having a return value initially - in that instance we will return an error and any manifests in this state will be 'lost'
>
> ```cs
> if (allUpdates) throw new NotImplementedException("#352 will do this");
> ```

donaldgray:
Added `hold` label - see `note` on ticket, merging this will result in errors until #352 is implemented

PR #386, opened 2025-05-20
Handle composite canvas

### LINKED ISSUE #279: Consistent approach to `CanvasPainting`

donaldgray:
Overall goal of PR is to allow composite canvases to be handled (ie canvases that contain multiple painting annotations that aren't a choice). 

> [!NOTE]
> This implements composite handling for `"items"` but manifest completion is not handled for `"paintedResource"` creation - this will come in follow up PR due to complexity. `ManifestConverter` will reflect back provisional items including composite canvases but these are not reflected in final manifest (see above).

This primarily resolves most of #378, also resolves #279. 

Summary of changes:
* Switch various points so that `CanvasId` becomes the grouping element, rather than `CanvasOrder` - composite canvases allow multiple `CanvasPaintings` with differing `CanvasOrder` to be on same canvas.
* Altered `ChoiceOrder` to be nullable. Previously this was nullable in entity but not in DB, this was to allow the value to be used in unique constraint but now `null` choice is valid this isn't required.
* Tweaked and simplified `PresentationManifestValidator` handle
* For #279 - refactored `CanvasPaintingResolver` to introduced common `ICanvasPaintingParser`, which is an interface shared when parsing from Items or CanvasPainting.
* Lots of test changes - adding new and simplifying some asset test setup.

> [!NOTE]
> This PR contains a migration


src/IIIFPresentation/Repository/Manifests/ManifestPaintedResourceParser.cs
```diff
@@ -0,0 +1,96 @@
...
+        var assetId = GetAssetIdForAsset(paintedResource.Asset!, customerId);
+        logger.LogTrace("Processing canvas painting for asset {AssetId}", assetId);
+        var cp = new CanvasPainting
+        {
+            Label = payloadCanvasPainting?.Label,
+            CanvasLabel = payloadCanvasPainting?.CanvasLabel,
+            CanvasOrder = canvasOrder,
+            AssetId = assetId,
+            ChoiceOrder = payloadCanvasPainting?.ChoiceOrder,
+            Ingesting = true, // TODO - this needs to conditionally be set once #351 implemented
```
JackLewis-digirati, line 58: I've added a note to #351 to make sure this is caught

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -242,125 +233,43 @@ private void UpdateCanvasPainting(List<CanvasPainting> existingCanvasPaintings,
...
+        catch (InvalidCanvasIdException cpId)
         {
-            // this means that the canvas id has not been set, but there's a matching canvas painting record
-            if (canvasPainting.CanvasId == null)
-            {
-                return (null, canvasPaintings.First(c => c.CanvasOrder == canvasOrder).Id);
-            }
-
-            return (ErrorHelper.CanvasOrderDifferentCanvasId<PresentationManifest>(canvasPainting.CanvasId), null);
+            logger.LogDebug(cpId, "InvalidCanvasId encountered in {ManifestId}", presentationManifest.Id);
```
JackLewis-digirati, line 271: should this be higher than debug?

JackLewis-digirati (approved):

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
donaldgray: I left it as debug as it's an exception that is caused by bad data and we don't tend to logs things that cause 400 exceptions (e.g. from validators).

PR #411, opened 2025-07-01
Feature/handle reingest property

JackLewis-digirati:


src/IIIFPresentation/API.Tests/Integration/ModifyManifestAssetUpdateTests.cs
```diff
@@ -2330,5 +2332,151 @@ await amazonS3.GetObjectAsync(LocalStackFixture.StorageBucketName,
...
+        
+        // check asset in another manifest causes the manifest PATCH to happen
+        A.CallTo(() => DLCSApiClient.UpdateAssetManifest(Customer,
+            A<List<string>>.That.Matches(a => a.First() == $"{Customer}/{NewlyCreatedSpace}/{assetId}_1"),
+            A<OperationType>.That.Matches(a => a == OperationType.Add),
+            A<List<string>>._, A<CancellationToken>._)).MustHaveHappened();
+    }
+    
+    [Fact]
+    public async Task UpdateManifest_IgnoresManifestsProperty_WhenNoneExist()
```
donaldgray, line 2423: What does `_WhenNoneExist` mean? Shouldn't it be ignored in all cases?

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -148,38 +171,56 @@ await dlcsApiClient.UpdateAssetManifest(dbManifest.CustomerId,
...
+            .ToList() ?? [];
+
+        var (assetsInDatabase, trackedAssetsToReingest) = RetrieveTrackedAssets(customerId, assetIdsWithReingestion);
 
-        if (dbManifest != null)
+        var assetsTrackedElsewhere = assetsInDatabase.Where(a => a.ManifestId != dbManifest?.Id)
+            .Select(a => a.AssetId!).Distinct().ToList();
+
+        // all assets are tracked and not to be reingested
+        if (assetsInDatabase.Count == assetIdsWithReingestion.Count && !assetsTrackedElsewhere.Any() && !trackedAssetsToReingest.Any())
```
donaldgray, line 188: For these two `.Any()` calls I'm seeing suggestion

> Prefer comparing 'Count' to 0 rather than using 'Any()', both for clarity and for performance

using extension method `.IsNullOrEmpty()` uses count while maintaining readability.

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -188,28 +229,76 @@ await dlcsApiClient.UpdateAssetManifest(dbManifest.CustomerId,
...
 
+    private (List<CanvasPainting> AssetsInDatabase, List<CanvasPainting> TrackedAssetsToReingest) RetrieveTrackedAssets(
+        int customerId, List<AssetIdsWithReingestion> assetIdsWithReingestion)
+    {
+        List<CanvasPainting> assetsInDatabase = [];
+        List<CanvasPainting> trackedAssetsToReingest = [];
 
-    private static List<JObject> GetUntrackedAssets(List<JObject> payloadAssets, List<CanvasPainting> assetsInDatabase, 
-        List<AssetId> dlcsAssetIds)
+        foreach (var canvasPainting in dbContext.CanvasPaintings.Where(cp => cp.CustomerId == customerId))
```
donaldgray, line 247: This is really expensive. Looking at UoL, we have 150K canvasPaintings, could easily end up over 1M so we don't want to iterate over every one every time.

Be more targeted? 

First start `dbContext.CanvasPaintings.Where(cp => cp.CustomerId == customerId && cp.Manifest = manifestId && cp.AssetId != null)` (do we have these already)? I suspect that would catch the 99% of cases. 

If we don't find them then make equivalent of `select * from canvas_painting where customer = @id and assetId in [@id_list]`. Will leverage database indexes etc. Depending on the number of assets we may need to batch this

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -83,30 +83,53 @@ public async Task<DlcsInteractionResult> HandleDlcsInteractions(
             
         }
 
+        return await UpdateDlcsWithAssets(request, manifestId, dbManifest, assets, spaceId, cancellationToken);
+    }
+
+    private async Task<DlcsInteractionResult> UpdateDlcsWithAssets(WriteManifestRequest request, string manifestId,
```
donaldgray, line 89: General observation about this class - it seems really complex. There's a lot happening - consider refactoring to simplify? As discussed on slack, would it be better to iterate over each CanvasPainting and attempt to categorise individually, rather than bulk categorisations? Starting with the most likely/least expensive to identify use case, shortcutting if we can?

Failing that adding code comments that just give a narrative description would help the reader

donaldgray (changes requested):
Overall it looks good and it functionally does what we need it to do.

The following couple of points need addressed:
* Set `"reingest"` prop at correct level
* Don't iterate through every canvasPainting for a customer when checking.
* `DlcsManifestCoordinator` - either refactor to simplify/clarify or add explanatory comments? (if possible, if this'll be time consuming comments will help)

src/IIIFPresentation/API.Tests/Integration/ModifyManifestAssetUpdateTests.cs
JackLewis-digirati: made sense in my head when I did this - it means `when no assets exist`

src/IIIFPresentation/API/Features/Manifest/ManagedAssetResultFinder.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        return dlcsInteractionRequests;
+    }
+
+    private List<CanvasPainting> FindAssetsInAnotherManifest(int customerId, 
+        List<(AssetId assetId, PaintedResource paintedResource)> assetsNotFoundInSameManifest)
+    {
+        List<CanvasPainting> inAnotherManifest = [];
+        
+        foreach (var chunkedAssetsToCheck in assetsNotFoundInSameManifest.Chunk(500))
```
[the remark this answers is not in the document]
JackLewis-digirati: I'v e created a ticket here to help track this better - https://github.com/dlcs/iiif-presentation/issues/434

p-kaczynski (approved):
Thanks for creating the tickets for follow up discussions. LGTM

PR #428, opened 2025-07-10
Allow duplicate assets on the manifest

### LINKED ISSUE #392: Duplicate assets on Manifest
Duplicate assets on a manifest will result in errors when constructing the final Manifest. 

This is due to the `BatchCompletionMessageHandler` overwriting the Canvas, AnnoPage, PaintingAnno ids on the NQ manifest before handing to `ManifestMerger` for final construction. 

The Ids are set using `var canvasPaintingForAsset = canvasPaintings.FirstOrDefault(cp => cp.AssetId == assetId);` which will result on some canvases (and annos etc) having incorrect ids, as the ids will be set using the first matching.

The fix is to _not_ set the ids upfront on the NQ canvases.

This may not be common but is a valid use case. E.g. a customer may have Asset `/99/100/redacted` asset that is used as a placeholder for any redacted images. If a 200 page book has 10 redacted pages that asset could be used 10 times.

> [!IMPORTANT]
> This should be done after #378 is completed as that's making changes to `ManifestMerger`

In order to handle duplicate Assets we need to accept these in API. We need to verify whether we de-duplicate any `PaintedResource` payloads that contain duplicate assets as these will be rejected by the Protagonist API. If we receive duplicate `Asset` then we need to de-dup before passing to DLCS API _but_ if we receive the same asset-id with anything else that differs (metadata, delivery-channels) we need to reject.

JackLewis-digirati:
Resolves #392 

This PR makes it so that the DLCS coordinator is modified so that assets with the same asset id are only ingested into the DLCS once

### DIFF
--- a/src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
@@ -207,12 +207,13 @@ public class DlcsManifestCoordinator(
 
             if (!assets.TryAdd(dlcsInteractionRequest.AssetId, dlcsInteractionRequest.Asset))
             {
+                logger.LogDebug("Asset {AssetId} has been specified multiple times, validating they match", dlcsInteractionRequest.AssetId);
                 var assetInDictionary = assets[dlcsInteractionRequest.AssetId];
                 
                 if (!JToken.DeepEquals(assetInDictionary, dlcsInteractionRequest.Asset))
                 {
                     return EntityResult.Failure(
-                        $"Asset {dlcsInteractionRequest.AssetId} is referred to multiple times, and does not match for ingestion",
+                        $"Asset {dlcsInteractionRequest.AssetId} is specified multiple times but has conflicting data",
                         ModifyCollectionType.AssetsDoNotMatch, WriteResult.BadRequest);
                 }
             }
--- a/src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
@@ -7,10 +7,12 @@ using Core;
 using DLCS.API;
 using DLCS.Exceptions;
 using DLCS.Models;
+using JsonDiffPatchDotNet;
 using Models.API.General;
 using Models.API.Manifest;
 using Models.Database.Collections;
 using Models.DLCS;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using Repository;
 using CanvasPainting = Models.Database.CanvasPainting;
@@ -212,8 +214,11 @@ public class DlcsManifestCoordinator(
                 
                 if (!JToken.DeepEquals(assetInDictionary, dlcsInteractionRequest.Asset))
                 {
+                    var jsonDiffPatch = new JsonDiffPatch();
+                    var diff = jsonDiffPatch.Diff(assetInDictionary, dlcsInteractionRequest.Asset);
+                    
                     return EntityResult.Failure(
-                        $"Asset {dlcsInteractionRequest.AssetId} is specified multiple times but has conflicting data",
+                        $"Asset {dlcsInteractionRequest.AssetId} is specified multiple times, but has conflicting data - diff: {JsonConvert.SerializeObject(diff)}",
                         ModifyCollectionType.AssetsDoNotMatch, WriteResult.BadRequest);
                 }
             }
--- a/src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
@@ -165,71 +165,6 @@ public class DlcsManifestCoordinator(
         await dlcsApiClient.UpdateAssetManifest(dbManifest.CustomerId,
         canvasPaintingsToRemove.Select(cp => cp.AssetId?.ToString()).ToList(), OperationType.Remove,
     [dbManifest.Id], cancellationToken);
-    
-    private async Task<(List<JObject> untrackedAssets, List<AssetId> dlcsAssetIds)> FindAssetsThatRequireAdditionalWork(List<JObject> assets, 
-        Models.Database.Collections.Manifest? dbManifest, int customerId, CancellationToken cancellationToken = default)
-    {
-        var assetIds = assets.Select(a => a.GetAssetId(customerId));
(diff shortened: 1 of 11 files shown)


src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
```diff
@@ -87,11 +87,12 @@ private Dictionary<AssetId, Canvas> BuildAssetIdToCanvasLookup(Manifest namedQue
         {
             return namedQueryManifest
                 .Items!
-                .ToDictionary(canvas => canvas.GetAssetIdFromNamedQueryCanvasId(), canvas => canvas);
+                .GroupBy(canvas => canvas.GetAssetIdFromNamedQueryCanvasId())
+                .ToDictionary(canvas => canvas.Key, canvas => canvas.First());
```
donaldgray, line 91: Not sure this is required? This would only be required if the NQ is returning > 1 separate canvases for the same Asset, which shouldn't be possible. The "duplicate" Asset is only a Presentation concern, Protagonist doesn't have any notion of that - it just has a single Asset.

src/IIIFPresentation/Services.Tests/Manifests/Helpers/ManifestMergerTests.cs
```diff
@@ -734,4 +734,32 @@ public void ProcessCanvasPaintings_PreservesNonStandardContext()
...
+    [Fact]
+    public void ProcessCanvasPaintings_GeneratesExpectedManifest_WhenMultipleAssetsWithTheSameId()
+    {
+        // Arrange
+        var blankManifest = new Manifest();
+        var assetId = TestIdentifiers.AssetId();
+
+        var namedQueryManifest = ManifestTestCreator.New()
+            .WithCanvas(assetId, c => c.WithImage())
+            .WithCanvas(assetId, c => c.WithImage())
```
donaldgray, line 747: Related to previous comment - this wouldn't reflect how a NQ would be returned?

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -196,19 +198,30 @@ private static List<JObject> GetAssetJObjectList(WriteManifestRequest request) =
...
                     break;
                 }
                 case IngestType.ManifestId:
                     dlcsInteractionRequest.Asset[AssetProperties.Manifests] = new JArray(new List<string> { manifestId });
                     break;
             }
+
+            if (!assets.TryAdd(dlcsInteractionRequest.AssetId, dlcsInteractionRequest.Asset))
+            {
+                var assetInDictionary = assets[dlcsInteractionRequest.AssetId];
```
donaldgray, line 210: Debug log or something? _"Asset xxx has been specified multiple times, validating they match"_. 

Would help explain what's happening in this codeblock and also help with debugging.

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -196,19 +198,30 @@ private static List<JObject> GetAssetJObjectList(WriteManifestRequest request) =
...
+                
+                if (!JToken.DeepEquals(assetInDictionary, dlcsInteractionRequest.Asset))
+                {
+                    return EntityResult.Failure(
+                        $"Asset {dlcsInteractionRequest.AssetId} is referred to multiple times, and does not match for ingestion",
```
donaldgray, line 215: `", and does not match for ingestion"` doesn't seem the clearest explanation... `$"Asset {dlcsInteractionRequest.AssetId} is specified multiple times but don't match"` or `"has conflicts"`? 

Should we output the 2 payloads here to aid troubleshooting?

donaldgray (changes requested):
When running locally fetching asset details is failing, happening on POST/PUT and GET

```json
"asset": {
   "@id": "https://dlcs.host/customers/7/spaces/144/images/dup_test2",
   "error": "Unable to retrieve asset details"
}
```

The `/allImages` endpoint rejects duplicates with a 400 response.

Thought, may or may not be viable - is this duplicate check the type of thing we could do in the validator? Before it gets further into the code? Or is it okay to do further dowm?

JackLewis-digirati:
I could swap it to do the duplicate check in the validator, but it does mean interrogating the `JObject` within the validator as well as in code, which I think would mean duplicating the `assetId` retrieval, so I felt it would be more work than if we do it at the point we already know the `assetId`

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
JackLewis-digirati: I don't think the index will work as we've modified it due to managed assets - looking at the jsondiffpatch will probably give the most relevant output (as opposed to here's the entire object) so will look to add this

src/IIIFPresentation/DLCS/API/DlcsApiClient.cs
```diff
@@ -132,7 +132,8 @@ public async Task<IList<JObject>> GetCustomerImages(int customerId, ICollection<
         var results = new List<JObject>();
         foreach (var idBatch in assetIds.Chunk(settings.MaxImageListSize))
         {
-            var hydraImages = new HydraCollection<JObject>(idBatch.Select(id => new JObject {["id"] = id}).ToArray());
+            // duplicate images cause errors in the DLCS, so strip them out
+            var hydraImages = new HydraCollection<JObject>(idBatch.Distinct().Select(id => new JObject {["id"] = id}).ToArray());
```
donaldgray, line 136: This is de-duping per chunk - should the entire list be deduped before chunking? Slim chance but there could be duplicate Assets returned from this method if there are duplicates in `assetIds` but they're in different chunks.

E.g. `99/1/bookend` is used for front page and cover page of a book and there are 600 pages. This would mean 2 chunks sent to Protagonist, both contain `99/1/bookend` so the final `GetCustomerImages()` result set would contain `99/1/bookend` twice.

OR `99/1/redacted` is used as placeholder for redacted page. 100 pages of a 550 page book are redacted, so 450 unique assets, but this ends up in 2 chunks, when it could be one (so 2 HTTP requests rather than 1) and result set would contain duplicates.

donaldgray (commented):
Single comment on how de-dup works. Also - should description be changed as there are now no changes to ManifestMerger?

src/IIIFPresentation/DLCS/API/DlcsApiClient.cs
JackLewis-digirati: hrm, yeah, makes sense - will change it to de-dup the full list

PR #436, opened 2025-07-21
Allow the canvas id to be successfully picked up from rewritten paths

### LINKED ISSUE #435: Rewritten `canvasId` values are are not accepted
#366 allows rewritten paths to be provided as values for Presentation resource ID properties (e.g `"parent"` and `"publicId"`)
It should have allowed Presentation to also parse rewritten `"canvasID"` properties, but this is currently not the case:
```
"canvasPainting": {
 	....
     "canvasId": "https://cf.dlcs.digirati.io/presentation/151/canvases/my-image-a",
     ....
 }
```
This currently results in a 400 response that states `The canvas ID /presentation/151/canvases/my-image-a is invalid`.

JackLewis-digirati:
Resolves #435

Fixes the API to accept a rewritten `canvasId` to correctly set the value


p-kaczynski (approved):
Code wise no complaints from me, functionally will be checked by automated tests so lgtm

PR #438, opened 2025-07-22
mixed manifest validation

JackLewis-digirati:


p-kaczynski (approved):
It seems very much in line with some changes I have made on my branch, which I think is a good sign for both of us. It looks good, though I will have much better feel for the code after I rebase my stuff onto it.

Issue #290, opened 2025-02-21
Prevent "items" and "paintedResources" updates


donaldgray:
As an MVP Manifest can only be created and modified by `"items"` OR `"paintedResources"`, as soon as one way is chosen subsequent updates need to use this same method.

If a manifest has been created via one method and we receive an update using a different method return a 409 with a clear error message. This can be removed in the future as we implement full update logic.

The method used can be determined by the existence of Batches related to the Manifest (if Manifest has Batches it was created via `"paintedResources"`).

Issue #424, opened 2025-07-09
Identify assets in Canvases


donaldgray:
Canvases in `"items"` may refer to Assets that are already managed by IIIF-CS. This ticket is to implement the identification of those assets.

> [!NOTE]
> This logic only applied when there is no matching `"paintedResource"` in the payload. If there is a matching `"paintedResource"` then the `Canvas` cannot contain any annotationPages.

Main points:
* `CanvasPainting` records that are created from `"items"` will stay as-is. If they reference a managed asset then the `AssetId` column will be populated.
* Any managed assets should be updated in the Protagonist database, this currently happens for assets that originate from `paintedResource` so this might automatically happen.

Identification of assets differ based on type

> [!NOTE]
> This ticket will only handle identification of assets from canonical paths only, identifying rewritten asset paths will come later

Acceptance Criteria:
* Images are identified from body services OR body id if there are no services
* If body services and body id point to different Assets, BadRequest
* AV are identified from body id only.
* If we detect the format of a managed Asset but the asset is then found not to exist in the DLCS, reject. When we reject, we need to ensure we do so before making any write requests to DLCS (aside from Space creation possibly, although ideall we wouldn't do that yet).
* Only managed Assets for the current customer are considered.
* Only accept Assets from known hostnames

--------
_Below taken from RFC_

## Images

The canonical path is `/iiif-img/{version?}/{customer}/{space}/{asset}/{image-request}`

Attempt to identify the Asset from any referenced ImageService on a `body` with `"type": "Image"`. If this is a Protagonist imageService then this asset must be managed by that instance. This will be `id`, for `ImageService3`, or `@id`, for `ImageService2`. 
* `$.items[*].items[*].items[*].body.service[*].id` OR `$.items[*].items[*].items[*].body.service[*]['@id']`
* Examples:
  * `https://dlcs.host/iiif-img/v3/99/10/tk_421` - explicit V3
  * `https://dlcs.host/iiif-img/v2/99/10/tk_421` - explicit V2
  * `https://dlcs.host/iiif-img/99/10/tk_421` - canonical version

We can also attempt to identify an Asset from `body.id` alone, which is a static image, not a IIIF ImageService.
* `$.items[*].items[*].items[*].body.id`
* Example:
  * `https://dlcs.host/iiif-img/99/10/tk_421/full/1350,900/0/default.jpg`

If the `body.id` is identified as one managed asset but its service is a different managed asset, then that's a 400|BadRequest. 

## AV

[...]

PR #442, opened 2025-08-05
Mixed manifests provisional canvases and merging of manifests implementation

### LINKED ISSUE #432: Provisional Canvases
How does the handling of provisional canvases (#227) need tweaked?

Current logic does not save the provisional canvases to the "staging" version of the Manifest; it instead writes them dynamically on the way out via https://github.com/dlcs/iiif-presentation/blob/0736923b785c4df9d57dc6639f9ea5669f8d3b90/src/IIIFPresentation/API/Converters/ManifestConverter.cs#L61-L62

This might no longer be the case. If we have received `"items"` and `"paintedResources"` we'll need to merge

* Rather than set `.Items` on the way out each time, populate and save in "staging" S3 object.
* Note this will affect what we need to do in the `ManifestMerger`
* Implementation nice to have - could the code that builds the provisional canvases and the "real" canvases share.

### LINKED ISSUE #433: Update `ManifestMerger`
Changes are required to `ManifestMerger`. The code makes 2 assumptions that are no longer necessarily true:
* That every `CanvasPainting` associated with the Manifest is for an Asset that will be rendered on the final Canvas.
* That the `baseManifest` has no `"items"`. `baseManifest` is either the Manifest :
  * from the "staging" location, if this is async process and picked up by background handler OR
  * from the incoming payload.

We need to handle the case where there are a mix of `CanvasPaintings`; those that are to be added to the Manifest because they have an `AssetId` and those that have an `AssetId` but shouldn't be added to the Manifest (sees #424, this will be due to the Canvas body pointing at an already managed Asset).

We also need to handle there being `"items"` on the Manifest already. In 2 cases:
* Those that are to be ignored, as there's no content resources to be added (these were Canvas only in payload, no `"paintedResource"`)
* Those that are to be populated with content resource from DLCS. The may have annotation pages (added via #432) present. If there are annotationPages present these can be overwritten as they will just be provisional.
  * These Canvases may have other content (e.g. `"metadata"`, `"rendering"`, `"navPlace"`) that must be maintained in the final Manifest.
* See [examples](/TODO-path to RFC when merged) in RFC for sample payloads. Note that these are not exhaustive!

> [!NOTE]
> Calls to NamedQueries can return more items than we will add to the Manifest.

### LINKED ISSUE #448: Received a public manifest with `"processing"` behaviour
Encountered when testing #442 with @JackLewis-digirati.

Somehow received a public manifest that had provisional canvases. Not sure how it's possible, looking in DB there was still a canvasOriginalId on the record.

Possible steps:
* Create a matched canvas (there was a lot of back + forth saving with this)
* Remove the "items" part of matched canvas and resaved.
Other notes - there was a lot of other canvases added. This canvas had been reordered.

### LINKED ISSUE #450: manifestMerger matches a canvas with a canvasPainting record using an FQDN
as part of the `manifestMerger` process, the manifests match a `canvas` with a `canvasPainting` record using a fully qualified canvas id like this:

```csharp
var canvas = new Canvas { Id = pathGenerator.GenerateCanvasId(firstCanvasPaintingInCanvas) };

var canvasMatchedFromCanvasId = canvasesFromManifest.FirstOrDefault(i => i.Id == canvas.Id);
```

this works fine in a usual scenario, but if a customer asks to change the path rewrite rules for the URI used to generate a canvas id, it would mean that anything in the `staging` bucket would have to be processed, otherwise the check `FirstOrDefault(i => i.Id == canvas.Id)`  would not be able to match.

This could potentially happen if a customer asks for a change to path rewrites, and at the same time ingests several thousand manifests

This could be improved so that we can handle this scenario.

One suggestion, is that the `pathRewriteParser` could be used to generate a dictionary (saving time searching the list) of the canvas id to the canvas, so that an FQDN is not required to be matched, and instead the match could be something like `canvasDictionary[firstCanvasPaintingInCanvas.Id]`

JackLewis-digirati:
resolves #432 
resolves #433
resolves #448
resolves #450

This PR modifies the provisional canvas paintings to work with the new mixed manifests feature.

Additionally, the manifest merger has been updated to provide final manifests that work with a mixture of items and painted resources

There has been a modification to settings to allow the settings based path generator to generate provisional paths, as opposed to the config generator.  This has meant a reorganisation of settings and __the api now implements the `presentationApiUrl` settings previously only available in the background handler.__ 

An example of this new settings block now looks like this:

```json
"PathSettings": {
    "PresentationApiUrl": "https://localhost:7230",
    "CustomerPresentationApiUrl": {
      "2": "https://dlcs-dev.example",
      "3": "https://dlcs-test.example"
    },
    "PathRules": {
      "Defaults": {
        "ResourcePublic": "/presentation/{hierarchyPath}",
        "ManifestPrivate": "/{customerId}/manifests/{resourceId}",
        "CollectionPrivate": "/{customerId}/collections/{resourceId}",
        "Canvas": "/{customerId}/canvases/{resourceId}"
      },
      "Overrides": {
        "example.com": {
          "ResourcePublic": "https://foo.com/{customerId}/{hierarchyPath}",
          "ManifestPrivate": "/{customerId}/manifests/{resourceId}",
          "CollectionPrivate": "/{customerId}/collections/{resourceId}",
          "Canvas": "/{customerId}/canvases/{resourceId}"
        }
      }
    }
```

> [!NOTE]
> where these settings supersede settings with the same name in the previous iteration

> [!IMPORTANT]
> this PR will require changes to the terraform

> [!NOTE]
> This PR also reverts changes from #290 which disallows manifests being created with either assets or items in the MVP

### DIFF
--- a/src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
+++ b/src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
@@ -66,7 +66,7 @@ public class ManifestMerger(SettingsBasedPathGenerator pathGenerator, IPathRewri
     private void ValidateManifests(Manifest baseManifest, Manifest? namedQueryManifest)
     {
         // Ensure NQ has items or we can't do anything
-        if (namedQueryManifest?.Items == null && baseManifest.Items == null)
+        if (namedQueryManifest?.Items == null && baseManifest.Items.IsNullOrEmpty())
         {
             logger.LogWarning("NamedQuery Manifest '{ManifestId}' null or missing items",
                 namedQueryManifest?.Id ?? "no-id");
@@ -107,8 +107,10 @@ public class ManifestMerger(SettingsBasedPathGenerator pathGenerator, IPathRewri
 
         var items = new List<Canvas>();
         
+        // dictionary where the key will either be the canvasOriginalId OR the canvasPainting id which is used
+        // to provide a fast match between the canvas and the canvasPainting record
         var manifestCanvasDictionary = baseManifest.Items?.ToDictionary(
-            // customer id should be the same across all canvas paintings
+            // customer id must be the same across all canvas paintings
             key => pathRewriteParser.ParsePathWithRewrites(key.Id, canvasPaintings.First().CustomerId).Resource ??
                    key.Id!,
             val => val) ?? new Dictionary<string, Canvas>();
@@ -135,7 +137,6 @@ public class ManifestMerger(SettingsBasedPathGenerator pathGenerator, IPathRewri
         // Instruction is grouped by canvasId, so any can be used to generate canvas level ids
         var firstCanvasPaintingInCanvas = canvasInstruction.First();
         
-        // work out if any of the canvas painting records match a canvas in the current manifest
         if (firstCanvasPaintingInCanvas.CanvasOriginalId != null)
         {
             return manifestCanvasDictionary.Values.First(i =>
@@ -145,12 +146,11 @@ public class ManifestMerger(SettingsBasedPathGenerator pathGenerator, IPathRewri
         // At this stage we will be populating the paintingAnnotation with content from NQ manifest. 
         // Check to see if we can use an existing Canvas, falling back to creating a new one
         var canvasId = pathGenerator.GenerateCanvasId(firstCanvasPaintingInCanvas);
-        manifestCanvasDictionary.TryGetValue(firstCanvasPaintingInCanvas.Id, out var canvas);
-
-        if (canvas == null)
+        if (!manifestCanvasDictionary.TryGetValue(firstCanvasPaintingInCanvas.Id, out var canvas))
         {
             canvas = new Canvas { Id = canvasId };
         }
+        
         // modifies the id to match the rewritten path from settings
         else if (canvas.Id != canvasId)
         {
--- a/src/IIIFPresentation/Services.Tests/Manifests/Helpers/CanvasHelperTests.cs
+++ b/src/IIIFPresentation/Services.Tests/Manifests/Helpers/CanvasHelperTests.cs
@@ -1,0 +1,79 @@
+﻿using Core.Exceptions;
+using Repository.Paths;
+using Services.Manifests.Helpers;
+
+namespace Services.Tests.Manifests.Helpers;
+
+public class CanvasHelperTests
+{
+    [Fact]
+    public void CheckForProhibitedCharacters_AllowsString_WithNoProhibitedCharacters()
+    {
+        // Arrange
+        var stringToCheck = "valid";
+
+        // Act
+        Action action = () => CanvasHelper.CheckForProhibitedCharacters(stringToCheck);
+
+        // Assert
+        action.Should().NotThrow();
+    }
+    
+    [Theory]
+    [InlineData('/')]
+    [InlineData('=')]
+    [InlineData(',')]
+    public void CheckForProhibitedCharacters_ThrowsError_WithProhibitedCharacters(char invalidCharacter)
+    {
+        // Arrange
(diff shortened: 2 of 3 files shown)


src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
```diff
@@ -154,9 +151,26 @@ private Canvas GenerateCanvas(Dictionary<AssetId, Canvas> canvasDictionary,
                 canvasOrderCount);
 
             var isChoice = canvasOrderCount > 1;
+
+            var firstCanvasPainting = canvasOrderGroup.First();
```
donaldgray, line 155: What's happening here? Do we need a clarifying comment? If this is a choice is it enought to look at the first only, or do we need to consider everything in group?

Does this cover, from #433 
>We need to handle the case where there are a mix of CanvasPaintings; those that are to be added to the Manifest because they have an `AssetId` and those that have an `AssetId` but shouldn't be added to the Manifest (sees https://github.com/dlcs/iiif-presentation/issues/424, this will be due to the Canvas body pointing at an already managed Asset).

donaldgray (commented):
Couple of comments. I think `items` / `item` is overused in params and vars and more accurate naming might help clarify what's happening?

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -353,6 +345,13 @@ private async Task SaveToS3(DbManifest dbManifest, WriteManifestRequest request,
         bool canBeBuiltUpfront, CancellationToken cancellationToken)
     {
         var iiifManifest = request.RawRequestBody.FromJson<IIIF.Presentation.V3.Manifest>();
+        
+        var orderedCanvasPaintings = dbManifest.GetOrderedCanvasPaintings()?.ToList();
+
+        if (orderedCanvasPaintings is not null)
+        {
+            iiifManifest.Items = orderedCanvasPaintings.GenerateItems(pathGenerator, iiifManifest.Items, minimalCanvas: true);
```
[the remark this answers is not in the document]
donaldgray: I think it makes sense to call `GenerateItems()` here, and only here. It makes it obvious it's only called when written (so, once) rather than when converted.

If we only call it here then `GenerateItems()` can also order the canvasPaintings internally - the caller doesn't need to pre-order them?

```cs
iiifManifest.Items = canvasPaintingsInWhateverOrder.GenerateItems(pathGenerator, iiifManifest.Items);
```

src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
JackLewis-digirati: I'll add a comment to clarify, but if it's a choice, the id has to match, which is what we're interested in

how it works in this is purely based on the `canvasOriginalId` - if we get a match to one, then it returns the original item, otherwise it will use the asset id to generate, which I think is what we want?

donaldgray (changes requested):
As discussed on call, couple of issues to resolve

* Invalid `canvas.Id` on matched Canvas results in the Canvas being ignored (extra slug - https://localhost:7230/presentation/7/canvases/asimov)
* Receiving `"Error creating DLCS space"` when there was a general DLCS error. This was related to attempting to delete `null` AssetId (maybe one for another ticket)?
* General comment - too much reliance on `canvasOriginalId`. We can accept different forms of CanvasId, which we reduce to the "id" part but we join on `canvasOriginalId`, which could result in us joining 2 different forms.
* Empty annotationPages (`"items": []`) in a matched canvas breaks handling (check `.IsNullOrEmpty()` rather than `== null` somewhere?)
* `GenerateProvisionalCanvas()` - if matched canvas then we should add provisional `paintingAnnotation`. Linked to above, don't use canvasOriginalId for matching.

PR #444, opened 2025-08-08
Draft: #424 Identify Assets in Items

p-kaczynski:
- **Remove `IgnorePaintedResourcesWithItems` option as no longer needed**
- **Update PresentationManifestValidatorTests.cs**
- **Impl #424 - Recognize managed asset path in manifest items**
- **Adjust existing tests to new code**
- **Fix to previously expected functionality**

### DIFF
--- a/src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
@@ -10,10 +10,12 @@ using Core.Auth;
 using Core.Helpers;
 using Core.IIIF;
 using DLCS.Exceptions;
+using IIIF.Presentation.V3.Annotation;
 using IIIF.Serialisation;
 using Models.API.General;
 using Models.API.Manifest;
 using Models.Database.General;
+using Models.DLCS;
 using Repository;
 using Repository.Helpers;
 using Repository.Paths;
@@ -160,14 +162,18 @@ public class ManifestWriteService(
             // Ensure we have a manifestId
             manifestId ??= await GenerateUniqueManifestId(request, cancellationToken);
             if (manifestId == null) return ErrorHelper.CannotGenerateUniqueId<PresentationManifest>();
+
+            // This will throw ManifestException if some of the rules for managed assets in .Items are broken
+            var recognizedItemsAssets =
+                await dlcsManifestCoordinator.VerifyCanvasItemAssets(request, cancellationToken);
             
-            // Carry out any DLCS interactions (for paintedResources with _assets_) 
+            // Carry out any DLCS interactions
             var dlcsInteractionResult =
                 await dlcsManifestCoordinator.HandleDlcsInteractions(request, manifestId, cancellationToken: cancellationToken);
             if (dlcsInteractionResult.Error != null) return dlcsInteractionResult.Error;
             
             var (error, dbManifest) =
-                await CreateDatabaseRecord(request, parsedParentSlug, manifestId, dlcsInteractionResult.SpaceId, dlcsInteractionResult, cancellationToken);
+                await CreateDatabaseRecord(request, parsedParentSlug, manifestId, dlcsInteractionResult.SpaceId, dlcsInteractionResult, recognizedItemsAssets, cancellationToken);
             if (error != null) return error;
             
             var hasAssets = PaintedResourcesHasAssets(request);
@@ -187,7 +193,7 @@ public class ManifestWriteService(
 
     private static bool PaintedResourcesHasAssets(WriteManifestRequest request)
     {
-        return request.PresentationManifest.PaintedResources?.Any(x => x.Asset != null) == true;
+        return request.PresentationManifest.PaintedResources.HasAsset();
     }
 
     private async Task<PresUpdateResult> UpdateInternal(UpsertManifestRequest request,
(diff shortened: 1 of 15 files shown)


src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -160,14 +162,18 @@ private async Task<PresUpdateResult> CreateInternal(WriteManifestRequest request
             // Ensure we have a manifestId
             manifestId ??= await GenerateUniqueManifestId(request, cancellationToken);
             if (manifestId == null) return ErrorHelper.CannotGenerateUniqueId<PresentationManifest>();
+
+            // This will throw ManifestException if some of the rules for managed assets in .Items are broken
+            var recognizedItemsAssets =
+                await dlcsManifestCoordinator.VerifyCanvasItemAssets(request, cancellationToken);
             
-            // Carry out any DLCS interactions (for paintedResources with _assets_) 
+            // Carry out any DLCS interactions
```
JackLewis-digirati, line 170: this comment could maybe be updated - there are DLCS actions happening in the method above

src/IIIFPresentation/API/Features/Manifest/PaintableAssetIdentifier.cs
```diff
@@ -0,0 +1,108 @@
...
+public partial class PaintableAssetIdentifier(IHttpContextAccessor httpContextAccessor, IOptionsMonitor<TypedPathTemplateOptions> templateOptionsMonitor, ILogger<PaintableAssetIdentifier> logger)
+{
+    private TypedPathTemplateOptions TemplateOptions => templateOptionsMonitor.CurrentValue;
+    
+    /// <summary>
+    /// Get <see cref="AssetId"/> from <see cref="IPaintable"/>, if it represents an asset managed by this instance, otherwise <c>null</c>
+    /// </summary>
+    /// <param name="paintable">Paintable resource, can be null</param>
+    /// <param name="customerId">Id of the customer making the current request</param>
+    /// <returns></returns>
```
JackLewis-digirati, line 24: empty comment value

src/IIIFPresentation/API/Features/Manifest/PaintableAssetIdentifier.cs
```diff
@@ -0,0 +1,108 @@
...
+            throw new PresentationException("Image body and services point to different managed assets");
+
+        return fromServices ?? fromBody;
+    }
+
+    private AssetId? ResolveFromImageServices(List<IService>? services, int customerId)
+    {
+        if (services is null) return null;
+
+        foreach (var service in services)
```
JackLewis-digirati, line 62: could maybe do with wrapping {}? - it's a little hard to follow given the multiline switch

src/IIIFPresentation/API/Features/Manifest/PaintableAssetIdentifier.cs
```diff
@@ -0,0 +1,108 @@
...
+            logger.LogTrace("Asset path '{Path}' does not belong to customer '{CustomerId}'", uri.AbsolutePath, customerId);
+            return null;
+        }
+        
+        return new AssetId(customerId, int.Parse(canonicalMatch.Groups["space"].Value), canonicalMatch.Groups["asset"].Value);
+    }
+
+    private bool IsHostKnown(Uri uri) =>
+        uri.Host.Equals(httpContextAccessor.SafeHttpContext().Request.Host.Host)
+        || TemplateOptions.Overrides.ContainsKey(uri.Host);
```
JackLewis-digirati, line 98: what happens if it's in the `Defaults`, rather than `Overrides`?

src/IIIFPresentation/Services.Tests/Manifests/ManifestPaintedResourceParserTests.cs
```diff
@@ -85,7 +85,7 @@ public void Parse_Parses_WhenRewrittenCanvasId(string canvasId)
             ]
         };
 
-        var parsed = sut.ParseToCanvasPainting(manifest, CustomerId);
+        var parsed = sut.ParseToCanvasPainting(manifest, CustomerId, null!);
```
JackLewis-digirati, line 88: assuming this will be fixed in future tests?

src/IIIFPresentation/Services/Manifests/ManifestPaintedResourceParser.cs
```diff
@@ -18,7 +19,8 @@ public class ManifestPaintedResourceParser(
     IPresentationPathGenerator presentationPathGenerator,
     ILogger<ManifestPaintedResourceParser> logger) : ICanvasPaintingParser
 {
-    public IEnumerable<CanvasPainting> ParseToCanvasPainting(PresentationManifest presentationManifest, int customerId)
+    public IEnumerable<CanvasPainting> ParseToCanvasPainting(PresentationManifest presentationManifest, int customerId,
+        Dictionary<IPaintable, AssetId> _)
```
JackLewis-digirati, line 23: this feels like a code smell - can this just be removed?

src/IIIFPresentation/API/Features/Manifest/PaintableAssetIdentifier.cs
JackLewis-digirati: I think assets in the service will _only_ use the DLCS id, so really this should use the `OrchestratorUri` from the `DlcsSettings` instead of the details from `TypedPathTemplateOptions` (this gets paths like `https://presentation-api.something.com`, which it would never match with  - additionally, the [RFC](https://github.com/dlcs/iiif-presentation/blob/develop/docs/rfcs/0005-mixed-manifests.md#identify-managed-assets) says not to worry about rewritten paths for now

src/IIIFPresentation/API/Features/Manifest/PaintableAssetIdentifier.cs
```diff
@@ -0,0 +1,116 @@
...
+    private AssetId? Resolve(ExternalResource av, int customerId)
+    {
+        // AV are identified from body id only.
+        return ParseId(CanonicalAvRegex(), av.Id, customerId);
+    }
+    
+    private AssetId? Resolve(Image image, int customerId)
+    {
+        var fromServices = ResolveFromImageServices(image.Service, customerId);
+        var fromBody = ParseId(CanonicalImageRegex(), image.Id, customerId);
```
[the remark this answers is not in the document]
p-kaczynski: Don't we want to - if don't get from services - check the body id? Is that sth that's deemed "invalid" (and potentially sth else already checks that so it won't get to this point anyway)? To me the ticket read "it might be in the services, or it might be in the body.id, or maybe both - but then they have to match"

PR #456, opened 2025-08-27
in progress work

JackLewis-digirati:


src/IIIFPresentation/Services/Manifests/ManifestItemsParser.cs
```diff
@@ -174,22 +177,38 @@ private CanvasPainting CreatePartialCanvasPainting(ResourceBase? resource,
...
             var parsedCanvasId =
                 pathRewriteParser.ParsePathWithRewrites(canvasId.Host, canvasId.AbsolutePath, customerId);
-            CanvasHelper.CheckParsedCanvasIdForErrors(parsedCanvasId, canvasId.AbsolutePath);
-            return parsedCanvasId.Resource;
+
+            var checkedCanvasId =
+                CanvasHelper.CheckParsedCanvasIdForErrors(parsedCanvasId, canvasId.AbsolutePath, logger, false);
+            
+            // cannot match up with a painted resource, so don't set the canvas id
+            if (!(paintedResourceCanvasPaintings?.Any(cp => cp.Id == checkedCanvasId) ?? false))
```
donaldgray, line 206: Small - would it be clearer if this returned val if we _can_ match, else let it fall through? (only mentioning as it took me a sec to understand when scanning code).

`if isMatching return match` generally reads better than `if not isMatching return nothing else return match`

src/IIIFPresentation/Services/Manifests/ManifestItemsParser.cs
```diff
@@ -174,22 +177,38 @@ private CanvasPainting CreatePartialCanvasPainting(ResourceBase? resource,
...
+        List<CanvasPainting>? paintedResourceCanvasPaintings)
     {
         if (currentCanvas.Id == null) return null;
 
         if (!Uri.TryCreate(currentCanvas.Id, UriKind.Absolute, out var canvasId))
         {
-            CanvasHelper.CheckForProhibitedCharacters(currentCanvas.Id);
+            currentCanvas.Id = CanvasHelper.CheckForProhibitedCharacters(currentCanvas.Id, logger, false);
+            
+            if (currentCanvas.Id != null && !(paintedResourceCanvasPaintings?.Any(cp => cp.Id == currentCanvas.Id) ?? false))
```
donaldgray, line 189: `paintedResourceCanvasPaintings?.Any(cp => cp.Id == xx) ?? false` is repeated twice - make helper method?

Would allow for 
```cs
// if (currentCanvas.Id != null && !(paintedResourceCanvasPaintings?.Any(cp => cp.Id == currentCanvas.Id) ?? false))
if (currentCanvas.Id != null && !HasMatchingThing(currentCanvas.Id))

// if (!(paintedResourceCanvasPaintings?.Any(cp => cp.Id == checkedCanvasId) ?? false))
if (!HasMatchingThing(checkedCanvasId))
``

donaldgray (approved):

PR #458, opened 2025-09-03
Add canvas id parsing notes

JackLewis-digirati:
This PR adds some notes on how a canvas id is parsed from the payload.  It contains information from #456, so will not be fully accurate until that PR is merged

### DIFF
--- a/docs/notes/canvas-id-parsing.md
+++ b/docs/notes/canvas-id-parsing.md
@@ -1,0 +1,71 @@
+# Canvas Id parsing
+
+When parsing payloads, the `canvasId` can be used to generate a `canvasPainting` `id` record.  This can be accepted either directly from a `canvasPainting` `canvasId` and also from `items` `id` of a canvas.  In addition to the 2 locations that we can generate a `canvasId` from, we can accept either a short form (i.e.: just the id itself) or alternatively from a recognized URI.  Finally, this `canvasId` can be used to join a `canvas` declared in `items` with a `canvasPainting` record in order to decorate an asset from the DLCS with additional IIIF.   This document is set to explain the various ways this `canvasId` can be parsed from a payload.
+
+## Accepted formats
+
+As mentioned above, there are a few formats that will be accepted 
+
+
+| Format      | Example     |
+| ------------- | ------------- |
+| short canvas | `someId` |
+| general API URL | `https://presentation-api.com/1/canvas/someId` |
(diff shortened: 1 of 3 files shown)


docs/notes/canvas-id-parsing.md
```diff
@@ -0,0 +1,71 @@
...
+    fourth -- no --> fifth
+    fifth -- yes --> eleventh
+    fifth -- no --> sixth
+    fourth -- yes --> seventh
+    seventh -- items --> eighth
+    eighth -- no --> third
+    eighth -- yes --> ninth
+    seventh -- canvasPaintings --> ninth
+    ninth --> tenth
+    tenth -- no --> third
```
donaldgray, line 48: If we receive `canvasPainting.canvasId` and can't parse it shouldn't it be rejected, rather than us creating an alternative id?

donaldgray (commented):
Flow chart is helpful.

Is it worth having concrete examples to show different permutations and what the result would be?

docs/notes/canvas-id-parsing.md
donaldgray: This is going down the PaintedResources path though (unless I'm reading this wrong)

```
where is this canvas id from? -> canvasPaintings -> parse path with rewrites -> is the path parsed -> no`

seventh -- canvasPaintings --> ninth --> tenth --> no                       
```

docs/notes/canvas-id-parsing.md
```diff
@@ -0,0 +1,198 @@
...
+    eighth -- yes --> fifth
+    eighth -- no --> ninth
+```
+
+There is a slight difference between the id being parsed from `items` versus `canvasPaintings`, in that `items` has an additional check for "is a recognised host".  This is because `items` needs to be slightly tighter than `canvasPaintings` as the `canvasId` will be _generated_ for the `canvasPainting` table, but the id in the `items`will be left alone, with a `canvasOriginalId` added to the `canvasPainting` record.  This ultimately helps to avoid rejecting payloads that are purely IIIF that have been copied around from another customer.  This check is essentially checking that the passed URL is either the general API URL or the customer specific URL.
+
+The "is recognised host" check depends on the below settings to recognise a host:
+
+```json
+"PresentationApiUrl": "https://presentation-api.com/1/canvas/someId",
```
donaldgray, line 89: These are just the root URIs, no path

PR #459, opened 2025-09-03
Allow reingest property to be deserialized, but not serialized

### LINKED ISSUE #446: Don't return `"reingest"` property in payloads
API consumers can specify the `"reingest"` property in a paintedResources payload to instruct the API to reingest the asset, regardless of whether it existed or not, added in #355 .

This is currently being returned in paintedResources payload with the default `false` value. This shouldn't be in the return payload as it doesn't mean anything - we don't know whether the asset was reingested or not, it will always be false.

In summary - returned payloads shouldn't have the `"reingest"` property

JackLewis-digirati:
Resolves #446 

This PR makes it so that the `reingest` property will still be accepted, but won't show up on the response body.

### DIFF
--- a/src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
+++ b/src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
@@ -50,13 +50,7 @@ public class PaintedResource
     [JsonProperty(Order = 3)]
     public JObject? Asset { get; set; }
     
-    public bool Reingest { get; set; }
-    
-    // Newtonsoft method to stop serialization, but allow deserialization of the reingest property
-    public bool ShouldSerializeReingest()
-    {
-        return false;
-    }
+    public bool? Reingest { get; set; }
 }
 
 public class CanvasPainting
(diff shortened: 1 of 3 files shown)


src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
```diff
@@ -51,6 +51,12 @@ public class PaintedResource
     public JObject? Asset { get; set; }
     
     public bool Reingest { get; set; }
+    
+    // Newtonsoft method to stop serialization, but allow deserialization of the reingest property
+    public bool ShouldSerializeReingest()
```
donaldgray, line 56: Would `public bool? Reingest { get; set; }` result in the same without tying to particular ser/deser lib?

This'll work now (and likely for the near future) as iiif-net `.AsJson()` is used for ser but if we updated that to STJ the `Reingest` field would start appearing again.

JackLewis-digirati: yeah, makes sense to use a nullable bool - will modify this

PR #469, opened 2025-09-17
Add migration to resolve #467 - root collection updated for all customers unintentionally

### LINKED ISSUE #467: `root` collection `modified` column gets updated when it shouldn't
## Observed incorrect behaviour
Requesting root collections for different customers returns the same Etag, causing incorrect 304 responses.

When logging to the frontend UI in the same browser as first one, then a different customer, requesting the root collection for the second one receives 304, causing the response body to be determined unchanged, presenting the first customer's root collection when logged in as a different one.

While uncommon in real-world user scenario, this was not intended and causes issues when testing.

## Behaviour details
`root` collections for all customers on DEV have the same Etag, hence when request is made with Etag of a different customer's root collection, it causes 304 response, as Etag matches the `If-None-Match` provided one.

This stems from them all having the same `modified` timestamp, used to calculate the Etag. Having root collections be identical to a fraction of a second between two customer's collections is extremely unlikely, and all of them having the same one is - assuming realistic and correct behaviour - impossible.

## Identified issue

`modified` is bumped - specifically to change Etag - whenever a direct descendant of it is added or removed, as this modifies the collection "indirectly".

This is achieved by a trigger:
```
CREATE TRIGGER storage_collection_bump_trigger
  AFTER INSERT OR UPDATE OR DELETE ON hierarchy
  FOR EACH ROW
    EXECUTE FUNCTION storage_collection_bump();
```

`storage_collection_bump` performed an update on `collections`:
```
UPDATE collections
  SET modified = now()
  WHERE id = ANY (affected_ids);
```
The assumption made here is incorrect, because `id` is NOT an unique key for `collections` - the actual unique key is a composite of `(id, customer_id)`. And while _most_ collections have uniquely generated `id`s, this wasn't immediately obvious in testing.

However, the "root" collection, representing the `/` hierarchical collection, always has `id` set to `root`. Therefore any additon/move/deletion of a hierarchical item inside their `root` collection by any customer caused an update:
```
UPDATE collections
  SET modified = now()
  WHERE id = ANY (["root");
```
updating `modified` to `now()` for EVERY customer's root collection, and as both the `id` and `modified` was therefore identical, every `root` collection had the same Etag

p-kaczynski:
* update function to limit side effects to the same customer to the same customer
For more details refer to https://github.com/dlcs/iiif-presentation/issues/467 - full diagnosis and solution description available there.

This PR contains a migration that will perform an update of the function storage_collection_bump that had a flaw in its UPDATE condition.

Migration has Up with the new, corrected text, and Down which reverts to what it is currently.

Note that declared variable cus_id is always set along with the affected_ids, and only used within  IF affected_ids IS NOT NULL - hence separate condition is not necessary.


src/IIIFPresentation/Repository/Migrations/20250916153416_updateStorageCollectionBumpTriggerFunction.cs
```diff
@@ -0,0 +1,118 @@
...
+
+namespace Repository.Migrations
+{
+    /// <inheritdoc />
+    public partial class updateStorageCollectionBumpTriggerFunction : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql(
```
JackLewis-digirati, line 13: is it worth writing a test around this bug? i.e.: creating/deleting a child resource only updates the parent root record?

JackLewis-digirati (approved):

PR #503, opened 2025-11-27
Consistent POST/PUT behaviour

### LINKED ISSUE #464: Consistent API PUT and POST behaviour
This ticket is to have a consistent, predictable method for allowing consumers to specify IIIF-Presentation specific data (e.g. slug, parent, identifier) when making HTTP requests. 

See netlify documentation: https://deploy-preview-2--dlcs-docs.netlify.app/api-doc/iiif#general-rules-for-create-and-update-http-request-bodies for narrative description of this.

> [!WARNING]
> Netlify documentation mentions `PATCH` but this is not implemented yet - we only care about `PUT` and `POST`

## API Endpoints

For clarity, the rules for PUT and POST are:
* PUT - you PUT _to_ a resource location. This will insert or update that resource. The URL you PUT to is the resource you are creating.
* POST - you POST _into_ a container. This will insert only, if the resource already exists request will fail. The URL you POST to is the parent of the resource you are creating.

We want to support PUT and POST for both Flat and Hierarchical endpoints. Therefore we should have:
* `POST /{customer}/collections` - create flat collection
* `PUT /{customer}/collections/{id}` - upsert flat collection
* `POST /{customer}/manifests` - create flat manifest
* `PUT /{customer}/manifests/{id}` - upsert flat manifest 
* `POST {customer}/{parent-path}` - create hierarchical resource
* `PUT {customer}/{parent-path}/{slug}` - upsert hierarchical resource

> [!NOTE]
> For hierarchical paths we must interogate the `"type"` to determine whether it is Collection or Manifest.

## Properties

For each resource we are saving we need to know the internal `id`, `parent` and `slug`. There are various ways that the request can indicate what these values are, if we receive payloads that contain multiple expressions of the same property they must match or we will reject (e.g. if given `"slug"` explicitly and via URL they must be the same).

Below is a summary of where we can get each from.

### `id`
- For flat PUT the `id` is the final path segment of the URL
- For any PUT/POST the `"id"` field in the body can indicate the `id` if it is expressed in flat form.
- If we aren't given a specific `id` we generate one. This is the internal identifier, not all consumers will care.
  - Note that we may be given an `"id"` that is effectively random - it could be a copy/paste of an external resource. In which case we'll ignore the `"id"` value. See `"canvasId"` recognition for similar - this uses incoming hostname to differentiate between internal identifiers, that may carry meaning, from external ones, that are ignored.

### `parent`
- For flat PUT/POST this can be specified explicitly in the body. Accept flat or hierarchical form
- For flat PUT/POST this can be derived from `"publicId"` field
- For hierarchical POST this can be derived from the URL - the URL will indicate the parent we are creating resource in
- For hierarchical PUT this can be derived from the URL - the final path segment is the Id of resource we are creating, everything before that is the container.
- For any PUT/POST can be derived from the body `"id"` if it is expressed in hierarchical form.

### `slug`
- For flat PUT/POST this can be specified explicitly in the body
- For flat PUT/POST this can be derived from `"publicId"` field
- For hierarchical PUT this can be derived from the URL
- For any PUT/POST can be derived from the body `"id"` if it is expressed in hierarchical form.

## Prior Art

https://github.com/dlcs/iiif-presentation/issues/272 - ticket introduced `publicId` property and above logic for flat manifest endpoints.

## Acceptance Criteria

* Ensure the above 6 endpoints are implemented and we can PUT and POST Manifests and Collections using Flat or Hierarchical paths.
  * Flat endpoints should already exist.
  * Hierarchical POST exists for collections only but it is using incorrect logic - it creates the resource POSTed to, rather than treating this as parent.
* Ensure that all endpoints support expressing `id`, `parent` and `slug` as documented above. 
  * With exception of `"id"` handling this should already be implemented for Manifests.
  * I don't think `"id"` is handled in any scenario - we should look at it to determine if it expresses anything but keep to the same rules as to when we overwrite it or when we store it as-is.

p-kaczynski:
Implements #464

## Notes
This PR includes:
* New Collection hierarchical PUT endpoint
* Modified hierarchical Collection POST endpoint logic
* Modified Manifest PUT/POST flat endpoints logic
* New Manifest hierarchical POST/PUT endpoints
* Tests

As per the source ticket this includes an integration test that performs all allowed flat and hierarchical POST/PUTs for Manifests and Collections, providing the **required** data in a variety of ways.

To make the test pass, i.e. to implement that desired behaviour, the biggest change is to Collection handling.

### Controllers
For hierarchical POST/PUTs the type of the item being upserted is not specified by the path. As both POST and PUT might be creating new resource, we cannot rely on hierarchy-by-slug resolution to check the type. As per the ticket, we have to inspect the `type` property.

This is implemented via new utility `FastJsonPropertyRead` class, `FindAtLevel(string json, string targetPropertyName, int level = 1)` method.

This uses `Utf8JsonReader` to quickly read through the provided JSON **until** a selected property (here: `type`) is found at a specified nesting level (here: `1`, top level).

### Collections
Before: In hierarchical paths, only Collection POST implemented. URL path handling as if it was PUT. Only "pure" IIIF collections supported.

Now: `POST /path/to/parent` of either "pure" or "presentation" Collection supported. `PUT /path/to/resource` also supported for both types of Collections.

Necessary information (slug + parent) can be provided in many ways, which has been expressed in the `CollectionWriteRequestHandler`. The handling method has been extensively annotated to explain what is being done and why.

The general process is very similar in all scenarios, but at nearly every step small details vary, which makes finding an optimal way of implementing the desired logic very hard. The approach I took avoids code duplication and attempts to allow future changes/maintenance to be done easily via being able to follow the logic step-by-step, aided by the generous comments.

Essentially all logic previously spread between `CollectionController`, `StorageController`, `CreateCollection(Handler)`, `UpsertCollection(Handler)` and `PostHierarhicalCollection(Handler)` has been unified, primarily to exemplify the actual desired flow and logic of the concept of "creating or changing a collection", regardless of which of many, many possible ways of doing it was chosen.

#### Show-Extra handling
As the hierarchical endpoints provide "standard", IIIF storage capabilities, i.e. one that does not require knowledge of the extended properties added by `Presentation*` entities on top of the base IIIF models, making POST/PUTs does not **require** providing the custom header.

Likewise, it is expected that the response will be plain IIIF, without the extended properties. However, "under the hood", we work with `PresentationCollection`, as we will add various properties like `Slug` or `Parent`. To allow future-proof way of obtaining the base class `V3.Collection` from `PresentationCollection` (and same for manifests) there's a new class `PresentationIIIFCleaner`

This class uses a static constructor to perform one-per-application-lifetime reflection of the relevant models and creates a method that will first create an empty instance of the base model (e.g. `V3.Collection` and then rewrite all properties it can handle from the source `Presentation*` derived class.

Because the list of properties to "rewrite" is loaded on each startup, any new properties added to the base model will automatically be included, reliving developers from remembering to and adding manually a property rewrite.

#### Deletion
As it was at one point useful in testing, I have implemented the hierarchical DELETE as well, which currently just determines type of resource and fires one of already-existing requests for the specified resource. No change to the implementation of those requests was done, just the endpoint.

### Manifests
Before: Only flat paths. Assuming `ManifestWriteService` is "correct", hence leaving as-is.

Now: `POST /path/to/parent` and `PUT/path/to/parent/manifest_slug` implemented via `type` property inspection in `StorageController`.

The amount of combinations is rather large with both the method (POST/PUT) and the "style" (flat/hierarchical) affecting some details of the overall process.

This logic has been captured in `DispatchManifestRequest` which supersedes `CreateManifest` (flat POST) and `UpsertManifest` (flat PUT), and also handles hierarchical versions. In the end it determines values necessary for successful handling (parent, slug), setting them on the body if necessary, and also performs checks to catch up mismatch between path-derived and body values.

As with the `CollectionWriteRequest`, it is richly annotated with comments to explain logic/assumptions/decisions in detail sufficient for both review and future changes.

As the validation logic from controllers would have to be cloned (or pulled up class tree too high up), it has been moved to this new `DispatchManifestRequest`.


src/IIIFPresentation/API.Tests/Converters/PresentationIIIFCleanerTests.cs
```diff
@@ -0,0 +1,87 @@
...
+            Thumbnail =
+            [
+                new Image
+                {
+                    Id = "https://localhost/thumbs/12/23/blabla/full/143,200/0/default.jpg"
+                }
+            ],
+            Rights = "https://creativecommons.org/licenses/by/4.0/"
+            // outside IIIF
+            ,
```
JackLewis-digirati: formatting?

src/IIIFPresentation/API.Tests/Integration/RestPathBehaviourTests.cs
```diff
@@ -0,0 +1,281 @@
...
+
+    public RestPathBehaviourTests(StorageFixture storageFixture, RestPathFixture fixture)
+    {
+        dbContext = storageFixture.DbFixture.DbContext;
+
+        amazonS3 = storageFixture.LocalStackFixture.AWSS3ClientFactory();
+
+        httpClient = fixture.httpClient;
+        // parent = dbContext.Collections
+        //     .First(x => x.CustomerId == Customer && x.Hierarchy!.Any(h => h.Slug == string.Empty)).Id;
```
JackLewis-digirati: commented out code

src/IIIFPresentation/API.Tests/Integration/RestPathBehaviourTests.cs
```diff
@@ -0,0 +1,281 @@
...
+    private int _slugCounter = 1;
+    private string UniqueSlug() => $"{Slug}-{_slugCounter++}";
+
+    public CollectionPathTestProvider()
+    {
+        // ( method,  url,  id,  parent,  slug,  publicId)
+        
+        // POST to flat collections endpoint, use FLAT parent
+        var slug = UniqueSlug();
+        Add(HttpMethod.Post.ToString(),Collections,null,Collections+ParentId, slug,null, slug);
```
JackLewis-digirati: missing space

src/IIIFPresentation/API.Tests/Integration/RestPathBehaviourTests.cs
```diff
@@ -0,0 +1,281 @@
...
+        
+        // POST into the parent public hierarchical collection
+        slug = UniqueSlug();
+        Add(HttpMethod.Post.ToString(),$"{Grandparent}/{Parent}", null, null, slug,null, slug);
+        
+        // PUT to the FLAT API Collections endpoint (2 variants, hier/flat parent)
+        slug = UniqueSlug();
+        Add(HttpMethod.Put.ToString(), Collections + Id + "-1", null, $"{Grandparent}/{Parent}", slug, null, slug);
+        slug = UniqueSlug();
+        Add(HttpMethod.Put.ToString(), Collections + Id + "-2", null, Collections+ParentId, slug, null, slug);
```
JackLewis-digirati: string interpolation feels cleaner to me + same for above

src/IIIFPresentation/API.Tests/Integration/RestPathBehaviourTests.cs
```diff
@@ -0,0 +1,281 @@
...
+        Add(HttpMethod.Put.ToString(), Manifests + Id, Manifests + Id, Collections + ParentId, Slug,
+            $"/{Grandparent}/{Parent}/{Slug}");
+        Add(HttpMethod.Put.ToString(), Manifests + Id, Manifests + Id, $"{Grandparent}/{Parent}", Slug,
+            $"/{Grandparent}/{Parent}/{Slug}");
+
+        // API POST
+        Add(HttpMethod.Post.ToString(), Manifests, Manifests + Id, null, null,
+            $"{Grandparent}/{Parent}/{Slug}");
+        Add(HttpMethod.Post.ToString(), Manifests, Manifests + Id, Collections + ParentId, Slug, null);
+        Add(HttpMethod.Post.ToString(), Manifests, Manifests + Id, $"{Grandparent}/{Parent}", Slug, null);
```
JackLewis-digirati: seems to mix string interpolation and just adding values together, along with `Parent`/`Grandparent` needing a forward slash, whereas it's built into `manifest`/`collection` - this might be easier to follow if it was all rationalized

src/IIIFPresentation/API/Converters/FastJsonPropertyRead.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System.Text;
+using System.Text.Json;
+
+namespace API.Converters;
+
+public static class FastJsonPropertyRead
```
JackLewis-digirati: the `fast` feels a little superfluous? maybe better as `JsonPropertyReader` or similar? (happy to disagree on this though)

src/IIIFPresentation/API/Features/Storage/CollectionController.cs
```diff
@@ -67,61 +63,23 @@ await Mediator.Send(new GetCollection(customerId, id, Request.Headers.IfNoneMatc
 
     [Authorize]
     [HttpPost("collections")]
```
JackLewis-digirati: should this be changed to match the manifests controller (or the opposite)?

src/IIIFPresentation/API/Features/Manifest/Requests/DispatchManifestRequest.cs
```diff
@@ -0,0 +1,334 @@
...
+    public string RawRequestBody { get; set; } = rawRequestBody;
+    public bool IsHierarchical { get; set; } = isHierarchical;
+    public bool IsShowExtras { get; set; } = isShowExtras;
+    public bool IsCreateSpace { get; set; } = isCreateSpace;
+    public string? ETag { get; set; } = eTag;
+}
+
+public class DispatchManifestRequestHandler(
+    PresentationContext dbContext,
+    ILogger<CollectionWriteService> logger,
```
JackLewis-digirati: should be `ILogger<DispatchManifestRequestHandler>`

src/IIIFPresentation/API/Features/Manifest/Requests/DispatchManifestRequest.cs
```diff
@@ -0,0 +1,334 @@
...
+        // POST -> Create
+        // PUT -> Upsert
+
+        // An id might either be one of existing manifest, or a "desired" id, if it's provided as a valid flatId
+        // A null is also acceptable - if so, we will treat it as "try creating new manifest with minted id"
+        string? manifestId;
+
+        // Note: handling of `publicId` prop is done by MWS - don't think we need to do anything with it here.
+
+        if (request.RequestMethod == HttpMethod.Post)
```
JackLewis-digirati: this method is pretty large - might be worth encapsulating `POST`/`PUT` into their own methods

src/IIIFPresentation/API/Features/Manifest/Requests/DispatchManifestRequest.cs
```diff
@@ -0,0 +1,334 @@
...
+/// Common encapsulation of data available in a controller action, including the caller's intent
+/// </summary>
+/// <param name="customerId">Caller's customer id - assumed verified by authentication</param>
+/// <param name="requestMethod">HTTP method used by caller, which implies intent (create at parent vs. upsert resource at location)</param>
+/// <param name="requestPath">Entire path AFTER /{customerId} - might be flat, providing resource type and id, might be hierarchical showing parent and possibly the intended/desired slug</param>
+/// <param name="rawRequestBody">Raw string data - will be parsed, validated etc.</param>
+/// <param name="isHierarchical">Could be deduced from <paramref name="requestPath"/>, but this value is known by controller, so to simplify code it is provided explicitly</param>
+/// <param name="isShowExtras">Whether the request included the "show extra properties" header</param>
+///// <param name="initiatingController">Controller which created and sent this request. Used to create a response.</param>
+public class DispatchManifestRequest(
```
JackLewis-digirati: why `Dispatch` over `Upsert`?

donaldgray:
Closing - superseded by #641

PR #505, opened 2025-12-01
Fix label from items to work correctly with a matched canvas

### LINKED ISSUE #501: `canvasLabel` and `label` validation not matching the manifest merger behavior in the canvas painting merger
During testing it was noticed that the following payload wasn't working

```
{
  "type": "Manifest",
  "label": {
    "en": [
      "Mixed Manifest Tests - Painted resource with matched canvas, and a matching label"
    ]
  },
  "behavior": [
    "paged"
  ],
  "items": [
    {
      "id": "example",
      "type": "Canvas",
      "label": {
        "en": [
          "my-label"
        ]
      }
    }
  ],
  "paintedResources": [
    {
      "asset": {
        "id": "example",
        "mediaType": "image/jpeg",
        "origin": ""
      },
      "canvasPainting": {
        "canvasId": "painted-resource-example",
        "canvasOrder": 0,
        "label": {
          "en": [
            "my-label"
          ]
        }
      }
    }
  ],
  "slug": "0009,
  "parent": "root"
}
```

where it was getting back a bad request due to not matching label as `item` sets `canvasLabel`, but `paintedResource` sets `label`, while also defaulting `canvasLabel` to the `canvasLabel` of `item` when merging canvas paintings

This is different behaviour to that discussed in the [docs around `canvasLabel` and `label`](https://deploy-preview-2--dlcs-docs.netlify.app/api-doc/iiif#canvaspainting) where the `label` value _should_ have been picked up for use as `canvasLabel`

Additionally, the validation needs to take into account "the first non-null value for label and canvasLabel for matching", whereas currently matching _ONLY_ occurs on the equivalent `canvasPainting` from `item`

see #393 for further details on how this was fixed in the manifest merger

JackLewis-digirati:
Resolves #501

This PR makes it so that `canvasLabel` and `label` work consistently across `items` and `canvasPainting` when in a matched canvas.  

Essentially, the `items.label` value will act like `canvasPainting.CanvasLabel`, and as such there is a removed check on `label` in the `canvasPaintingMerger` and will override the first `canvasPainting.CanvasLabel` if no `CanvasLabel` is set

See additional scenarios for how this will work in the above bug ticket

### DIFF
--- a/src/IIIFPresentation/API/Features/Manifest/CanvasPaintingMerger.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/CanvasPaintingMerger.cs
@@ -94,7 +94,7 @@ public class CanvasPaintingMerger(IPathRewriteParser pathRewriteParser) : ICanva
         {
             var matchedPaintedResourceCanvasPaintings = paintedResourceCanvasPaintings.Where(cp =>
                 !string.IsNullOrEmpty(cp.Id) && string.Equals(cp.Id, itemsCanvasPainting.Id,
-                    StringComparison.OrdinalIgnoreCase)).OrderInterimCanvasPaintings().ToList();
+                    StringComparison.OrdinalIgnoreCase)).OrderCanvasPaintings().ToList();
 
             if (matchedPaintedResourceCanvasPaintings.Count == 0) continue;
             
@@ -168,7 +168,7 @@ public class CanvasPaintingMerger(IPathRewriteParser pathRewriteParser) : ICanva
 
     private void ValidateItemCanvasPainting(InterimCanvasPainting itemsCanvasPainting,
         InterimCanvasPainting paintedResourceCanvasPainting,
-        List<Canvas>? items, int currentCanvasOrder, bool isMultiCanvas, LanguageMap? firstCanvasLabelInMultiItemCanvas)
+        List<Canvas>? items, int currentCanvasOrder, bool isMultiItemCanvas, LanguageMap? firstCanvasLabelInMultiItemCanvas)
     {
         var canvas = items?.FirstOrDefault(c =>
             pathRewriteParser.ParsePathWithRewrites(c.Id, itemsCanvasPainting.CustomerId).Resource ==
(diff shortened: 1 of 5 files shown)


src/IIIFPresentation/API/Features/Manifest/CanvasPaintingMerger.cs
```diff
@@ -92,7 +93,8 @@ private void JoinPaintedResourcesWithItems(List<InterimCanvasPainting> itemsCanv
         {
             var matchedPaintedResourceCanvasPaintings = paintedResourceCanvasPaintings.Where(cp =>
                 !string.IsNullOrEmpty(cp.Id) && string.Equals(cp.Id, itemsCanvasPainting.Id,
-                    StringComparison.OrdinalIgnoreCase)).ToList();
+                    StringComparison.OrdinalIgnoreCase)).OrderBy(cp => cp.CanvasOrder)
```
donaldgray: I think `.OrderBy(cp => cp.CanvasOrder).ThenBy(cp => cp.ChoiceOrder)` is used in a few places - is it worth having this as a reusable method?

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingMerger.cs
```diff
@@ -169,31 +182,35 @@ private void ValidateItemCanvasPainting(InterimCanvasPainting itemsCanvasPaintin
                     $"Canvas painting with id {itemsCanvasPainting.Id} cannot contain an annotation body");
             }
         }
-
-        if (paintedResourceCanvasPainting.CanvasLabel == null && itemsCanvasPainting.CanvasLabel != null)
+        
+        if (paintedResourceCanvasPainting.CanvasLabel == null &&
```
donaldgray: Comment explaining? Or can it be extracted into a method? `if (SomethingExplanatory()) { ... }`

src/IIIFPresentation/Services/Manifests/Helpers/InterimCanvasPaintingX.cs
```diff
@@ -88,4 +88,8 @@ public static List<CanvasPainting> ConvertInterimCanvasPaintings(
         this List<InterimCanvasPainting>? interimCanvasPaintings) =>
         interimCanvasPaintings?.Where(icp =>
             icp is { SuspectedAssetId: not null, CanvasPaintingType: CanvasPaintingType.Items }).ToList();
+
+    public static IEnumerable<InterimCanvasPainting> OrderInterimCanvasPaintings(
```
donaldgray: `CanvasOrder` and `ChoiceOrder` aren't specific to `InterimCanvasPainting`, neither is the logic to `.OrderBy(cp => cp.CanvasOrder).ThenBy(cp => cp.ChoiceOrder)`, those are relevant to `CanvasPainting` too.

If this method was also usable for `CanvasPainting` then it would have more utility and could be used https://github.com/dlcs/iiif-presentation/blob/develop/src/IIIFPresentation/Services/Manifests/ManifestMerger.cs#L101-L102 and https://github.com/dlcs/iiif-presentation/blob/develop/src/IIIFPresentation/API/Converters/ManifestConverter.cs#L173-L174 (or could replace that `ManifestConverter` method entirely).

PR #528, opened 2026-01-12
Fix to allow canvas paintings with a short canvas id to output iiif properties

### LINKED ISSUE #522: Async processed Manifests lose IIIF Canvas properties if using short Id
Submitting the following payload should result in a final Manifest that contains a single canvas, that canvas should have `"metadata"` and `"rendering"` properties, and an annotation page containing content resources for `page_1` asset from Protagonist:

```json
{
    "items": [
        {
            "type": "Canvas",
            "id": "alpha",
            "rendering": [
                { 
                    "id": "https://customer.example/svg/first",
                    "type": "Image",
                    "label": { "en": ["SVG XML for page text"] },
                    "format": "image/svg+xml"
                }
            ],
            "metadata": [
                {
                    "label": { "en": [ "Date" ] },
                    "value": { "en": [ "1941-1954" ] }
                }
            ]
        }
    ],
    "paintedResources": [
        {
            "canvasPainting": {
                "canvasId": "alpha"
            },
            "asset": {
                "id": "page_1",
                "origin": "https://example.com/photos/book-page.jpg",
                "mediaType": "image/jpeg"
            }
        }
    ]
}
```

However, it results in the `"rendering"` and `"metadata"` values effectively being ignored - both in the returned provisional canvases and in the final generated manifest.

After some troubleshooting it was identified that the issue was due to the use of short canvasIds (I think for the Canvas only but confirm) but _only if the Manifest couldn't be returned immediately_. If using fully qualified ids, or returning immediately, the expected behaviour was observed - provisional canvases and the final manifest contained the `"rendering"` and `"metadata"` props.

Related tickets
* https://github.com/dlcs/iiif-presentation/issues/430
* https://github.com/dlcs/iiif-presentation/issues/432

-----
## Troubleshooting notes

The issue looks related to how the `CanvasId:Canvas` lookup is built when generating the provisional canvases, https://github.com/dlcs/iiif-presentation/blob/82ac46d09f68a9359d80aa00a4beca2fa68d084d/src/IIIFPresentation/API/Converters/ManifestConverter.cs#L96-L97 the method called, https://github.com/dlcs/iiif-presentation/blob/15bcccfe2282ddfab97c1f14b9ab4e1cbebb84cf/src/IIIFPresentation/Repository/Paths/PathRewriteParser.cs#L78-L86 falls back to returning `new PathParts(null, null, true);` if the passed id is _not_ a URI. This results in the lookup having `null` for CanvasId, meaning that when we later try to find it https://github.com/dlcs/iiif-presentation/blob/82ac46d09f68a9359d80aa00a4beca2fa68d084d/src/IIIFPresentation/API/Converters/ManifestConverter.cs#L142 nothing is found so a new Canvas is created.

Thoughts
* Should `ParsePathWithRewrites` return `new PathParts(null, uri, true);` rather than `new PathParts(null, null, true);`? What are the side effects of this as there are multiple uses of this.
* Can `canvasIdFromCanvases` key default to canvasId if `.Resource` is null? (ie `canvasId: pathRewriteParser.ParsePathWithRewrites(c.Id, canvasPainting.CustomerId).Resource ?? canvasId`? If so does this need echoed elsewhere?
* Should the Id be a complex type that can contain long + short values and have appropriate comparison operators?

### LINKED ISSUE #526: Matching canvas loses standard IIIF values on background-handler completion
Using the following payload to create a manifest results in a final manifest that is missing the `"seeAlso"`, `"rendering"` and `"metadata"` properties. 

The standard IIIF properties are present on the provisional canvases but are _not_ present on the final Manifest. This only seems to happen on asynchronously completed manifests, if the Manifest can be completed by API the returned Manifest has all properties.

```json
{
    "type": "Manifest",
    "parent": "https://presentation-api.dlcs/7/collections/mixed_manifest_coll",
    "slug": "bug_single1_2",
    "items": [
        {
            "type": "Canvas",
            "id": "https://presentation-api.dlcs/presentation/7/canvases/bug_single1_2",
            "label": { "en": [ "foo" ] },
            "seeAlso": [
                {
                    "id": "https://iiif.example/text/alto/item",
                    "type": "Dataset",
                    "profile": "http://www.loc.gov/standards/alto/v3/alto.xsd",
                    "label": {
                        "none": [
                            "METS-ALTO XML"
                        ]
                    },
                    "format": "application/xml+alto"
                }
            ],
            "rendering": [
                {
                    "id": "https://iiif.example/text/svg/item",
                    "type": "Image",
                    "label": {"en": [ "SVG XML for page text" ] },
                    "format": "image/svg+xml"
                }
            ],
            "metadata": [ 
                {
                    "label": {"none": ["Bugcheck"]},
                    "value": {"fr": ["Extra"]}
                }
            ]
        }
    ],
    "paintedResources": [
        {
            "canvasPainting": {
                "canvasId": "https://presentation-api.dlcs/presentation/7/canvases/bug_single1_2"
            },
            "asset": {
                "id": "single1_b",
                "origin": "https://burst.shopifycdn.com/photos/wood-handled-espresso-machine.jpg",
                "mediaType": "image/jpeg"
            }
        }
    ]
}
```

Provisional canvases are returned as expected, showing the processing annotation and also the custom elements:

```json
"items": [
    {
        "id": "https://dlcs.example/presentation/7/canvases/bug_single1_2",
        "type": "Canvas",
        "label": {...},
        "metadata": [...],
        "rendering": [...],
        "seeAlso": [...],
        "items": [
            {
                "id": "https://presentation-api.dlcs/presentation/7/canvases/bug_single1_2/annopages/0",
                "type": "AnnotationPage",
                "items": [
                    {
                        "id": "https://presentation-api.dlcs/presentation/7/canvases/bug_single1_2/annotations/0",
                        "type": "Annotation",
                        "motivation": "painting",
                        "behavior": [
                            "processing"
                        ],
                        "target": "https://presentation-api.dlcs/presentation/7/canvases/bug_single1_2"
                    }
                ]
            }
        ]
    }
]
```

but when the Manifest is complete the properties are lost or overwritten

```json
"items": [
    {
        "id": "https://presentation-api.dlcs/presentation/7/canvases/single1_1",
        "type": "Canvas",
        "label": {...},
        "width": 3648,
        "height": 5472,
        "thumbnail": [...],
        "items": [...]
    }
]
```

~Unsure when these are being lost - are they being persisted to the staging version and subsequently lost? Or are they being replaced as part of final manifest creation?~ edit: checked an in-flight Manifest containing Audio (as these are slower to process) and confirmed that the staging version contains the custom IIIF.

JackLewis-digirati:
Resolves #522
Resolves #526

Makes it so that short canvas id will correctly output IIIF properties by adding a fallback value

> [!NOTE]
> This PR also fixes a masked issue around overrides not being correctly set in the background handler, and adds some additional logging

### DIFF
--- a/src/IIIFPresentation/Repository/Paths/PathRewriteParser.cs
+++ b/src/IIIFPresentation/Repository/Paths/PathRewriteParser.cs
@@ -82,7 +82,7 @@ public class PathRewriteParser(IOptions<TypedPathTemplateOptions> options, ILogg
             return ParsePathWithRewrites(uriResult.Host, uriResult.AbsolutePath, customer);
         }
 
-        return new PathParts(null, null, true);;
+        return new PathParts(null, null, true);
     }
 
     private static (int customerId, string? resourceId) MatchValuesInTemplate(string[] pathSplit,
--- a/src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
+++ b/src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
@@ -142,25 +142,8 @@ public class ManifestMerger(SettingsBasedPathGenerator pathGenerator, IPathRewri
                 i.Id! == firstCanvasPaintingInCanvas.CanvasOriginalId.ToString());
         }
         
-        // At this stage we will be populating the paintingAnnotation with content from NQ manifest. 
-        // Check to see if we can use an existing Canvas, falling back to creating a new one
-        var canvasId = pathGenerator.GenerateCanvasId(firstCanvasPaintingInCanvas);
-        if (!manifestCanvasDictionary.TryGetValue(firstCanvasPaintingInCanvas.Id, out var canvas))
-        {
-            logger.LogDebug("Couldn't find a canvas matched in the canvas paintings for the id {CanvasPaintingId}. Creating a new one", firstCanvasPaintingInCanvas.Id);
-            canvas = new Canvas { Id = canvasId };
-        }
-        // modifies the id to match the rewritten path from settings
-        else if (canvas.Id != canvasId)
-        {
-            logger.LogDebug("Found a matched canvas in the canvas paintings, but the id needs rewriting: actual - {CanvasId}, rewritten - {CanvasPaintingId}", canvas.Id, canvasId);
-            canvas.Id = canvasId;
-        }
-        else
-        {
-            logger.LogDebug("Found a matched canvas in the canvas paintings for id {CanvasPaintingId}", firstCanvasPaintingInCanvas.Id);
-        }
-        
+        var canvas = GetWorkingCanvas(manifestCanvasDictionary, firstCanvasPaintingInCanvas);
+
         // Instantiate a new AnnotationPage - this is what we'll populate with PaintingAnnotations below
         var annoPage = new AnnotationPage
         {
@@ -204,6 +187,30 @@ public class ManifestMerger(SettingsBasedPathGenerator pathGenerator, IPathRewri
         }
         
         canvas.Items = [annoPage];
+        return canvas;
+    }
+
+    private Canvas GetWorkingCanvas(Dictionary<string, Canvas> manifestCanvasDictionary, CanvasPainting firstCanvasPaintingInCanvas)
+    {
+        // At this stage we will be populating the paintingAnnotation with content from NQ manifest. 
+        // Check to see if we can use an existing Canvas, falling back to creating a new one
+        var canvasId = pathGenerator.GenerateCanvasId(firstCanvasPaintingInCanvas);
+        if (!manifestCanvasDictionary.TryGetValue(firstCanvasPaintingInCanvas.Id, out var canvas))
+        {
+            logger.LogDebug("Couldn't find a canvas for the id {CanvasPaintingId}. Creating a new one", firstCanvasPaintingInCanvas.Id);
+            canvas = new Canvas { Id = canvasId };
+        }
+        // modifies the id to match the rewritten path from settings
+        else if (canvas.Id != canvasId)
+        {
+            logger.LogDebug("Found a matched canvas, but the id needs rewriting: actual - {CanvasId}, rewritten - {CanvasPaintingId}", canvas.Id, canvasId);
+            canvas.Id = canvasId;
+        }
+        else
+        {
+            logger.LogDebug("Found a canvas for the id {CanvasPaintingId}", firstCanvasPaintingInCanvas.Id);
+        }
+
         return canvas;
     }
 
--- a/src/IIIFPresentation/API/Program.cs
+++ b/src/IIIFPresentation/API/Program.cs
@@ -16,6 +16,7 @@ using Newtonsoft.Json;
 using Repository;
 using Repository.Paths;
 using Serilog;
+using Services;
 using Services.Manifests;
 using Services.Manifests.AWS;
 using Services.Manifests.Helpers;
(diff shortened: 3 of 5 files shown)


src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
```diff
@@ -147,14 +147,19 @@ private Canvas GenerateCanvas(Dictionary<AssetId, Canvas> canvasDictionary,
         var canvasId = pathGenerator.GenerateCanvasId(firstCanvasPaintingInCanvas);
         if (!manifestCanvasDictionary.TryGetValue(firstCanvasPaintingInCanvas.Id, out var canvas))
         {
+            logger.LogDebug("Couldn't find a canvas matched in the canvas paintings for the id {CanvasPaintingId}. Creating a new one", firstCanvasPaintingInCanvas.Id);
```
donaldgray: `"Couldn't find a canvas matched in the canvas paintings"` <- looking at `Dictionary<string, Canvas> manifestCanvasDictionary`, those aren't canvas paintings yet, they're just `Canvas` from manifest?

Also - is it worth considering moving lines 147..162 into a method? IMO it would make `GenerateCanvas()` a little easier to read, and it would keep all the "what Canvas should I use" logic in one single place.

e.g.
```cs
if (firstCanvasPaintingInCanvas.CanvasOriginalId != null)
{
    return manifestCanvasDictionary.Values.First(i =>
        i.Id! == firstCanvasPaintingInCanvas.CanvasOriginalId.ToString());
}

// At this stage we will be populating the paintingAnnotation with content from NQ manifest. 
// Check to see if we can use an existing Canvas, falling back to creating a new one
var canvasId = pathGenerator.GenerateCanvasId(firstCanvasPaintingInCanvas);
if (!manifestCanvasDictionary.TryGetValue(firstCanvasPaintingInCanvas.Id, out var canvas))
{
    logger.LogDebug("Couldn't find a canvas matched in the canvas paintings for the id {CanvasPaintingId}. Creating a new one", firstCanvasPaintingInCanvas.Id);
    canvas = new Canvas { Id = canvasId };
}
// modifies the id to match the rewritten path from settings
else if (canvas.Id != canvasId)
{
    logger.LogDebug("Found a matched canvas in the canvas paintings, but the id needs rewriting: actual - {CanvasId}, rewritten - {CanvasPaintingId}", canvas.Id, canvasId);
    canvas.Id = canvasId;
}
else
{
    logger.LogDebug("Found a matched canvas in the canvas paintings for id {CanvasPaintingId}", firstCanvasPaintingInCanvas.Id);
}

// Instantiate a new AnnotationPage - this is what we'll populate with PaintingAnnotations below
var annoPage = new AnnotationPage
{
    Id = pathGenerator.GenerateAnnotationPagesId(firstCanvasPaintingInCanvas),
    Items = []
};
```

could become something like:

```cs
if (firstCanvasPaintingInCanvas.CanvasOriginalId != null)
{
    return manifestCanvasDictionary.Values.First(i =>
        i.Id! == firstCanvasPaintingInCanvas.CanvasOriginalId.ToString());
}

var canvas = GetWorkingCanvasOrSomething(manifestCanvasDictionary, firstCanvasPaintingInCanvas);

// Instantiate a new AnnotationPage - this is what we'll populate with PaintingAnnotations below
var annoPage = new AnnotationPage
{
    Id = pathGenerator.GenerateAnnotationPagesId(firstCanvasPaintingInCanvas),
    Items = []
};
```

`GetWorkingCanvasOrSomething()` contains all logic / logging etc for finding correct Canvas and assigning Id etc together. If I'm scanning the method I don't necessarily need to know the nitty gritty of what's happening in there, which can make the code easier to understand.

src/IIIFPresentation/Repository/Paths/PathRewriteParser.cs
```diff
@@ -80,7 +80,7 @@ public PathParts ParsePathWithRewrites(string? uri, int customer)
         if (Uri.TryCreate(uri, UriKind.Absolute, out var uriResult))
         {
             return ParsePathWithRewrites(uriResult.Host, uriResult.AbsolutePath, customer);
-        };
+        }
 
         return new PathParts(null, null, true);;
```
donaldgray: I know it's not changed where but noticed `;;`

PR #535, opened 2026-01-15
Fix errors not being thrown when assets differ across service

### LINKED ISSUE #516: Different ImageServices can reference different assets
The following test scenario failing as API is returning a 200:

```gherkin
Scenario: Asset id in different ImageService differ
Given the user has a manifest payload containing a single canvas
And the canvas has a body with multiple ImageServices that reference different DLCS image assets
Then the response status could should be 400
And the "detail" and "type" fields in the response body give appropriate information
```

It looks like the logic for working out collisions is only taking the _first_ ImageService into account. E.g. this works and returns 200:

```json
"body": {
    "id": "https://dlcs.example/iiif-img/2/1/foo/full/1024,683/0/default.jpg",
    "type": "Image",
    "width": 1024,
    "height": 683,
    "format": "image/jpeg",
    "service": [
        {
            "@context": "http://iiif.io/api/image/2/context.json",
            "@id": "https://dlcs.example/iiif-img/2/1/foo",
            "@type": "ImageService2",
            "profile": "http://iiif.io/api/image/2/level2.json",
            "width": 4886,
            "height": 3257
        },
        {
            "@context": "http://iiif.io/api/image/3/context.json",
            "id": "https://dlcs.example/iiif-img/v3/2/1/bar",
            "type": "ImageService3",
            "profile": "level2",
            "width": 4886,
            "height": 3257
        }
    ]
},
```

but this doesn't, it returns an error correct 400 or 500 depending on create or update (see #514)

```json
"body": {
    "id": "https://dlcs.example/iiif-img/2/1/foo/full/1024,683/0/default.jpg",
    "type": "Image",
    "width": 1024,
    "height": 683,
    "format": "image/jpeg",
    "service": [
        {
            "@context": "http://iiif.io/api/image/2/context.json",
            "@id": "https://dlcs.example/iiif-img/2/1/bar",
            "@type": "ImageService2",
            "profile": "http://iiif.io/api/image/2/level2.json",
            "width": 4886,
            "height": 3257
        },
        {
            "@context": "http://iiif.io/api/image/3/context.json",
            "id": "https://dlcs.example/iiif-img/v3/2/1/foo",
            "type": "ImageService3",
            "profile": "level2",
            "width": 4886,
            "height": 3257
        }
    ]
},
```

JackLewis-digirati:
Resolves #516 

This PR updates how the service is used to determine an asset id, so that mismatches between these values are caught and an exception thrown

### DIFF
--- a/src/IIIFPresentation/Services/Manifests/PaintableAssetIdentifier.cs
+++ b/src/IIIFPresentation/Services/Manifests/PaintableAssetIdentifier.cs
@@ -59,30 +59,33 @@ public partial class PaintableAssetIdentifier(IOptionsMonitor<DlcsSettings> dlcs
     private AssetId? ResolveFromImageServices(List<IService>? services, int customerId)
     {
         if (services is null) return null;
-
         AssetId? assetFromServices = null;
+        var regex = CanonicalImageRegex();
 
         foreach (var service in services)
         {
             AssetId? resultFromService = null;
-            
+
             switch (service)
             {
                 case ImageService3 service3:
-                    if (ParseId(CanonicalImageRegex(), service3.Id, customerId) is {} result3) resultFromService = result3;
+                    if (ParseId(regex, service3.Id, customerId) is {} result3) resultFromService = result3;
                     break;
                 case ImageService2 service2:
-                    if (ParseId(CanonicalImageRegex(), service2.Id, customerId) is {} result2) resultFromService = result2;
+                    if (ParseId(regex, service2.Id, customerId) is {} result2) resultFromService = result2;
                     break;
             }
 
-            if (assetFromServices != null && resultFromService != null && assetFromServices != resultFromService)
+            var hasMismatchBetweenServices = assetFromServices != null && resultFromService != null &&
+                                             assetFromServices != resultFromService;
+
+            if (hasMismatchBetweenServices)
             {
-                throw new PaintableAssetException(assetFromServices, resultFromService,
-                    $"Suspected asset from image service ({assetFromServices}) is not matched by a second value found in the service ({resultFromService})");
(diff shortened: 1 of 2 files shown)


src/IIIFPresentation/Services/Manifests/PaintableAssetIdentifier.cs
```diff
@@ -59,21 +60,32 @@ public partial class PaintableAssetIdentifier(IOptionsMonitor<DlcsSettings> dlcs
     {
         if (services is null) return null;
 
+        AssetId? assetFromServices = null;
```
donaldgray: Small: renaming vars might help? `*FromService` and `*FromService` was difficult to follow, initially I couldn't work out how the same var was declared outside then inside a loop... 

Given the method is `ResolveFromImageServices()` we know they're `FromServices` drop postfix?

src/IIIFPresentation/Services/Manifests/PaintableAssetIdentifier.cs
```diff
@@ -59,21 +60,32 @@ public partial class PaintableAssetIdentifier(IOptionsMonitor<DlcsSettings> dlcs
...
-                // default: break;
             }
+
+            if (assetFromServices != null && resultFromService != null && assetFromServices != resultFromService)
+            {
+                throw new PaintableAssetException(assetFromServices, resultFromService,
+                    $"Suspected asset from image service ({assetFromServices}) is not matched by a second value found in the service ({resultFromService})");
+            }
+            
+            if (assetFromServices == null) assetFromServices = resultFromService;
```
donaldgray: Small, use null-coalescing assignment `??=`?

PR #553, opened 2026-02-02
Make delete requests require an Etag

### LINKED ISSUE #531: ETag should be required for DELETE operations
Delete endpoints should require a valid ETag to be supplied, can currently delete resources by just supplying auth headers. Same method of fetching/validating as PUT uses.

This will avoid issues of accidentally deleting an outdated resource.

See Slack conversation https://digirati.slack.com/archives/C110R7N5Q/p1732282724832269 (this and then subequent reply/threads).

JackLewis-digirati:
Resolves #531

This PR makes it so that an Etag is required when calling `DELETE` for manifests and collections.  This shares the logic of `PUT` using the `EtagComparer`

Additionally, there's a small amount of refactoring around the naming of classes to make it clear which is for delete and upsert

### DIFF
--- a/src/IIIFPresentation/API/Features/Common/Helpers/HierarchyResourceDeleter.cs
+++ b/src/IIIFPresentation/API/Features/Common/Helpers/HierarchyResourceDeleter.cs
@@ -1,0 +1,83 @@
+﻿using API.Infrastructure.Helpers;
+using AWS.Helpers;
+using Core;
+using Microsoft.EntityFrameworkCore;
+using Models.API.General;
+using Models.Database.Collections;
+using Repository;
+using Repository.Helpers;
+
+namespace API.Features.Common.Helpers;
+
+public class HierarchyResourceDeleter(
+    PresentationContext dbContext,
+    IIIIFS3Service iiifS3,
+    ILogger<HierarchyResourceDeleter> logger)
+{
+    public async Task<ResultMessage<DeleteResult, DeleteResourceErrorType>> DeleteResource(string? etagFromRequest, 
+        IHierarchyResource? resource, CancellationToken cancellationToken)
+    {
+        if (resource is null) return DeleteErrorHelper.NotFound();
+
+        if (!EtagComparer.IsMatch(resource.Etag, etagFromRequest)) return DeleteErrorHelper.EtagNotMatching();
+        
+        switch (resource)
+        {
+            case Collection collection:
+            {
+                var error = await DeleteCollection(resource, collection, cancellationToken);
+                if (error != null) return error;
+                break;
+            }
+            case Models.Database.Collections.Manifest manifest:
+                await DeleteManifest(resource, manifest);
+                break;
+        }
+
+        try
+        {
+            await dbContext.SaveChangesAsync(cancellationToken);
+        }
+        catch (DbUpdateConcurrencyException ex)
+        {
+            var hierarchy = resource.Hierarchy.GetCanonical();
+            var resourceType = hierarchy.CollectionId != null ? "collection" : "manifest";
+            
+            logger.LogError(ex, "Error attempting to delete {ResourceType} {ResourceId} for customer {CustomerId}",
+                resourceType, hierarchy.CollectionId ?? hierarchy.ManifestId, hierarchy.CustomerId);
+            return DeleteErrorHelper.UnknownError(resourceType);
+        }
+
+        return new ResultMessage<DeleteResult, DeleteResourceErrorType>(DeleteResult.Deleted);
+    }
+    
+    private async Task<ResultMessage<DeleteResult, DeleteResourceErrorType>?> DeleteCollection(IHierarchyResource resource, 
+        Collection collection, CancellationToken cancellationToken)
+    {
+        var hasItems = await dbContext.Hierarchy.AnyAsync(
+            c => c.CustomerId == collection.CustomerId && c.Parent == collection.Id,
+            cancellationToken: cancellationToken);
+
+        if (hasItems)
+        {
+            return new ResultMessage<DeleteResult, DeleteResourceErrorType>(DeleteResult.BadRequest,
+                DeleteResourceErrorType.CollectionNotEmpty, "Cannot delete a collection with child items");
+        }
+        
+        dbContext.Collections.Remove(collection);
+
+        if (!collection.IsStorageCollection)
+        {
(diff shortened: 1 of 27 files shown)


src/IIIFPresentation/Models/API/General/DeleteCollectionType.cs
```diff
@@ -1,8 +1,9 @@
 ﻿namespace Models.API.General;
 
-public enum DeleteCollectionType
+public enum DeleteResourceType
```
donaldgray: Could this be better named? These look like different error states but `DeleteResourceType` doesn't seem to fit that, from the name I'd expect `Manifest`, `Collection`, `Hierarchy` etc.

Also - rename file to match type

src/IIIFPresentation/API/Infrastructure/PresentationController.cs
```diff
@@ -131,6 +131,7 @@ private IActionResult ConvertDeleteToHttp<TType>(DeleteResult result, string? me
             DeleteResult.Conflict => this.PresentationProblem(detail: message, type: GetErrorType(type), statusCode: (int)HttpStatusCode.Conflict, title: "Conflict"),
             DeleteResult.Error => this.PresentationProblem(detail: message, type: GetErrorType(type), statusCode: (int)HttpStatusCode.InternalServerError, title: "Internal Server Error"),
             DeleteResult.BadRequest => this.PresentationProblem(detail: message, type: GetErrorType(type), statusCode: (int)HttpStatusCode.BadRequest, title: "Bad Request"),
+            DeleteResult.PreConditionFailed => this.PresentationProblem(detail: message, type: GetErrorType(type), statusCode: (int)HttpStatusCode.PreconditionFailed, title: "Pre-condition failed"),
```
donaldgray: `Precondition` is one word. Also, I think `failed` should be capitilised to keep consistent with the other titles here.

src/IIIFPresentation/API/Features/Manifest/Requests/DeleteManifest.cs
```diff
@@ -8,20 +10,21 @@
 
 namespace API.Features.Manifest.Requests;
 
-public class DeleteManifest(int customerId, string manifestId)
-    : IRequest<ResultMessage<DeleteResult, DeleteManifestType>>
+public class DeleteManifest(int customerId, string manifestId, string etag)
```
donaldgray: `string? etag`? There's no guarantee it's been provided. 

Assuming there's a warning around this in `ManifestController` too? Also brings in line with `DeleteCollection`.

donaldgray (commented):
Is there enough commonality in `DeleteCollection` and `DeleteManifest` to warrant some shared code?

src/IIIFPresentation/Models/API/General/DeleteCollectionType.cs
JackLewis-digirati: This is now shared between `Manifest` and `Collection` and we've used `Resource` elsewhere to show this (`PathSettings`) - I've renamed it to `DeleteResourceErrorType` to make it a bit clearer what's going on though

src/IIIFPresentation/API/Features/Common/Helpers/HierarchyResourceDeleter.cs
```diff
@@ -0,0 +1,86 @@
...
+
+public class HierarchyResourceDeleter(
+    PresentationContext dbContext,
+    IIIIFS3Service iiifS3,
+    ILogger<HierarchyResourceDeleter> logger)
+{
+    public async Task<ResultMessage<DeleteResult, DeleteResourceErrorType>> DeleteResource<T>(string? etagFromRequest, int customerId, 
+        string resourceId, CancellationToken cancellationToken) where T : class, IHierarchyResource
+    {
+        var resource = await dbContext.Set<T>().Retrieve(customerId,  resourceId, cancellationToken: cancellationToken);
```
donaldgray: Should this pass `tracking: true` as the default is `false`?

src/IIIFPresentation/API/Features/Common/Helpers/HierarchyResourceDeleter.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        try
+        {
+            await dbContext.SaveChangesAsync(cancellationToken);
+        }
+        catch (DbUpdateConcurrencyException ex)
+        {
+            var hierarchy = resource.Hierarchy.GetCanonical();
+            var resourceType = hierarchy.CollectionId != null ? "collection" : "manifest";
```
donaldgray: You could use `T` to get the resourceType, rather than Hierarchy

src/IIIFPresentation/API/Features/Common/Helpers/HierarchyResourceDeleter.cs
```diff
@@ -0,0 +1,86 @@
...
+            
+            logger.LogError(ex, "Error attempting to delete {ResourceType} {ResourceId} for customer {CustomerId}",
+                resourceType, hierarchy.CollectionId ?? hierarchy.ManifestId, customerId);
```
donaldgray: `hierarchy.CollectionId ?? hierarchy.ManifestId` => `resource.Id` or `resourceId`?

src/IIIFPresentation/API/Infrastructure/PresentationController.cs
```diff
@@ -131,6 +131,7 @@ private IActionResult ConvertDeleteToHttp<TType>(DeleteResult result, string? me
             DeleteResult.Conflict => this.PresentationProblem(detail: message, type: GetErrorType(type), statusCode: (int)HttpStatusCode.Conflict, title: "Conflict"),
             DeleteResult.Error => this.PresentationProblem(detail: message, type: GetErrorType(type), statusCode: (int)HttpStatusCode.InternalServerError, title: "Internal Server Error"),
             DeleteResult.BadRequest => this.PresentationProblem(detail: message, type: GetErrorType(type), statusCode: (int)HttpStatusCode.BadRequest, title: "Bad Request"),
+            DeleteResult.PreconditionFailed => this.PresentationProblem(detail: message, type: GetErrorType(type), statusCode: (int)HttpStatusCode.PreconditionFailed, title: "Pre-condition Failed"),
```
donaldgray: Pre-condition is 1 word

donaldgray (approved):

PR #556, opened 2026-02-04
Stop persisting the space id between the asset when removing the space

### LINKED ISSUE #518: Assets with negative spaceId are accepted (low priority)
Assets with correct canonical format but a negative number as a spaceId `-1` are accepted and saved.

The AssetId is **not** stored so it's not recognised as an Asset but how should this be handled?

### LINKED ISSUE #544: `spaceId` can persist between payloads
Protagonist AssetIds are made up of `{customer}/{space}/{asset}`, see https://github.com/dlcs/protagonist/blob/develop/src/protagonist/DLCS.Core/Types/AssetId.cs. This allows the same `asset` value to be used for different customers or spaces. There is a behaviour that looks like presentation is only considering the `asset` part for some operations. This can lead to Manifests being "completed" without having their content-resources populated.

This has been tested and is working in https://github.com/dlcs/iiif-presentation/releases/tag/v0.7.4 so it looks like something in mixed-manifest work has introduced the issue.

## Steps to reproduce

1. Create a Manifest containing a single PaintedResource. The `Asset` property specifies an existing space:

```json
{
    "type": "Manifest",
    "parent": "https://{{presentationUrl}}/{{customerId}}/collections/root",
    "slug": "space_oddity4",
    "paintedResources": [
        {
            "canvasPainting": {
                "canvasId": "https://{{presentationUrl}}/{{customerId}}/canvases/hello"
            },
            "asset": {
               "id": "donald_1",
                "origin": "https://host.example/image.jpg",
                "mediaType": "image/jpeg",
                "space": 228
            }
        }
    ]
}
```

This returns a 202 or 201, depending on whether the provided asset exists already. This doesn't have a bearing on the issue.

2. Update the manifest, removing the `"space"` property from the asset.

```json
{
    "type": "Manifest",
    "parent": "https://{{presentationUrl}}/{{customerId}}/collections/root",
    "slug": "space_oddity4",
    "paintedResources": [
        {
            "canvasPainting": {
                "canvasId": "https://{{presentationUrl}}/{{customerId}}/canvases/hello"
            },
            "asset": {
               "id": "donald_1",
                "origin": "https://host.example/image.jpg",
                "mediaType": "image/jpeg",
                "space": 228
            }
        }
    ]
}
```

This returns a 202. However, the paintedResources>asset property shows that space 228 is still used despite not being specified in payload. This results in an `"Asset not found"` error.

```json
    "paintedResources": [
        {
            "type": "PaintedResource",
            "canvasPainting": {
                "canvasId": "https://presentation.dlcs/499/canvases/hello",
                "canvasOrder": 0
            },
            "asset": {
                "@id": "https://api.dlcs/customers/499/spaces/228/images/donald_1",
                "error": "Asset not found"
            }
        }
    ],
   "items": [
        {
            "id": "https://presentation.dlcs/499/canvases/hello",
            "type": "Canvas",
            "items": [
                {
                    "id": "https://presentation.dlcs/499/canvases/hello/annopages/0",
                    "type": "AnnotationPage",
                    "items": [
                        {
                            "id": "https://presentation.dlcs/499/canvases/hello/annotations/0",
                            "type": "Annotation",
                            "motivation": "painting",
                            "behavior": [
                                "processing"
                            ],
                            "target": "https://presentation.dlcs/499/canvases/hello"
                        }
                    ]
                }
            ]
        }
    ]
```

Subsequent GET requests return a 200, showing the the background-handler picks up the "batch completed" message and marks this Manifest as completed. However, the `"items"` property is incomplete and the `Asset` is still shown as `"Asset not found"`. 

Subsequent attempts to remove this asset can hit the error detailed in #536, where it then becomes impossible to remove the Asset from the Manifest.

------

After investigation I can see:
* IIIF-Presentation creates a space in Protagonist for use but then doesn't use it (e.g. Space 400). So something identifies the need for a space.
* This spaceId isn't saved against the Manifest, the value is `null`.
* GETting the BatchId that was created shows that `499/400/donald_1` was created in DLCS.

It seem that the resolution to this involves checking the full `{c}/{s}/{a}` format when comparing Assets. 

Also - can we add _something_ to the background handler to protect against this? This would leave the Manifest in a 202|Accepted state but that means that any previous working copy of the Manifest would be served until the issue is resolved.

JackLewis-digirati:
Resolves #544 
Resolves #518

This issue occurred when an asset was updated with the same `canvasId` as a previous asset, which caused the API to think this was an update to an existing asset, where previously it would reuse the old asset's space id if it was left blank, rather than using a space attached to the manifest.  This then compounded with potentially not knowing the manifest space until after the `dlcsManifestCoordinator` has run.

 This pull request makes it so that if a space id is not specified on an updated `canvasPainting`, it will use the manifest space instead.  Additionally, due to the workflow of `canvasPaintings` -> `dlcsManifestCoordinator` an extra step was added after the `dlcsManifestCoordinator` to update assets like this with a space id, if the space needs to be created by the `dlcsManifestCoordinator` 

Finally, another fix was implemented to make sure the manifest is updated with a space id if it receives one in an update operation.

> [!Note]
> Due to `-1` being used internally to signal that a `canvasPainting` requires updating later,  #518 has been implemented to disallow a client from specifying a negative space number to avoid issues.

### DIFF
--- a/src/IIIFPresentation/API/Features/Common/Helpers/UpsertErrorHelper.cs
+++ b/src/IIIFPresentation/API/Features/Common/Helpers/UpsertErrorHelper.cs
@@ -110,6 +110,11 @@ public static class UpsertErrorHelper
         => ModifyEntityResult<TCollection, ModifyCollectionType>.Failure(error,
             ModifyCollectionType.PaintableAssetError, WriteResult.BadRequest);
     
+    public static ModifyEntityResult<TCollection, ModifyCollectionType> AssetError<TCollection>(string error)
+        where TCollection : JsonLdBase
+        => ModifyEntityResult<TCollection, ModifyCollectionType>.Failure(error,
+            ModifyCollectionType.AssetError, WriteResult.BadRequest);
+    
     private static string CollectionType(bool isStorageCollection)
     {
         return isStorageCollection ? "Storage" : "IIIF";
--- a/src/IIIFPresentation/API/Features/Common/Helpers/UpsertErrorHelper.cs
+++ b/src/IIIFPresentation/API/Features/Common/Helpers/UpsertErrorHelper.cs
@@ -2,6 +2,7 @@
 using Core;
 using IIIF;
 using Models.API.General;
+using Services.Manifests.Exceptions;
 
 namespace API.Features.Common.Helpers;
 
@@ -110,9 +111,9 @@ public static class UpsertErrorHelper
         => ModifyEntityResult<TCollection, ModifyCollectionType>.Failure(error,
             ModifyCollectionType.PaintableAssetError, WriteResult.BadRequest);
     
-    public static ModifyEntityResult<TCollection, ModifyCollectionType> AssetError<TCollection>(string error)
+    public static ModifyEntityResult<TCollection, ModifyCollectionType> AssetError<TCollection>(AssetException exception)
         where TCollection : JsonLdBase
-        => ModifyEntityResult<TCollection, ModifyCollectionType>.Failure(error,
+        => ModifyEntityResult<TCollection, ModifyCollectionType>.Failure(exception.Message,
             ModifyCollectionType.AssetError, WriteResult.BadRequest);
     
     private static string CollectionType(bool isStorageCollection)
--- a/src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
@@ -72,8 +72,9 @@ public class CanvasPaintingResolver(
         
         existingManifest.CanvasPaintings ??= [];
         Debug.Assert(manifestParseResult.CanvasPaintings is not null, "manifestParseResult.CanvasPaintings is not null");
-        
-        var toInsert = UpdateCanvasPaintingRecords(existingManifest.CanvasPaintings, manifestParseResult.CanvasPaintings);
+
+        var toInsert = UpdateCanvasPaintingRecords(existingManifest.CanvasPaintings,
+            manifestParseResult.CanvasPaintings, existingManifest.SpaceId);
         
         var insertCanvasPaintingsError = await HandleInserts(toInsert, customerId, cancellationToken);
         if (insertCanvasPaintingsError != null) return CanvasPaintingRecords.Failure(insertCanvasPaintingsError);
@@ -81,14 +82,14 @@ public class CanvasPaintingResolver(
     }
     
     private List<InterimCanvasPainting> UpdateCanvasPaintingRecords(List<CanvasPainting> existingCanvasPaintings, 
-        List<InterimCanvasPainting> incomingCanvasPaintings)
+        List<InterimCanvasPainting> incomingCanvasPaintings, int? existingSpace)
     {
         var processedCanvasPaintingIds = new List<int>(incomingCanvasPaintings.Count);
         var toInsert = new List<InterimCanvasPainting>();
         
         foreach (var incoming in incomingCanvasPaintings)
         {
-            UpdateCanvasPainting(existingCanvasPaintings, incoming, processedCanvasPaintingIds, toInsert);
+            UpdateCanvasPainting(existingCanvasPaintings, incoming, processedCanvasPaintingIds, toInsert, existingSpace);
         }
         
         // Delete canvasPaintings from DB that are not in payload
@@ -103,7 +104,7 @@ public class CanvasPaintingResolver(
     }
     
     private void UpdateCanvasPainting(List<CanvasPainting> existingCanvasPaintings, InterimCanvasPainting incoming,
-        List<int> processedCanvasPaintingIds, List<InterimCanvasPainting> toInsert)
+        List<int> processedCanvasPaintingIds, List<InterimCanvasPainting> toInsert, int? existingSpace)
     {
         var candidates = GetCandidates(existingCanvasPaintings, incoming);
         var matching = TryFindMatching(incoming, processedCanvasPaintingIds, candidates);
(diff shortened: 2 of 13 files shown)


src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -72,23 +72,24 @@ public async Task<CanvasPaintingRecords> UpdateCanvasPaintings(int customerId, P
...
+            manifestParseResult.CanvasPaintings, existingManifest.SpaceId);
         
         var insertCanvasPaintingsError = await HandleInserts(toInsert, customerId, cancellationToken);
         if (insertCanvasPaintingsError != null) return CanvasPaintingRecords.Failure(insertCanvasPaintingsError);
         return CanvasPaintingRecords.Success(toInsert, manifestParseResult.AssetsIdentifiedInItems);
     }
     
     private List<InterimCanvasPainting> UpdateCanvasPaintingRecords(List<CanvasPainting> existingCanvasPaintings, 
-        List<InterimCanvasPainting> incomingCanvasPaintings)
+        List<InterimCanvasPainting> incomingCanvasPaintings, int? existingSpace)
```
donaldgray: Would this be better named as `manifestSpace`, rather than `existingSpace`? More informative as to where it's coming from

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -265,22 +266,32 @@ private ManifestParseResult ParseManifest(int customerId,
...
                 "Canvas painting exception encountered in {ManifestId}, could not retrieve an asset id",
                 presentationManifest.Id);
-            return new ManifestParseResult(UpsertErrorHelper.CouldNotRetrieveAssetId<PresentationManifest>(), null, null);
+            return new ManifestParseResult(UpsertErrorHelper.CouldNotRetrieveAssetId<PresentationManifest>(), null,
+                null);
+        }
+        catch (AssetException assetException)
+        {
+            logger.LogDebug(assetException,
+                "Error when parsing asset ");
```
donaldgray: Additional whitespace before closing `"` - worth adding any identifiers here?

src/IIIFPresentation/Services/Manifests/Helpers/SpaceHelper.cs
```diff
@@ -0,0 +1,42 @@
...
+public static class SpaceHelper
+{
+    /// <summary>
+    /// The space used to signal that the space needs to be populated following DLCS interactions
+    /// </summary>
+    /// <remarks>This is a negative number as negative numbers are disallowed for a space id</remarks>
+    public static int DefaultSpaceForLaterPopulation => -1;
+
+    /// <summary>
+    /// Updates asset JObjects that don't have a space with a spoecified space
```
donaldgray: Typo `spoecified`

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -253,6 +245,20 @@ private async Task<PresUpdateResult> UpdateInternal(UpsertManifestRequest reques
         }
     }
 
+    private void UpdateCanvasPaintingsAfterDlcsInteractions(DbManifest existingManifest,
```
donaldgray: This is called in 1 place but there are 2 calls to DlcsInteraction - should it be called twice? If not is it worth adding a comment as to what it differs?

src/IIIFPresentation/Services/Manifests/Helpers/InterimCanvasPaintingX.cs
```diff
@@ -35,12 +35,12 @@ public static string GetGroupingForIdAssignment(this InterimCanvasPainting canva
     public static bool ContainsId(this List<InterimCanvasPainting>? canvasPainting, string? id) =>
         canvasPainting?.Any(cp => cp.Id?.Equals(id, StringComparison.OrdinalIgnoreCase) ?? false) ?? false;
 
-    public static CanvasPainting ToCanvasPainting(this InterimCanvasPainting interimCanvasPainting, int? space)
+    public static CanvasPainting ToCanvasPainting(this InterimCanvasPainting interimCanvasPainting, int? existingSpace)
```
donaldgray: `manifestSpace`?

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -300,7 +306,7 @@ await dlcsManifestCoordinator.GetAssets(customerId, dbManifest, cancellationToke
     }
 
     private async Task<(PresUpdateResult?, DbManifest?)> UpdateDatabaseRecord(WriteManifestRequest request,
-        ParsedParentSlug parsedParentSlug, DbManifest existingManifest, CancellationToken cancellationToken)
+        ParsedParentSlug parsedParentSlug, DbManifest existingManifest, int? dlcsInteractionResultSpace, CancellationToken cancellationToken)
```
donaldgray: Better name than `dlcsInteractionResultSpace`?

src/IIIFPresentation/API/Features/Common/Helpers/UpsertErrorHelper.cs
```diff
@@ -110,6 +110,11 @@ public static ModifyEntityResult<TCollection, ModifyCollectionType> PaintableAss
         => ModifyEntityResult<TCollection, ModifyCollectionType>.Failure(error,
             ModifyCollectionType.PaintableAssetError, WriteResult.BadRequest);
     
+    public static ModifyEntityResult<TCollection, ModifyCollectionType> AssetError<TCollection>(string error)
```
donaldgray: If this will only be returned in the event of an `AssetException` being thrown - is it worth having it take that rather than a string?

donaldgray (approved):

Issue #453, opened 2025-08-22
Correct handling of composite-canvas/canvasOrder


donaldgray:
> [!WARNING]
> #451 should be implemented first

Handling of composite canvases within mixed manifests needs to be improved as it's not consistent. At time of writing it is only possible to add a composite canvas that contains paintedResources if it's the last canvas in the manifest. This is due to an issue with how `canvas_order` is used in validation.

Calling out the following from the [Mixed-manifest RFC](https://github.com/dlcs/iiif-presentation/blob/develop/docs/rfcs/0005-mixed-manifests.md) before looking at examples
> Start with strict validation. If anything is ambiguous, return a 400|BadRequest with an explanation of what couldn't be understood. We can loosen these rules moving forward if required.

## Example 1 - valid mixed manifest

The below example works
* `basic-canvas` - gains `canvasOrder = 0` from `"items"`. Has same order specified from `paintedResource`.
* `composite-canvas` - gains `canvasOrder = 1` from `"items"`. Has order 1 and 2 specified from `paintedResource`.

```json
{
    "type": "Manifest",
    "parent": "https://localhost:7230/7/collections/abc123",
    "slug": "example1",
    "items": [
         { 
            "type": "Canvas",
            "id": "basic-canvas"
        },
        {
            "type": "Canvas",
            "id": "composite-canvas"
        }
    ],
    "paintedResources": [
        {
            "canvasPainting": {
                "canvasId": "basic-canvas",
                "canvasOrder": 0
            },
            "asset": {
                "id": "standalone"
            }
        },
        {
            "canvasPainting": {
                "canvasId": "composite-canvas",
                "canvasOrder": 1
            },
            "asset": {
                "id": "one"
            }
        },
        {
            "canvasPainting": {
                "canvasId": "composite-canvas",
                "canvasOrder": 2,
                "target": "xywh=800,1000,900,900"
            },
            "asset": {
                "id": "two"
            }
        }
    ]
}
```

| canvas_id        | asset_id         | canvas_order |
| ---------------- | ---------------- | ------------ |
| basic-canvas     | 99/10/standalone | 0            |
| composite-canvas | 99/10/one        | 1            |
| composite-canvas | 99/10/two        | 2            |

## Example 2 - valid mixed manifest (currently fails)

The below currently fails validation:
[...]

PR #557, opened 2026-02-05
Reenable implicit and explicit ordering

### LINKED ISSUE #532: Can create Manifest with duplicate CanvasOrder but no ChoiceOrder
If I submit the following:

```json
{
    "type": "Manifest",
    "parent": "https://presentation.dlcs/499/collections/parent",
    "slug": "test",
    "paintedResources": [
        {
            "canvasPainting": { // explicit order=1
                "canvasId": "https://presentation.dlcs/499/canvases/two",
                "canvasOrder": 1
            },
            "asset": { ... }
        },
        {
            "canvasPainting": { // implicit order=1
                "canvasId": "https://presentation.dlcs/499/canvases/one"
            },
            "asset": { ... }
        }
    ]
}
```

the payload is accepted. Both of the initially returned `CanvasPainting` records are minimal, they only have a `CanvasOrder`, both are 1, e.g.

```json
"canvasPainting": {
  "canvasId": "https://presentation.dlcs/499/canvases/one",  
  "canvasOrder": 1
},
```

If this is an asynchronously processed Manifest, the Provisional canvases only show a single Canvas with a single painting-anno **with a choice** E.g.

```json
    "items": [
        {
            "id": "https://presentation.dlcs/499/canvases/two",
            "type": "Canvas",
            "items": [
                {
                    "id": "https://presentation.dlcs/499/canvases/two/annopages/1",
                    "type": "AnnotationPage",
                    "items": [
                        {
                            "id": "https://presentation.dlcs/499/canvases/two/annotations/1",
                            "type": "Annotation",
                            "motivation": "painting",
                            "behavior": [
                                "processing"
                            ],
                            "body": {
                                "type": "Choice"
                            },
                            "target": "https://presentation.dlcs/499/canvases/two"
                        }
                    ]
                }
            ]
        }
    ]
```

The final manifest reflects this and returns a choice, however the payload did not specify a choice - this should be rejected by validation as there are duplicate CanvasOrders without a choice.

Note that if I explicitly specify this same `CanvasOrder` this is rejected with

```json
{
    "title": "Bad request",
    "detail": "Canvases that share 'canvasOrder' must have same 'canvasId'",
    "instance": "https://presentation-api.dlcs.digirati.io",
    "type": "https://presentation-api.dlcs.digirati.io/errors/ModifyCollectionType/ValidationFailed",
    "status": 400
}
```

JackLewis-digirati:
Resolves #532 

This PR reenables mixing implicit and explicit ordering while adding in additional verification after the point that implicit ordering is worked out to detect collisions between explicit and implicit ordering.  

Initially, working out where and how to do this validation was complicated.  However, taking advantage of another validation rule around `null` `choiceOrder` allows collisions to be detected and reported to the user

> [!Note]
>Additional test to verify example 3 of #453 works


src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -201,6 +202,10 @@ private static string GetCanvasLoggingId(InterimCanvasPainting incoming)
         var canvasPaintingIds = await GenerateUniqueCanvasPaintingIds(requiredIds, customerId, cancellationToken);
         if (canvasPaintingIds == null) return UpsertErrorHelper.CannotGenerateUniqueId<PresentationManifest>();
 
+        var stuff = canvasPaintings
```
donaldgray: ?

src/IIIFPresentation/API/Features/Manifest/Validators/PresentationManifestDatabaseValidator.cs
```diff
@@ -0,0 +1,19 @@
+﻿using FluentValidation;
+using Models.Database;
+
+namespace API.Features.Manifest.Validators;
+
+/// <summary>
+/// Used for any validation that must be completed after implicit canvases have been given an order
+/// </summary>
+public class PresentationManifestDatabaseValidator : AbstractValidator<IEnumerable<CanvasPainting>>
```
donaldgray: Would this be better named after the type that's being validated? CanvasPaintingsValidator maybe?

Missing unit tests.

src/IIIFPresentation/API/Features/Manifest/Validators/PresentationManifestDatabaseValidator.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public PresentationManifestDatabaseValidator()
+    {
+        RuleFor(m => m)
+            .Must(lcp => !lcp.GroupBy(pr => pr.CanvasOrder)
+                .Where(g => g.Count() > 1)
+                .Any(grp => grp.Count() > 1 &&  grp.Any(pr => pr.ChoiceOrder == null)))
+            .WithMessage("Painted resources cannot have a null 'choiceOrder' within a detected choice construct. This can happen when implicit and explicit 'canvasOrder' values conflict");
```
donaldgray: Simplify error message - I'm not sure consumers will understand "detected choice construct".

src/IIIFPresentation/API.Tests/Integration/ModifyManifestAssetCreationTests.cs
```diff
@@ -1647,4 +1647,165 @@ public async Task CreateManifest_CreatesManifest_WhenMatchedManifestWithLabelFro
...
+        requestMessage.Headers.Add("Link", "<https://dlcs.io/vocab#Space>;rel=\"DCTERMS.requires\"");
+        
+        // Act
+        var response = await httpClient.AsCustomer().SendAsync(requestMessage);
+
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.BadRequest);
+        var errorResponse = await response.ReadAsPresentationResponseAsync<Error>();
+
+        errorResponse.Detail.Should().Be("Errors after processing canvas paintings. Painted resources cannot have a null 'choiceOrder' within a detected choice construct. This can happen when implicit and explicit 'canvasOrder' values conflict");
```
donaldgray: This doesn't seem the right error here? They are different `canvasId` values so shouldn't be a detected choice - I'd expect the error to be more around "Different canvases have different order"

src/IIIFPresentation/API.Tests/Integration/ModifyManifestAssetCreationTests.cs
```diff
@@ -1647,4 +1647,165 @@ public async Task CreateManifest_CreatesManifest_WhenMatchedManifestWithLabelFro
...
+                                         "origin": "some/origin"
+                                     }
+                                 }
+                             ] 
+                         }
+                         """;
+        
+        var requestMessage =
+            HttpRequestMessageBuilder.GetPrivateRequest(HttpMethod.Post, $"{Customer}/manifests", manifestWithSpace);
+        requestMessage.Headers.Add("Link", "<https://dlcs.io/vocab#Space>;rel=\"DCTERMS.requires\"");
```
donaldgray: (small) - are these erroneous for these tests?

donaldgray (commented):
It might be worth having a call with @tomcrane to see what makes sense for these rules, I think that assigning array order is okay but good to double check (e.g. an alternative would be if the first item has an order we start counting from there?)

src/IIIFPresentation/API/Features/Manifest/Validators/PresentationManifestDatabaseValidator.cs
JackLewis-digirati: spent some time thinking here.. I think using `Detected conflicting implicit and explicit 'canvasOrder' values` might be the most valid error message? At this point that's essentially what we're validating as we've previously validated an explicit choice doesn't have a null `choiceOrder`  - this would also fix the comment below as it would no longer be talking about choice constructs.  I shied away from using this initially, as what we're detecting is essentially an erroneous `choice` - but you're right that this has potential for not making sense to a user, even though the new error message doesn't quite match what we're actually checking for

JackLewis-digirati:
Closing this PR based on comment in #532 to park this ticket

PR #558, opened 2026-02-11
Allow audio, video and painting choices to correctly set labels and other fields

### LINKED ISSUE #545: Choice labels incorrect for duplicate audio files
Encountered an issue where a `Choice` body created via paintedResouces, results in a final manifest with duplicated 'label' set on each choice item. Not sure if this is restricted to audio items only, will need further investigation.

E.g. the sample payload should create a single Canvas with a choice. The label on each choice item would be `choice1` and `choice2` respectively.

```json
{
    "type": "Manifest",
    "parent": "https://{{presentationUrl}}/{{customerId}}/collections/root",
    "slug": "choice_label5",
    "paintedResources": [
        {
            "canvasPainting": {
                "canvasId": "https://{{presentationUrl}}/{{customerId}}/canvases/choice",
                "canvasOrder": 0,
                "choiceOrder": 1,
                "label": { "en": [ "choice1" ] },
                "canvasLabel": { "en": [ "canvaslabel1" ] }
            },
            "asset": {
               "id": "single1_x",
                "origin": "https://example.com/sample.wav",
                "mediaType": "audio/wav",
                "space": 228
            }
        },
        {
            "canvasPainting": {
                "canvasId": "https://{{presentationUrl}}/{{customerId}}/canvases/choice",
                "canvasOrder": 0,
                "choiceOrder": 2,
                "label": { "en": [ "choice2" ] }
            },
            "asset": {
                "id": "single1_x",
                "origin": "https://example.com/sample.wav",
                "mediaType": "audio/wav",
                "space": 228
            }
        }
    ]
}
```

Howver, the output is that both of the items in the choice have a label of `"choice2"`. The `canvasPainting` records returned have the correct label values set.

```json
{
    "id": "https://presentation.dlcs/499/canvases/choice",
    "type": "Canvas",
    "label": {
        "en": [
            "canvaslabel1"
        ]
    },
    "duration": 14.675,
    "items": [
        {
            "id": "https://presentation.dlcs/499/canvases/choice/annopages/0",
            "type": "AnnotationPage",
            "items": [
                {
                    "id": "https://presentation.dlcs/499/canvases/choice/annotations/0",
                    "type": "Annotation",
                    "motivation": "painting",
                    "body": {
                        "type": "Choice",
                        "items": [
                            {
                                "duration": 14.675,
                                "id": "https://dlcs.example/iiif-av/499/228/single1_x/full/max/default.mp3",
                                "type": "Sound",
                                "label": {
                                    "en": [
                                        "choice2"
                                    ]
                                },
                                "format": "audio/mp3"
                            },
                            {
                                "duration": 14.675,
                                "id": "https://dlcs.example/iiif-av/499/228/single1_x/full/max/default.mp3",
                                "type": "Sound",
                                "label": {
                                    "en": [
                                        "choice2"
                                    ]
                                },
                                "format": "audio/mp3"
                            }
                        ]
                    },
                    "target": "https://presentation.dlcs/499/canvases/choice"
                }
            ]
        }
    ]
}
```

I wonder if internally a mapping is made of `assetId`: `label`?

JackLewis-digirati:
Resolves #545

This was an issue where audio and video resources didn't correctly break the reference between the named query and the final manifest.  The ultimate issue occurs when the asset id is reused across multiple `canvasPaintings` and changes something like the `label`, which caused the final `canvasPainting` to set the label of all the previous `canvasPaintings` which referenced this asset.

This was handled for images, but it didn't extend to audio or video

Additionally, given audio and video can show as a `PaintingChoice` if multiple transcodes are specified, this has been handled as well

### DIFF
--- a/src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
+++ b/src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
@@ -1,4 +1,5 @@
-﻿using Core.Helpers;
+﻿using Core.Exceptions;
+using Core.Helpers;
 using Core.IIIF;
 using IIIF;
 using IIIF.Presentation;
@@ -268,7 +269,8 @@ public class ManifestMerger(SettingsBasedPathGenerator pathGenerator, IPathRewri
             HandleCanvasPainting(canvasPaintingChoice, canvas, namedQueryCanvas, body);
 
             // We might have 1 or more IPaintable elements (e.g. if NQ resource is already a choice, flatten it)
-            var paintables = GetPaintablesForChoice(body);
+
+            var paintables = new List<IPaintable>(GetPaintablesForChoice(body));
             if (canvasPaintingChoice.Label != null)
             {
                 logger.LogTrace("CanvasPainting {CanvasPaintingId} has label, setting on choice paintables",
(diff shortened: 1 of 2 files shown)


src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
```diff
@@ -268,7 +269,8 @@ private void ProcessChoice(Dictionary<AssetId, Canvas> canvasDictionary,
             HandleCanvasPainting(canvasPaintingChoice, canvas, namedQueryCanvas, body);
 
             // We might have 1 or more IPaintable elements (e.g. if NQ resource is already a choice, flatten it)
-            var paintables = GetPaintablesForChoice(body);
+
+            var paintables = new List<IPaintable>(GetPaintablesForChoice(body));
```
donaldgray: Why did this line need to change? `GetPaintablesForChoice(body)` already returns a `List<IPaintable>` ?

src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
```diff
@@ -426,15 +428,49 @@ private static PaintingAnnotation GetSafeFirstPaintingAnnotation(Canvas canvas)
...
+    /// <param name="paintable">The paintable to clone</param>
+    /// <param name="depth">The current depth of the function - used to track painting choice</param>
+    /// <returns>A safe paintable</returns>
+    /// <exception cref="PresentationException">Thrown when recursive depth exceeds the maximum allowed</exception>
+    private static IPaintable GetSafePaintable(IPaintable paintable, int depth)
+    {
+        if (depth == 3)
+        {
+            // Recursive function emergency exit - this shouldn't be hit
+            throw new PresentationException($"Recursive depth '{depth}' of painting choice deeper than can be handled");
```
donaldgray, line 447: How would this error manifest itself to a user?

donaldgray (approved):

src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
JackLewis-digirati: It would ultimately get caught by the `Unexpected error upserting manifest` - this error should honestly never be seen by the user as it means the named query has recursively provided `PaintingChoice`, which we have control over.  I added it as it felt like a bad idea to have a recursive call without an exit clause, but I did spend some time thinking it might not be even necessary

PR #559, opened 2026-02-12
Provide more information when an invalid asset id is provided

### LINKED ISSUE #546: Supplying a manifest with an invalid `paintedResource.asset.id` returns 500
Supplying an invalid asset to a painted resource should return a `dlcsException` error, but if the `id` is set to an invalid value - e.g:

- `"https://presentation-api.dlcs.digirati.io/52/canvases/random_resource_ooETA0"`, which I supplied by mistake
- `""`)

...Presentation will instead return a vague 500 error.

JackLewis-digirati:
Resolves #546

Modifies a generic error message to be more specific to the issue of an invalid asset id.  This occurred as it happens before the asset is sent to the DLCS and thus wasn't handled as well as it should have been

> [!Note]
> Takes advantage of `Exception.Data` as this allows for more efficient `rethrow` logic to add additional context to an error message

### DIFF
--- a/src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
@@ -1,6 +1,6 @@
 using System.Net;
 using API.Features.Common.Helpers;
-using API.Features.Storage.Helpers;
+using API.Features.Manifest.Exceptions;
 using API.Helpers;
 using API.Infrastructure.Helpers;
 using Core;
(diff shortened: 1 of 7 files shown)


src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -175,9 +175,29 @@ private async Task<DlcsInteractionResult> UpdateDlcsWithAssets(WriteManifestRequ
...
         
+        try
+        {
+            dlcsInteractionRequests = await knownAssetChecker.FindAssetsThatRequireAdditionalWork(
+                request.PresentationManifest, previousManifestAssetIds, spaceId, spaceCreated, request.CustomerId,
+                cancellationToken);
+        }
+        catch (ArgumentException argumentException)
+        {
+            logger.LogError(argumentException, "Error parsing  DLCS asset that requires more work for manifest {ManifestId}", manifestId);
```
donaldgray: I wonder if a custom exception type would be more informative than re-throwing `ArgumentException`? 

`ArgumentException` is a pretty vague _bad data passed to method_ error but this code assumes that the only time that will be thrown from `FindAssetsThatRequireAdditionalWork` is due to AssetId parsing. That might be the case _now_ but may change in the future. 

If you don't want to add a custom exception and use standard one `FormatException` feels like a better fit?

src/IIIFPresentation/API/Features/Manifest/ManagedAssetResultFinder.cs
```diff
@@ -38,7 +39,20 @@ public async Task<List<DlcsInteractionRequest>> FindAssetsThatRequireAdditionalW
         foreach (var paintedResource in presentationManifest.PaintedResources?.Where(pr => pr.Asset != null) ?? [])
         {
             var asset = paintedResource.Asset!;
-            var assetId = asset.GetAssetId(customerId);
+            AssetId assetId;
+
+            try
+            {
+                assetId = asset.GetAssetId(customerId);
```
donaldgray, line 46: Related to previous comment - depending on type thrown, should `.GetAssetId()` throw it, it's used in a few different places so would keep it consistent.

donaldgray (commented):
There are other asset values that would cause a failure in DLCS - I wouldn't expect we'd keep them in lock step so just checking that an invalid id that makes it's way to the DLCS would be handled appropriately? (ie error shown to requestor etc).

src/IIIFPresentation/API/Features/Manifest/ManagedAssetResultFinder.cs
JackLewis-digirati: There's a few separate errors that can come out of `DLCS.AssetId` - so added the custm error there, which will allow `GetAssetId` to use this custom exception

JackLewis-digirati:
Yeah, in the case it reaches the DLCS, as longs as we get back an error status code (which should be the case for an invalid asset id) there's a logic in the [DlcsHttpContent](https://github.com/dlcs/iiif-presentation/blob/develop/src/IIIFPresentation/DLCS/API/DlcsHttpContent.cs#L77) class to pull out error details and return it to the caller

donaldgray (approved):

PR #560, opened 2026-02-16
Validate customer id in paths

### LINKED ISSUE #527: API accepts incorrect CustomerId in body paths
It is possible to provide the "wrong" CustomerId value in paths. Noticed when specifying `parent` and `canvasId`. The specified customerId can be any numeric value, it looks like it's basically ignored. Verify whether there are any other properties that could be affected.

Affects both Manifest and Collection endpoints, PUT and POST. Hierarchical paths are untested.

E.g. if I POST the following Manifest to `https://presentation.example.com/999/manifests` (Customer 999)

```json
{
    "type": "Manifest",
    "parent": "https://presentation.example.com/7/collections/root",
    "slug": "single1_1",
    "paintedResources": [
        {
            "canvasPainting": {
                "canvasId": "`https://presentation.example.com/7/canvases/single1_1"
            },
        }
    ]
}
```

then the output will change the customer to be correct in the body, like so:

```json
{
    "type": "Manifest",
    "parent": "https://presentation.example.com/999/collections/root",
    "slug": "single1_1",
    "paintedResources": [
        {
            "canvasPainting": {
                "canvasId": "`https://presentation.example.com/999/canvases/single1_1"
            },
        }
    ]
}
```

CustomerId should be validated

> [!NOTE]
> Low priority

JackLewis-digirati:
Resolves #527 

Adds checks on differing customer id to the following values:

- canvas id
- parent
- public id

Now if there's a difference between the calling customer id vs an FQDN, an error will be returned to the user in painted resources and parent/public id

By contrast, if there's an incorrect customer id in `items`, a random canvas id is generated as this could be a customer dropping somebody else's manifest into the system

### DIFF
--- a/src/IIIFPresentation/API.Tests/Integration/ModifyManifestExternalItemsTests.cs
+++ b/src/IIIFPresentation/API.Tests/Integration/ModifyManifestExternalItemsTests.cs
@@ -1776,7 +1776,7 @@ public class ModifyManifestExternalItemsTests : IClassFixture<PresentationAppFac
     ""slug"": ""{slug}"",
     ""items"": [
         {{
-            ""id"": ""https://localhost:7230/2/canvases/{slug}"",
+            ""id"": ""https://localhost:7230/{Customer + 1}/canvases/{slug}"",
             ""type"": ""Canvas""
         }}
     ]
@@ -1796,6 +1796,6 @@ public class ModifyManifestExternalItemsTests : IClassFixture<PresentationAppFac
         var canvasPainting = responseManifest.PaintedResources.First().CanvasPainting;
         canvasPainting.CanvasId.Split('/').Last().Should().NotBe(slug, "recognised host with unrecognised path");
         canvasPainting.CanvasOriginalId.Should()
-            .Be($"https://localhost:7230/2/canvases/{slug}");
+            .Be($"https://localhost:7230/{Customer + 1}/canvases/{slug}");
     }
 }
--- a/src/IIIFPresentation/API.Tests/Integration/ModifyManifestExternalItemsTests.cs
+++ b/src/IIIFPresentation/API.Tests/Integration/ModifyManifestExternalItemsTests.cs
@@ -1761,4 +1761,41 @@ public class ModifyManifestExternalItemsTests : IClassFixture<PresentationAppFac
         canvasPainting.CanvasOriginalId.Should()
             .Be($"https://localhost:7230/{Customer}/canvases/additionalSlug/{slug}");
     }
+    
+    [Fact]
+    public async Task PutFlatId_ReturnsCanvasPaintingWithRandomId_WhenCanvasIdForAnotherCustomer()
+    {
+        // Arrange
+        var (slug, id) = TestIdentifiers.SlugResource();
+        var manifest = $@"
+{{
+    ""@context"": ""http://iiif.io/api/presentation/3/context.json"",
+    ""id"": ""https://iiif.example/manifest.json"",
+    ""type"": ""Manifest"",
+    ""parent"": ""http://localhost/{Customer}/collections/{RootCollection.Id}"",
+    ""slug"": ""{slug}"",
+    ""items"": [
+        {{
+            ""id"": ""https://localhost:7230/2/canvases/{slug}"",
+            ""type"": ""Canvas""
+        }}
+    ]
+}}";
+        
+        var requestMessage =
+            HttpRequestMessageBuilder.GetPrivateRequest(HttpMethod.Put, $"{Customer}/manifests/{id}", manifest);
+        
+        // Act
+        var response = await httpClient.AsCustomer().SendAsync(requestMessage);
+
+        // Assert
+        response.StatusCode.Should().Be(HttpStatusCode.Created);
+        
+        var responseManifest = await response.ReadAsPresentationResponseAsync<PresentationManifest>();
+
+        var canvasPainting = responseManifest.PaintedResources.First().CanvasPainting;
+        canvasPainting.CanvasId.Split('/').Last().Should().NotBe(slug, "recognised host with unrecognised path");
+        canvasPainting.CanvasOriginalId.Should()
+            .Be($"https://localhost:7230/2/canvases/{slug}");
+    }
 }
--- a/src/IIIFPresentation/API/Features/Common/Helpers/UpsertErrorHelper.cs
+++ b/src/IIIFPresentation/API/Features/Common/Helpers/UpsertErrorHelper.cs
@@ -116,9 +116,9 @@ public static class UpsertErrorHelper
         => ModifyEntityResult<TCollection, ModifyCollectionType>.Failure(exception.Message,
             ModifyCollectionType.AssetError, WriteResult.BadRequest);
     
-    public static ModifyEntityResult<TCollection, ModifyCollectionType> CustomerIdDoesNotMatchCaller<TCollection>(string field, int customerIdFromField, int callerCustomerId)
+    public static ModifyEntityResult<TCollection, ModifyCollectionType> CustomerIdDoesNotMatchCaller<TCollection>(string field)
         where TCollection : JsonLdBase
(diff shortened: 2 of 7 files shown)


src/IIIFPresentation/API.Tests/Integration/ModifyManifestExternalItemsTests.cs
```diff
@@ -1761,4 +1761,41 @@ public async Task PutFlatId_ReturnsCanvasPaintingWithRandomId_WhenCanvasIdNotRec
...
+        var manifest = $@"
+{{
+    ""@context"": ""http://iiif.io/api/presentation/3/context.json"",
+    ""id"": ""https://iiif.example/manifest.json"",
+    ""type"": ""Manifest"",
+    ""parent"": ""http://localhost/{Customer}/collections/{RootCollection.Id}"",
+    ""slug"": ""{slug}"",
+    ""items"": [
+        {{
+            ""id"": ""https://localhost:7230/2/canvases/{slug}"",
```
donaldgray: Small, is it worth using `{Customer + 1}` or something to make it obvious that `Customer` isn't 2?

src/IIIFPresentation/API/Features/Common/Helpers/UpsertErrorHelper.cs
```diff
@@ -116,6 +116,11 @@ public static ModifyEntityResult<TCollection, ModifyCollectionType> AssetError<T
         => ModifyEntityResult<TCollection, ModifyCollectionType>.Failure(exception.Message,
             ModifyCollectionType.AssetError, WriteResult.BadRequest);
     
+    public static ModifyEntityResult<TCollection, ModifyCollectionType> CustomerIdDoesNotMatchCaller<TCollection>(string field, int customerIdFromField, int callerCustomerId)
```
donaldgray: Is there any disclosure risk to saying this, rather than just treating it like any old id we couldn't parse? I can see logging int internally but not sure about returning to consumer?

There may also be instances (e.g. Caspian users) where they're not aware of their customerId but that's less of an issue.

src/IIIFPresentation/API/Helpers/ParentSlugParser.cs
```diff
@@ -116,28 +114,36 @@ private static bool IsRoot<T>(T presentation, string? id)
...
                 presentation.PublicId, presentation.Parent);
             return (UpsertErrorHelper.ParentMustMatchPublicId<T>(), null);
         }
 
-        return (null, parent);
+        return (null, parent.Parent);
     }
 
-    private async Task<Collection?> GetParentFromPublicId(IPresentation presentation, int customerId, CancellationToken cancellationToken)
+    private async Task<(ModifyEntityResult<T, ModifyCollectionType>? Errors, Collection? Parent)>
```
donaldgray: `(ModifyEntityResult<T, ModifyCollectionType>? Errors, Collection? Parent)`

Worth considering a custom class for this, or an alias, rather than a tuple, for readability? 

I wasn't sure as it's a private method but it's used for both this method and the one below.

src/IIIFPresentation/API/Helpers/ParentSlugParser.cs
```diff
@@ -148,31 +154,45 @@ private static bool IsRoot<T>(T presentation, string? id)
...
+        if (Uri.TryCreate(presentation.Parent, UriKind.Absolute, out var parentUri) is not true) return (null, null);
         var parentPath = pathRewriteParser.ParsePathWithRewrites(parentUri.Host, parentUri.AbsolutePath, customerId);
 
-        if (parentPath.Resource == null) return null;
+        if (parentPath.Resource == null) return (null, null);
+        
+        if (parentPath.Customer != customerId)
+        {
+            return (UpsertErrorHelper.CustomerIdDoesNotMatchCaller<T>("parent", parentPath.Customer!.Value, customerId), null);
+        }
```
donaldgray: Not dove much into it but from this PR alone `RetrieveParentFromPresentation` and `GetParentFromPublicId` have, at a surface level, some commonality - consider refactoring?

src/IIIFPresentation/Services/Manifests/ManifestPaintedResourceParser.cs
```diff
@@ -115,7 +116,14 @@ private InterimCanvasPainting CreatePartialCanvasPainting(int customerId, Painte
         }
         
         var parsedCanvasId = pathRewriteParser.ParsePathWithRewrites(canvasId.Host, canvasId.AbsolutePath, customerId);
+        
         CanvasHelper.CheckParsedCanvasIdForErrors(parsedCanvasId, canvasId.AbsolutePath, logger);
+        
+        if (customerId != parsedCanvasId.Customer)
```
donaldgray, line 122: As mentioned above - do we want to surface this?

src/IIIFPresentation/API/Features/Common/Helpers/UpsertErrorHelper.cs
JackLewis-digirati: I was initially thinking about conversations around trying to be as clear as possible around what happened.  However, you're right this could leak the customer id for a customer that is using rewritten paths that remove the customer id (technically, it would do this for paths that aren't rewritten, but in that scenario the customer id is clear in the URL).  Additionally, with Caspian, the customer id could be confusing if you're interacting with rewritten paths - though you would be interacting using `paintedResources` which requires a fairly deep understanding of the API

I think the best course of action is to keep the error message, but remove the customer id's from the detail

src/IIIFPresentation/API/Helpers/ParentSlugParser.cs
```diff
@@ -94,94 +92,109 @@ private static bool IsRoot<T>(T presentation, string? id)
...
-        var publicIdParentUri = PathParser.GetParentUriFromPublicId(presentation.PublicId);
+        Uri? parentUri;
+        if (fromPublicId)
+        {
+            if (presentation.PublicId == null) return new ParsedParent<T>();
+            parentUri = PathParser.GetParentUriFromPublicId(presentation.PublicId);
+        }
+        else
+        {
+            if (Uri.TryCreate(presentation.Parent, UriKind.Absolute, out parentUri) is not true) return new ParsedParent<T>();
```
donaldgray: Why use pattern matching here, rather than `!`? Is pattern matching against bools used elsewhere in sln?

src/IIIFPresentation/API/Helpers/ParentSlugParser.cs
```diff
@@ -94,94 +92,109 @@ private static bool IsRoot<T>(T presentation, string? id)
...
-            cancellationToken);
-        var parent = parentHierarchy?.Collection;
+        public ModifyEntityResult<T, ModifyCollectionType>? Errors { get; private init; }
         
-        if (parent != null) parent.Hierarchy.GetCanonical().FullPath = parentPath.Resource;
+        public static ParsedParent<T> Fail(ModifyEntityResult<T, ModifyCollectionType> errors) =>
+            new() { Errors = errors};
         
-        return parent;
+        public static ParsedParent<T> Success(Collection? parent) => new() { Parent = parent };
```
donaldgray, line 195: There are static `Success` and `Fail` methods in this class but `new ParsedParent<T>()` is returned in a few places, is it worth having a static `.Empty`, will avoid creating throw away objects.

src/IIIFPresentation/API/Helpers/ParentSlugParser.cs
```diff
@@ -94,94 +92,109 @@ private static bool IsRoot<T>(T presentation, string? id)
...
         TryGetParentFromPresentation<T>(
             IPresentation presentation,
             int customerId,
             CancellationToken cancellationToken) where T : JsonLdBase
     {
         // Try and get a parent from publicId 
-        var publicIdParent = await GetParentFromPublicId(presentation, customerId, cancellationToken);
+        var publicIdParent = await RetrieveParent<T>(presentation, customerId, true, cancellationToken);
+        
+        if (publicIdParent.Errors != null) return publicIdParent;
```
donaldgray: Combine with below `if` clause?

donaldgray (approved):

PR #561, opened 2026-02-18
Don't allow an unrecognised host to be parsed for a canvas id

### LINKED ISSUE #529: `PaintedResource.CanvasPainting.CanvasId` and `Canvas.Id` parsing logic differs
When providing Manifests to the API, the `paintedResource>canvasId` and `Canvas>id` values are parsed to determine if there are any matching canvases. The parsing rules for the 2 different properties seem to differ. This can lead to the confusion scenario where passing the same value for `item[*].id` and `paintedResource.canvasPainting.canvasId` can result in only one of the values being parsed.

For example, if we have the following path template values:

```ini
"PathSettings__PathRules__Overrides__example.com__Canvas" = "/example/{customerId}/canvases/{resourceId}"
```

I can submit the following payload:
```json
{
    "type": "Manifest",
    "parent": "https://presentation.example/499/collections/mixed_manifest_coll",
    "slug": "single1_2",
    "items": [
        {
            "type": "Canvas",
            "id": "https://example.com/example/499/canvases/foo"
        }
    ],
    "paintedResources": [
        {
            "canvasPainting": {
                "canvasId": "https://example.com/example/499/canvases/foo"
            },
            "asset": { ... }
        }
    ]
}
```

which results in the following error:

```json
{
    "title": "Operation failed",
    "detail": "The following canvas painting records conflict with the order from items - (id: foo, canvasOrder: 0)",
    "instance": "https://presentation-api.dlcs.digirati.io",
    "type": "https://presentation-api.dlcs.digirati.io/errors/ModifyCollectionType/ErrorMergingPaintedResourcesWithItems",
    "status": 400
}
```

> [!NOTE]
> The `"detail"` in the above error should be clarified, it's not immediately obvious what the issue is.

This error is because the `foo` is parsed as the value for the CanvasId from `"canvasId": "https://example.com/example/499/canvases/foo"` whereas the  `"id": "https://example.com/example/499/canvases/foo"` in the IIIF Canvas is _not_ parsed to result in `foo`. The parsing of Ids from `"items"` uses stricted logic than CanvasPainting - they should be aligned. 

I think the correct handling would be for `canvasId` to use as-strict handling as parsing `id` - however, does this conflict with https://github.com/dlcs/iiif-presentation/issues/357?

JackLewis-digirati:
Resolves #529

This PR completes the comment at the bottom of the above issue by making it so that an unrecognised host returns an error, as opposed to just trying to parse.

This will disallow both paths belonging to another customer as well as random paths that fit the URL structure


docs/rfcs/0005-mixed-manifests.md
```diff
@@ -154,6 +154,42 @@ The canonical paths: `/file/{customer}/{space}/{asset}`
 >
 > Referenced `/file/` types will be handled by adjuncts.
 
+## Canvas Id Parsing
+
+In order to match a manifest between `items` and `paintedResources`, the id of both **MUST** match.  The rules for this are as follows:
+
+```mermaid
+---
+title: Stuff
```
donaldgray: Placeholder title?

docs/rfcs/0005-mixed-manifests.md
```diff
@@ -154,6 +154,42 @@ The canonical paths: `/file/{customer}/{space}/{asset}`
...
+    checkErrors -- no --> no
+    checkErrors -- yes --> yes
+
+    no[No id]
+    yes[Id returned to user]
+```
+
+While there is no difference in the *logic* between parsing the id for `items` and `paintedResources`, where there **is** a difference is in the outcome of failures to  parse (i.e.: the `No id` outcome).  Namely, that in `items`, a failure will result in a generated id, whereas in `paintedResources` an error will be returned to the user.  This difference is because `items` can be copied and pasted from another user of the system, whereas `paintedResources` is purposeful as these can only be retrieved directly using an authenticated user.
+
+Additionally, in the checking for an error in the id, there is a check in `items` to make sure an id exists that matches an id from `paintedResources` as setting the canvas id like this nly matters with a matched canvas.
```
donaldgray: typo: `nly`.

What does this sentence mean? Does it mean that there's a change in behaviour as to when the Canvas is matched or not?

Would some examples help?

docs/rfcs/0005-mixed-manifests.md
```diff
@@ -154,6 +154,42 @@ The canonical paths: `/file/{customer}/{space}/{asset}`
...
+title: Stuff
+---
+
+flowchart
+    exist{Does the id exist?}
+    url{Is the id a URL?}
+    parseUrl[Parse the URL]
+    recognisedHost{Is the host recognised?}
+    parsePath[Parse the path, with rewrites if required]
+    checkErrors{Is there an error in the id?}
```
donaldgray: Explain what errors are checked? ie Canvas.Id can't be 'short', only viable when used in a match?

src/IIIFPresentation/Services/Manifests/Settings/PathSettings.cs
```diff
@@ -18,5 +18,11 @@ public class PathSettings
     public Uri GetCustomerSpecificPresentationUrl(int customerId)
         => CustomerPresentationApiUrl.GetValueOrDefault(customerId, PresentationApiUrl);
     
+    /// <summary>
+    /// Whether the host is either a customer specific host, or the standard presentation host URL
+    /// </summary>
+    public bool IsRecognisedHost(int customerId, string host) =>
```
donaldgray: Expand name for more clarify? `IsCustomerRecognisedHost` or `IsKnownHostForCustomer`?

donaldgray (commented):
LGTM - is RFC update more suited to an ADR rather than retrofitting an RFC? It's the recording of a decision, rather than a request for comments on an idea

docs/rfcs/0005-mixed-manifests.md
JackLewis-digirati: I've added some clarifying comments + examples, but it means that we generate a canvas id if we can't find a matching id in `paintedResources`

JackLewis-digirati:
I can move this to an ADR, but it felt like something that was missed from the RFC to me, over a decision being made

docs/ADR/0002-canvas-id-parsing.md
```diff
@@ -0,0 +1,88 @@
+# Canvas Id Parsing
+* Status: decided
+* Author: Jack Lewis
+* Decider: Jack Lewis
+* Date: 2026-02-24
+
+## Context and Problem Statement
+
+When a canvas id comes into the system there are several formats this id can be in (such as a flat id, a URL or a rewritten path).  On top of that, due to mixed manifests, we need  a way to match up canvases between `items` and `paintedResources` and this is done through the canvas id.  As such a set of rules needs to be enforced around how the canvas id is parsed and where
```
donaldgray: Sentence unfinished. _"As such a set of rules needs to be enforced around how the canvas id is parsed and where..."_.

docs/ADR/0002-canvas-id-parsing.md
```diff
@@ -0,0 +1,88 @@
...
+    url{Is the id a URL?}
+    parseUrl[Parse the URL]
+    recognisedHost{Is the host recognised?}
+    parsePath[Parse the path, with rewrites if required]
+    checkErrors{Is there an error in the id?}
+
+    exist -- no --> no
+    exist -- yes --> url
+    url -- no --> checkErrors
+    url --> parseUrl
```
donaldgray: Is this missing a "yes" label? `url -- yes --> parseUrl`

docs/ADR/0002-canvas-id-parsing.md
```diff
@@ -0,0 +1,88 @@
...
+
+    no[No id]
+    yes[Id returned to user]
+```
+
+There are a few separate errors that are checked for, these are as follows:
+- Prohibited characters - `'/', '=', ','`
+- If an id cannot be parsed from the URL
+
+While there is no difference in the *logic* between parsing the id for `items` and `paintedResources`, where there **is** a difference is in the outcome of failures to  parse (i.e.: the `No id` outcome).  Namely, that in `items`, a failure will result in a generated id, whereas in `paintedResources` an error will be returned to the user.  This difference is because `items` can be copied and pasted from another user of the system, whereas `paintedResources` is purposeful as these can only be retrieved directly using an authenticated user.
```
donaldgray: "outcome of failures to  parse" => "outcome of failure to parse" or "failing to parse"

docs/ADR/0002-canvas-id-parsing.md
```diff
@@ -0,0 +1,88 @@
...
+    "slug": "first-example",
+    "parent": "-container-",
+    "items": [
+        {
+            "id": "https://presentation.example/99/canvases/alpha",
+        }
+    ],
+    "paintedResources": [
+        {
+            "canvaSPainting": {
```
donaldgray: Small, canva**S**Painting

docs/ADR/0002-canvas-id-parsing.md
```diff
@@ -0,0 +1,88 @@
...
+    "parent": "-container-",
+    "items": [
+        {
+            "id": "https://presentation.example/99/canvases/alpha",
+        }
+    ]
+}
+```
+
+would result in a generated id
```
donaldgray: is this unfinished sentence? If not it's missing a `.`

docs/ADR/0002-canvas-id-parsing.md
```diff
@@ -0,0 +1,88 @@
...
+
+There are a few separate errors that are checked for, these are as follows:
+- Prohibited characters - `'/', '=', ','`
+- If an id cannot be parsed from the URL
+
+While there is no difference in the *logic* between parsing the id for `items` and `paintedResources`, where there **is** a difference is in the outcome of failures to  parse (i.e.: the `No id` outcome).  Namely, that in `items`, a failure will result in a generated id, whereas in `paintedResources` an error will be returned to the user.  This difference is because `items` can be copied and pasted from another user of the system, whereas `paintedResources` is purposeful as these can only be retrieved directly using an authenticated user.
+
+Additionally, in the checking for an error in the id, there is an additional check in `items` parsing to make sure an id matches an id from `paintedResources`, as setting the canvas id like this only matters with a matched canvas.  If there is no matching id, then the canvas id will be generated for items.  
+
+For example:
```
donaldgray, line 51: Call out that the items>id and canvasPainting>canvasId don't need to _exactly_ match? There can be `"alpha"` in one and `"https://presentation.example/99/canvases/alpha"` in the other and, as they both return `"alpha"` as parsed id, they're deemed a matched. Similarly you could use `https://customer.specific/canv/alpha` too (I think? - as long as it's a registered PresentationUrl for customer?)

docs/ADR/readme.md
```diff
@@ -1,4 +1,6 @@
 # IIIF-Presentation ADRs
 
 * [Database update failure](0001-database-update-failure.md)
-  * Details on how a failure that occurred when updating pg13 to pg14
\ No newline at end of file
+  * Details on how a failure that occurred when updating pg13 to pg14
```
donaldgray: I know it wasn't changed in this PR but this looks unfinished or needs reworded, not sure what the intention was.

* "Details on how a failure that occurred when updating pg13 to pg14 can be resolved"
* "Details on failure that occurred when updating pg13 to pg14"

donaldgray (approved):

PR #564, opened 2026-02-23
Remove pgcrypto from database

JackLewis-digirati:
Resolves DIM-28

When upgrading the database to pg14, the update failed.  There's a more detailed discussion in the notes folder explaining the issue, but in brief the `digest` function in the `pgcrypto` extension was being called for a computed column.  When this happens, the update failed as the extension either sin't loaded, or has no access to the `search_path` that `pgcrypto` is located on.  This then causes the update to fail as it can't find the `digest` function.

Ultimately, the fix for this was to replace the `digest` function call with the in-built `sha256` function.

The fix for this was to remove the `pgcrypto` calls from the column, and this also has the addon effect that the extension can be removed entirely.

> [!Note]
> This PR is just a migration which provides the below output

```sql
START TRANSACTION;


DO $EF$
BEGIN
    IF NOT EXISTS(SELECT 1 FROM "__EFMigrationsHistory" WHERE "migration_id" = '20260223093453_Remove pgcrypto') THEN
    CREATE EXTENSION IF NOT EXISTS citext;
    END IF;
END $EF$;

DO $EF$
BEGIN
    IF NOT EXISTS(SELECT 1 FROM "__EFMigrationsHistory" WHERE "migration_id" = '20260223093453_Remove pgcrypto') THEN
    CREATE OR REPLACE FUNCTION deterministic_uuid_sha256(ts timestamptz, txt text)
    RETURNS uuid
    LANGUAGE sql
    IMMUTABLE
    AS $$
      SELECT
        CASE
          WHEN ts IS NULL THEN '00000000-0000-0000-0000-000000000000'::uuid
          ELSE encode(substr(sha256(ts::text::bytea), 1, 16), 'hex')::uuid
        END
    $$;
    END IF;
END $EF$;

DO $EF$
BEGIN
    IF NOT EXISTS(SELECT 1 FROM "__EFMigrationsHistory" WHERE "migration_id" = '20260223093453_Remove pgcrypto') THEN
    INSERT INTO "__EFMigrationsHistory" (migration_id, product_version)
    VALUES ('20260223093453_Remove pgcrypto', '8.0.11');
    END IF;
END $EF$;
COMMIT;
```


src/IIIFPresentation/Repository/Migrations/20260223093453_Remove pgcrypto.cs
```diff
@@ -0,0 +1,43 @@
...
+                      WHEN ts IS NULL THEN '00000000-0000-0000-0000-000000000000'::uuid
+                      ELSE encode(substr(sha256(ts::text::bytea), 1, 16), 'hex')::uuid
+                    END
+                $$;
+                """);
+        }
+
+        /// <inheritdoc />
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
```
donaldgray, line 36: `down` statement missing to revert changes back

donaldgray (commented):
> When upgrading the database to pg14, the update failed. After investigation this was found to be an issue with pgcrypto being called for a computed column.

> The fix for this was to remove the pgcrypto calls from the column, and this also has the addon effect that the extension can be removed entirely.

Can you expand on this, enough that we'll remember what the issue was (`search_path` etc) should we need to look back and see _why_ removing pgcrypto was the best course of action? While writing this comment the thought occurred, following on from recent retro, should the decision to drop this be captured in an ADR - write up alternative options, pros/cons etc?

donaldgray (approved):

PR #565, opened 2026-02-25
Stop manifest converter from matching the incorrect canvas

### LINKED ISSUE #563: Manifests stuck in a `202 Accepted` state
While fixing tests from https://github.com/dlcs/iiif-presentation-tests/pull/195, it was discovered that the payload used for the `"Manifest with same image used more than once in both items and paintings"` test hangs at the `202 Accepted` stage even when both assets have finished ingesting. 

[Payload example](https://github.com/user-attachments/files/25443927/pav-payload.json)

JackLewis-digirati:
Resolves #563 

The issue was that the `ManifestConverter` was falling back to using the `canvasPainting` id instead of the canvas id if the path could not be parsed.  This caused an issue where a canvas was matched with a `canvasPainting` when it shouldn't have been

Additionally, there has been a validator added that disallows multiple canvases with the same id

### DIFF
--- a/src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
+++ b/src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
@@ -4,6 +4,7 @@ using DLCS;
 using IIIF.Presentation.V3;
 using IIIF.Presentation.V3.Annotation;
 using IIIF.Presentation.V3.Content;
+using IIIF.Presentation.V3.Traversal;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
 using Models.API.Manifest;
@@ -717,12 +718,10 @@ public class ManifestConverterTests
                                         {
                                             Id = $"{canvasId}/image/1"
                                         },
-
                                         new Image
                                         {
                                             Id = $"{canvasId}/image/2"
                                         }
-
                                     ]
                                 }
                             }
@@ -766,9 +765,9 @@ public class ManifestConverterTests
         var first = result.First();
         var last = result.Last();
         
-        first.Items.First().Items.First().As<PaintingAnnotation>().Body.Should().BeOfType<PaintingChoice>();
+        first.AllAnnotations().First().As<PaintingAnnotation>().Body.Should().BeOfType<PaintingChoice>();
         first.Id.Should().Be(canvasId);
-        last.Items.First().Items.First().As<PaintingAnnotation>().Body.Should().BeNull();
+        last.AllAnnotations().First().As<PaintingAnnotation>().Body.Should().BeNull();
         last.Id.Should().Be("http://base/123/canvases/assetId3");
     }
 }
--- a/src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
+++ b/src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
@@ -696,39 +696,41 @@ public class ManifestConverterTests
         // Arrange
         var customer = 123;
         
-        var canvases = new List<Canvas>
-        {
+        List<Canvas> canvases =
+        [
             new()
             {
                 Id = canvasId,
-                Items = 
+                Items =
                 [
                     new AnnotationPage
                     {
-                        Items = new List<IAnnotation>
-                        {
+                        Items =
+                        [
                             new PaintingAnnotation
                             {
                                 Body = new PaintingChoice
                                 {
-                                    Items = new List<IPaintable>
-                                    {
+                                    Items = (List<IPaintable>)
+                                    [
                                         new Image
                                         {
                                             Id = $"{canvasId}/image/1"
                                         },
+
                                         new Image
                                         {
                                             Id = $"{canvasId}/image/2"
-                                        },
-                                    }
+                                        }
+
+                                    ]
                                 }
                             }
-                        }
+                        ]
                     }
                 ]
             }
-        };
+        ];
 
         var canvasPaintings = new List<CanvasPainting>
         {
--- a/src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
+++ b/src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
@@ -687,4 +687,86 @@ public class ManifestConverterTests
         firstCanvas.Homepage.First().Id.Should().Be("https://foo.com/0/homepage/foo");
         result.First().Homepage.Should().NotBeNull();
     }
+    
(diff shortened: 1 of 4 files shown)


src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
```diff
@@ -687,4 +687,88 @@ public void GenerateProvisionalCanvases_RewritesProvisionalItemId_FromMatchedPai
...
+                                        new Image
+                                        {
+                                            Id = $"{canvasId}/image/1"
+                                        },
+
+                                        new Image
+                                        {
+                                            Id = $"{canvasId}/image/2"
+                                        }
+
```
donaldgray: tiny - extra whitespace here and 720.

src/IIIFPresentation/API.Tests/Converters/ManifestConverterTests.cs
```diff
@@ -687,4 +687,88 @@ public void GenerateProvisionalCanvases_RewritesProvisionalItemId_FromMatchedPai
...
+        // Act
+        var result = canvasPaintings.GenerateProvisionalCanvases(pathGenerator, canvases, pathRewriteParser);
+
+        // Assert
+        var first = result.First();
+        var last = result.Last();
+        
+        first.Items.First().Items.First().As<PaintingAnnotation>().Body.Should().BeOfType<PaintingChoice>();
+        first.Id.Should().Be(canvasId);
+        last.Items.First().Items.First().As<PaintingAnnotation>().Body.Should().BeNull();
```
donaldgray: On retro earlier we talked about removing warnings from old code but I'm Rider is showing a couple for possible null references in these new assertions - can they be resolved?

Would it be worth using some of the traversal helpers in iiif-net, or those in `ManifestX`, not sure if that would clarify or not but might make handling nulls easier https://github.com/digirati-co-uk/iiif-net/blob/main/src/IIIF/IIIF/Presentation/V3/Traversal/TraverseX.cs

src/IIIFPresentation/API/Converters/ManifestConverter.cs
```diff
@@ -96,7 +96,7 @@ Canvas GenerateProvisionalCanvas(IGrouping<string, CanvasPainting> groupedCanvas
             // Incoming grouping is by canvasId - so could contain 1:n paintingAnnos, some of which are choices
             var canvasPainting = groupedCanvasPaintings.First();
             var canvasIdFromCanvases = existingCanvases.Select(c =>
-                (canvasId: pathRewriteParser.ParsePathWithRewrites(c.Id, canvasPainting.CustomerId).Resource ?? canvasPainting.Id, canvas: c));
+                (canvasId: pathRewriteParser.ParsePathWithRewrites(c.Id, canvasPainting.CustomerId).Resource ?? c.Id, canvas: c));
```
donaldgray, line 99: Hard to see in isolation but is there any concern that this logic could deviate from similar elsewhere? 

This is for generating provisional canvases so affected Manifests that needed to be asynchronously processed. I'm assuming we weren't hitting this same issue for Manifests that are synchronously processed, so the logic for the sync-processing side of things must have had this logic already? If so - is there some common logic to be refactored? Or is the processing completely different? 

Was same payload working for synchronously processed?

donaldgray (approved):

src/IIIFPresentation/API/Converters/ManifestConverter.cs
JackLewis-digirati: this particular manifest was working synchronously and the processing is different as there's no need to build provisional items - what would happen is the provisional items would get saved to the staging DB with the wrong values (for remembering `items` etc.).   These then used the saved id to try and match up with items 

By contrast, if the manifest is built immediately, we essentially use the manifest (with the extra fields removed) from the request, so the wrong values were never generated from provisional items.

*however*, this did prompt me to test a scenario which doesn't work where somebody effectively duplicates items in the `items` property  (so you have duplicate id's) like this:

```json
{
  "type": "Manifest",
  "paintedResources": [
    {
      "asset": {
        "id": "identify_asset_image1",
        "mediaType": "image/jpeg",
        "origin": "origin.jpg"
      },
      "canvasPainting": {
        "canvasOrder": 4,
        "staticWidth": 683,
        "staticHeight": 1024,
      }
    }
  ],
  "items": [
    {
      "id": "https://root.io/canvas/1",
      "type": "Canvas",
      "items": [
        {
          "id": "https://root.io/canvas/1/page",
          "type": "AnnotationPage",
          "items": [
            {
              "id": "https://root.io/canvas/1/page/image-use-one",
              "type": "Annotation",
              "motivation": "painting",
              "body": {
                "id": "origin.jpg",
                "type": "Image",
                "width": 683,
                "height": 1024,
                "format": "image/jpeg"
              },
              "target": "https://digirati.io/canvas/1"
            },
            {
              "id": "https://root.io/canvas/1/page/image-use-two",
              "type": "Annotation",
              "motivation": "painting",
              "body": {
                "id": "origin.jpg",
                "type": "Image",
                "width": 683,
                "height": 1024,
                "format": "image/jpeg"
              },
              "target": "https://root.io/canvas/1"
            }
          ]
        }
      ]
    },
    {
      "id": "https://root.io/canvas/1",
      "type": "Canvas",
      "items": [
        {
          "id": "https://root.io/canvas/1/page",
          "type": "AnnotationPage",
          "items": [
            {
              "id": "https://root.io/canvas/1/page/image-use-one",
              "type": "Annotation",
              "motivation": "painting",
              "body": {
                "id": "origin.jpg",
                "type": "Image",
                "width": 683,
                "height": 1024,
                "format": "image/jpeg"
              },
              "target": "https://root.io/canvas/1"
            },
            {
              "id": "https://root.io/canvas/1/page/image-use-two",
              "type": "Annotation",
              "motivation": "painting",
              "body": {
                "id": "origin.jpg",
                "type": "Image",
                "width": 683,
                "height": 1024,
                "format": "image/jpeg"
              },
              "target": "https://root.io/canvas/1"
            }
          ]
        }
      ]
    }
  ],
  "slug": "stuck-manifest-issue",
  "parent": "root",
}
```

Synchronously, this results in a `500` error and asynchronously, it will be accepted, but the second `item` will be lost

I'll take a look and see if I can figure out the correct response in this scenario - it does feel like an invalid payload though - it's a mixed manifest, but you've got multiple `items` with the same id which confuses the converter and merger (albeit in different ways)

donaldgray: I think we should detect and reject duplicated canvas ids with a 400|BadRequest

PR #568, opened 2026-03-05
Initial adjuncts RFC

donaldgray:


docs/rfcs/0006-adjuncts.md
```diff
@@ -0,0 +1,348 @@
...
+                        "iiifLink": "annotations"
+                    }
+                ]
+            }
+        }
+    ]
+}
+```
+
+The above payload is an instruction to ingest asset `one`, containing 2 adjuncts: `mets.xml` and `annotation-page`. These are Canvas level adjuncts.
```
tomcrane: They are Canvas-level adjuncts in a named query or single-asset manifest, because we generate a single Canvas to represent all we know about the asset; the Canvas gives us a frame to hang all the things the DLCS generates for an asset from.

And for most scenarios, they can become Canvas-level adjuncts in a presentationManifest (i.e., generated by IIIF-CS Presentation), because that is almost always what you'd want to happen.

But strictly speaking, and in keeping with the `.adjuncts` property everywhere approach, they are really _asset-level_ adjuncts; we just have a policy of hoisting them up to the Canvas level in the produced IIIF Presentation JSON of a presentationManifest.

But this leaves a grey area. Eventually, with `.adjuncts` on any presentation resource, you could provide actual Canvas-level adjuncts from the Presentation end. And conversely, what if you actually wanted it to be an adjunct on the content resource in IIIF? A use case for this would be an annotation on a specific image in a choice (perhaps a feature than can only be seen in the X-Ray image). Or, you want an adjunct on the `Choice` resource? Some of these things we can address later.

I think "top down" Canvas-level adjuncts can just get merged with the "bottom-up" adjuncts coming from assets, where we are still in a one canvasPainting per Canvas scenario. And we can decide which ones go first in the list, we just have a standard behaviour and take it or leave it. It's the other scenarios I'm not sure about, where there is more than one CanvasPainting for a generated Canvas, more than one asset, each with one or more adjuncts; they are generating a Choice, or multiple paintings on the same canvas, or even both, AND they have adjuncts. This is going to be rare but it's not invalid, we just need to know what we're going to to about it.

The NQ that Presentation gets from protagonist obviously has one Canvas per asset / asset per canvas and it's presentation's job to group them according to CanvasPainting.

Perhaps the MVP approach is that during this process, any adjuncts accumulated for the multiple assets on a single Canvas ALL go to the created single Canvas. And we can worry about finessing adjuncts for `Choice` etc later. It's going to be niche but it's possible.

docs/rfcs/0006-adjuncts.md
```diff
@@ -0,0 +1,348 @@
...
+| Range      | Normalised Id of range        | `99/0/Range_bcdloifk0j1zgp1cvlga8v_iiif.io_api_cookbook_recipe_0024-book-4-toc_range_r0` | Range id was `https://iiif.io/api/cookbook/recipe/0024-book-4-toc/range/r0` on Manifest `bcdloifk0j1zgp1cvlga8v` |
+
+Manifest, Collection and Canvas are first-class citizens in IIIF-Presentation, we have internal identifiers for all of these types that are unique to a customer so we can use these as is. 
+Outside of these we will need to normalise the incoming `id`, applying scope with Manifest or Collection id prefix. 
+We shouldn't need to reverse engineer the stored asset id to find what the original id was, having a repeatable deterministic method for taking an `id` and generating the normalised form will be enough. When building a final Manifest we will have the incoming/staged Manifest and NamedQuery results; we need a predictable method of taking ids from the former and finding any representation in the latter.
+
+> [!CAUTION]
+> We will need to be aware of asset id length restrictions. Do we need to do something different to normalise? Possibly encode or hash the url?
```
tomcrane: Or something like `{first x characters, normalised}_{8-char-hash}` - you see this with URLs sometimes

docs/rfcs/0006-adjuncts.md
```diff
@@ -0,0 +1,348 @@
...
+## NamedQuery Results
+
+> Generated NamedQuery Manifests will output 'stub' Assets alongside normal assets, IIIF-Presentation will need to arrange them appropriately.
+
+A Manifest or Collection can have any of: Assets, Asset-level adjuncts and IIIF-level adjuncts. It can also have standard IIIF properties where adjuncts would be added (e.g. a Canvas can have "seeAlso", or a Manifest could have a "rendering").
+
+We have rules for handling Assets and how these are added to the final Manifest.
+
+For adjuncts we need a set of rules to determine how we generate final Manifests from NQs:
+* Where do we place asset adjuncts if those assets make up a choice or composite?
```
tomcrane: A more pithy summary!

docs/rfcs/0006-adjuncts.md
```diff
@@ -0,0 +1,348 @@
...
+* How do we handle the existence of adjunct target properties (e.g. if we have 2 `"seeAlso"` adjuncts but there are already 3 `"seeAlso"` resources)? Append? Is that safe? Could we end up with duplicates?
```
tomcrane: I think we have to append but check for duplicates - duplicates might be from round-tripping, we'll always have to recognise adjuncts that we've put in the Manifest (whether or not we are hosting them).

docs/rfcs/0006-adjuncts.md
```diff
@@ -0,0 +1,348 @@
...
+
+> Requests with show-extras headers will output `adjuncts` at relevant location.
+
+If IIIF resources have associated adjuncts then we need to render these if the show-extras headers are supplied.
+
+They will be output to the same location as they were supplied (ie they can't all be in a single "adjuncts" array, must be added at the appropriate level). The `CanvasPaintings`-like table can be used to assist in this construction.
+
+## Other Changes or Considerations
+
+* Collection changes. To date Collection handling is much simpler than Manifests - they are always JSON only. Now that they may contain adjuncts we will need to extend handling of Collections, e.g.:
```
tomcrane, line 231: May be worth just doing Manifests first and then Collections once that seems OK?

tomcrane (approved):
I wrote a rambling comment but I think you are well-aware of it.

docs/rfcs/0006-adjuncts.md
donaldgray: Added a note around this to existing comment.

docs/rfcs/0006-adjuncts.md
donaldgray: Updated comment.

docs/rfcs/0006-adjuncts.md
donaldgray: Agree - this will ultimately come down to planning but will add it to this list so that we capture.

docs/rfcs/0006-adjuncts.md
donaldgray: I've summarised this and added a link to discussion for further reading.

PR #574, opened 2026-03-24
Automatically filter queries by customer id

### LINKED ISSUE #573: Queries not filtering by customer
When the BackgroundHandler receives a 'batch completed' notification it loads details of that Batch via the BatchId. 

It then checks to see if there are other, incomplete batches with the same ManifestId, without filtering in CustomerId. 

Manifest Id is _not_ globally unique, it is only unique within an individual Customer - so there is a possibility that different Customers using the same ManifestId could result in the wrong Batch Id being picked up.

JackLewis-digirati:
Resolves #573

## What does this change?

This PR adds global query filter on customer id across the tables in the database so that it can be omitted from any data retrieval.  This will fix any occurrences of not filtering by customer id, which is required on effectively every query

> [!Note]
> Certain tests and test classes do require not filtering across customer id, this can be accomplished with adding `.IgnoreFilterQueries` to LINQ

Adding this filter to any future database tables just requires implementing the `ICustomerEntity` interface and through reflection the new table will receive the benefits of the customer id filter.  This should have minimal impact on performance as it happens once during startup

## Database Migration

While there is no migration for this work, there *is* changes to how querying for data works.  This has been called out on the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,3 +90,5 @@ The general hierarchy of dependencies from lowest to highest are as follows:
 This project takes advantage of a global query filter to filter all queries on the database by customer id automatically, so that it does not need to be specified on queries.  This is because the vast majority (if not all) require filtering by customer id and it's better to avoid issues with missing this filter out. 
 
 Some tests do require checking other customers than the one set for retrieval, and this can be done by using `.IgnoreQueryFilters` on the query to allow querying without the global query filter.
+
+Adding this query filter to tables in the database is controlled using the `ICustomerEntity` interface.  By implementing this interface, everything will automatically be pulled through that is required to get the global query filter working.
--- a/README.md
+++ b/README.md
@@ -91,4 +91,4 @@ This project takes advantage of a global query filter to filter all queries on t
 
 Some tests do require checking other customers than the one set for retrieval, and this can be done by using `.IgnoreQueryFilters` on the query to allow querying without the global query filter.
 
-Adding this query filter to tables in the database is controlled using the `ICustomerEntity` interface.  By implementing this interface, everything will automatically be pulled through that is required to get the global query filter working.
+Adding this query filter to entities in the context is controlled using the `ICustomerEntity` interface.  By implementing this interface, everything will automatically be pulled through that is required to get the global query filter working.
--- a/README.md
+++ b/README.md
@@ -84,3 +84,9 @@ The general hierarchy of dependencies from lowest to highest are as follows:
 | AWS, Repository, DLCS |
 | Services |
 | API, BackgroundHandler |
+
+#### Global query filters
+
+This project takes advantage of a global query filter to filter all queries on the database by customer id automatically, so that it does not need to be specified on queries.  This is because the vast majority (if not all) require filtering by customer id and it's better to avoid issues with missing this filter out. 
+
+Some tests do require checking other customers than the one set for retrieval, and this can be done by using `.IgnoreQueryFilters` on the query to allow querying without the global query filter.
--- a/src/IIIFPresentation/API/Features/Common/Helpers/HierarchyResourceDeleter.cs
+++ b/src/IIIFPresentation/API/Features/Common/Helpers/HierarchyResourceDeleter.cs
@@ -19,7 +19,7 @@ public class HierarchyResourceDeleter(
     public async Task<ResultMessage<DeleteResult, DeleteResourceErrorType>> DeleteResource<T>(string? etagFromRequest, int customerId, 
         string resourceId, CancellationToken cancellationToken) where T : class, IHierarchyResource
     {
-        var resource = await dbContext.Set<T>().Retrieve(customerId,  resourceId, true, cancellationToken);
+        var resource = await dbContext.Set<T>().Retrieve(resourceId, true, cancellationToken);
         
         if (resource is null) return DeleteErrorHelper.NotFound();
 
--- a/src/IIIFPresentation/API/Features/Common/Helpers/HierarchyResourceDeleter.cs
+++ b/src/IIIFPresentation/API/Features/Common/Helpers/HierarchyResourceDeleter.cs
@@ -2,12 +2,10 @@
 using API.Infrastructure.Helpers;
 using AWS.Helpers;
 using Core;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.EntityFrameworkCore;
 using Models.API.General;
 using Models.Database.Collections;
 using Repository;
-using Repository.Helpers;
 
 namespace API.Features.Common.Helpers;
 
@@ -58,7 +56,7 @@ public class HierarchyResourceDeleter(
     private async Task<ResultMessage<DeleteResult, DeleteResourceErrorType>?> DeleteCollection(IHierarchyResource resource, 
         Collection collection, CancellationToken cancellationToken)
     {
-        var hasItems = await dbContext.Hierarchy.AnyAsync( c => c.Parent == collection.Id,
+        var hasItems = await dbContext.Hierarchy.AnyAsync(c => c.Parent == collection.Id,
             cancellationToken: cancellationToken);
 
         if (hasItems)
--- a/src/IIIFPresentation/API/Features/Common/Helpers/HierarchyResourceDeleter.cs
+++ b/src/IIIFPresentation/API/Features/Common/Helpers/HierarchyResourceDeleter.cs
@@ -58,8 +58,7 @@ public class HierarchyResourceDeleter(
     private async Task<ResultMessage<DeleteResult, DeleteResourceErrorType>?> DeleteCollection(IHierarchyResource resource, 
         Collection collection, CancellationToken cancellationToken)
     {
-        var hasItems = await dbContext.Hierarchy.AnyAsync(
-            c => c.CustomerId == collection.CustomerId && c.Parent == collection.Id,
+        var hasItems = await dbContext.Hierarchy.AnyAsync( c => c.Parent == collection.Id,
             cancellationToken: cancellationToken);
 
         if (hasItems)
--- a/src/IIIFPresentation/API/Features/Manifest/ManagedAssetResultFinder.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/ManagedAssetResultFinder.cs
@@ -191,7 +191,7 @@ public class ManagedAssetResultFinder(
         trackedAssets.AddRange(dlcsAssetIds);
 
         var missingAssets = interimCanvasPaintings.Where(icp => !trackedAssets.Any(a =>
-            icp.SuspectedAssetId == a.Asset && icp.SuspectedSpace == a.Space)).ToList();
+            icp.SuspectedAssetId == a.Asset && icp.CustomerId == a.Customer && icp.SuspectedSpace == a.Space)).ToList();
         
         if (missingAssets.Count != 0)
         {
--- a/src/IIIFPresentation/API/Features/Manifest/ManagedAssetResultFinder.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/ManagedAssetResultFinder.cs
@@ -166,7 +166,7 @@ public class ManagedAssetResultFinder(
         var assetsToCheckInDlcs =
             assetsToAddToManifest.Where(asset =>
             {
-                if (dbContext.CanvasPaintings.Any(cp => cp.CustomerId == customerId && cp.AssetId == asset))
+                if (dbContext.CanvasPaintings.Any(cp => cp.AssetId == asset))
                 {
                     trackedAssets.Add(asset);
                     return false;
@@ -191,7 +191,7 @@ public class ManagedAssetResultFinder(
         trackedAssets.AddRange(dlcsAssetIds);
 
         var missingAssets = interimCanvasPaintings.Where(icp => !trackedAssets.Any(a =>
-            icp.SuspectedAssetId == a.Asset && icp.CustomerId == a.Customer && icp.SuspectedSpace == a.Space)).ToList();
+            icp.SuspectedAssetId == a.Asset && icp.SuspectedSpace == a.Space)).ToList();
         
         if (missingAssets.Count != 0)
         {
@@ -212,8 +212,7 @@ public class ManagedAssetResultFinder(
         {
             var assetIds = chunkedAssetsToCheck.Select(a => a.assetId);
             
-            inAnotherManifest.AddRange(dbContext.CanvasPaintings.Where(cp =>
-                assetIds.Contains(cp.AssetId) && cp.CustomerId == customerId));
+            inAnotherManifest.AddRange(dbContext.CanvasPaintings.Where(cp => assetIds.Contains(cp.AssetId)));
         }
 
         return inAnotherManifest;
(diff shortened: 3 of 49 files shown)


src/IIIFPresentation/API/Infrastructure/HttpContextCustomerIdProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        // Return default if no HTTP context (startup, migrations)
+        if (httpContextAccessor.HttpContext == null)
+        {
+            return DefaultCustomer;
+        }
+        
+        // Extract from route claims
+        httpContextAccessor.HttpContext.Request.RouteValues.TryGetValue(CustomerIdRouteValue, out var customerIdRouteVal);
```
JackLewis-digirati: I had a note to see if this could be shared with the `DelegatedAuthenticator`, but this is the only shared line,

src/IIIFPresentation/API/Features/Common/Helpers/HierarchyResourceDeleter.cs
```diff
@@ -58,8 +58,7 @@ public async Task<ResultMessage<DeleteResult, DeleteResourceErrorType>> DeleteRe
     private async Task<ResultMessage<DeleteResult, DeleteResourceErrorType>?> DeleteCollection(IHierarchyResource resource, 
         Collection collection, CancellationToken cancellationToken)
     {
-        var hasItems = await dbContext.Hierarchy.AnyAsync(
-            c => c.CustomerId == collection.CustomerId && c.Parent == collection.Id,
+        var hasItems = await dbContext.Hierarchy.AnyAsync( c => c.Parent == collection.Id,
```
donaldgray:
```suggestion
        var hasItems = await dbContext.Hierarchy.AnyAsync(c => c.Parent == collection.Id,
```

src/IIIFPresentation/API/Features/Manifest/ManagedAssetResultFinder.cs
```diff
@@ -191,7 +191,7 @@ public async Task<List<AssetId>> CheckAssetsFromItemsExist(
         trackedAssets.AddRange(dlcsAssetIds);
 
         var missingAssets = interimCanvasPaintings.Where(icp => !trackedAssets.Any(a =>
-            icp.SuspectedAssetId == a.Asset && icp.CustomerId == a.Customer && icp.SuspectedSpace == a.Space)).ToList();
+            icp.SuspectedAssetId == a.Asset && icp.SuspectedSpace == a.Space)).ToList();
```
donaldgray: Doesn't look like CustomerId filtered should be removed here? This isn't a `DbSet` so won't end up hitting the database.

src/IIIFPresentation/API/Infrastructure/HttpContextCustomerIdProvider.cs
```diff
@@ -0,0 +1,41 @@
+﻿using Repository.Helpers;
+
+namespace API.Infrastructure;
+
+public class HttpContextCustomerIdProvider(IHttpContextAccessor httpContextAccessor) : ICustomerIdProvider
```
donaldgray: Document what this is doing as it's important

donaldgray: As discussed
* Add comments
* Throw on Set()
* Consider using HttpContext.Items to store `customerId`, setting in auth handler and reading from here. Corresponding `Set()` and `Get()` helpers
* Throw at end,rather than return 0

src/IIIFPresentation/BackgroundHandler/CustomerCreation/CustomerCreatedMessageHandler.cs
```diff
@@ -49,9 +53,7 @@ private async Task EnsureRootCollection(CustomerCreatedMessage customerCreatedMe
         
         logger.LogInformation("Ensuring new customer {CustomerId} has root collection", customerId);
 
-        if (await dbContext.Collections.AnyAsync(
-                c => c.Id == KnownCollections.RootCollection && c.CustomerId == customerId,
-                cancellationToken))
+        if (await dbContext.Collections.AnyAsync( c => c.Id == KnownCollections.RootCollection, cancellationToken))
```
donaldgray:
```suggestion
        if (await dbContext.Collections.AnyAsync(c => c.Id == KnownCollections.RootCollection, cancellationToken))
```

src/IIIFPresentation/BackgroundHandler/Infrastructure/MessageBasedCustomerIdProvider.cs
```diff
@@ -0,0 +1,23 @@
...
+    private static readonly AsyncLocal<int?> CurrentCustomerId = new();
+    
+    public int GetCustomerId()
+    {
+        if (CurrentCustomerId.Value.HasValue)
+        {
+            return CurrentCustomerId.Value.Value;
+        }
+        
+        return 0;
```
donaldgray: Throw?

src/IIIFPresentation/BackgroundHandler/Infrastructure/MessageBasedCustomerIdProvider.cs
```diff
@@ -0,0 +1,23 @@
+using Repository.Helpers;
+
+namespace BackgroundHandler.Infrastructure;
+
+public class MessageBasedCustomerIdProvider : ICustomerIdProvider
```
donaldgray: Document. 

Tiny - is there a better name for this? It's not really anything to do with messages other than how it's used (ie it's not automatically set from messages, unlike the HTTPContext based on which is set from HTTP context).

src/IIIFPresentation/Repository/Helpers/ICustomerIdProvider.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Repository.Helpers;
+
+/// <summary>
+/// retrieves and sets a customer id used by a global query filter
```
donaldgray:
```suggestion
/// Retrieves and sets a customer id used by a global query filter
```

src/IIIFPresentation/Repository/PresentationContext.cs
```diff
@@ -139,4 +149,29 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+            if (typeof(ICustomerEntity).IsAssignableFrom(entityType.ClrType))
+            {
+                // grab the class from the entity
+                var parameter = Expression.Parameter(entityType.ClrType, "entity");
+                // grab the CustomerId property from the class
+                var customerIdProperty = Expression.Property(parameter, "CustomerId");
+                // get the method GetCustomerId from this class
+                var methodCall = Expression.Call(
+                    Expression.Constant(this),
+                    typeof(PresentationContext).GetMethod(nameof(GetCurrentCustomerId))!);
```
donaldgray: Small, not sure if it'll make any practical diff but the less reflection the better - finding `MethodInfo` could be moved outside the loop.

```cs
var currentCustomerIdMethod = typeof(PresentationContext).GetMethod(nameof(GetCurrentCustomerId))!;
foreach (var entityType in builder.Model.GetEntityTypes())
...
```

src/IIIFPresentation/Repository/PresentationContext.cs
```diff
@@ -139,4 +149,29 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+    {
+        // Automatically apply customer filter to all ICustomerEntity implementations
+        foreach (var entityType in builder.Model.GetEntityTypes())
+        {
+            if (typeof(ICustomerEntity).IsAssignableFrom(entityType.ClrType))
+            {
+                // grab the class from the entity
+                var parameter = Expression.Parameter(entityType.ClrType, "entity");
+                // grab the CustomerId property from the class
+                var customerIdProperty = Expression.Property(parameter, "CustomerId");
```
donaldgray:
```suggestion
                var customerIdProperty = Expression.Property(parameter, nameof(ICustomerEntity.CustomerId));
```

README.md
```diff
@@ -84,3 +84,9 @@ The general hierarchy of dependencies from lowest to highest are as follows:
 | AWS, Repository, DLCS |
 | Services |
 | API, BackgroundHandler |
+
+#### Global query filters
+
+This project takes advantage of a global query filter to filter all queries on the database by customer id automatically, so that it does not need to be specified on queries.  This is because the vast majority (if not all) require filtering by customer id and it's better to avoid issues with missing this filter out.
```
donaldgray, line 90: Details of which interface to use?

donaldgray (commented):
lgtm - couple of wee comments

src/IIIFPresentation/Repository/PresentationContext.cs
JackLewis-digirati: I did try this, but it produces an error with the message `Static method requires null instance, non-static method requires non-null instance.`

JackLewis-digirati: This was missing part of the invocation - adding this got it to work

src/IIIFPresentation/API/Infrastructure/HttpContextCustomerIdProvider.cs
JackLewis-digirati: Implemented this, but when looking at `HttpContext.Items` - the auth handler gets passed in the lower level `HttpRequest` as opposed to `HttpContext`, so I've left it as returning an integer

src/IIIFPresentation/API/Infrastructure/HttpContextCustomerIdProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        var customerId = httpContextAccessor.HttpContext.Request.GetCustomerId(logger);
+        if (customerId.HasValue) return  customerId.Value;
+
+        throw new PresentationException("Could not resolve customerId from the URL");
+    }
+
+    public void SetCustomerId(int customerId)
+    {
+        throw new PresentationException("Cannot set a customer id to the current customer");
```
donaldgray: Small - `"current customer"` isn't right

src/IIIFPresentation/API/Infrastructure/HttpContextCustomerIdProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public int GetCustomerId()
+    {
+        // Return default if no HTTP context (startup, migrations)
+        if (httpContextAccessor.HttpContext == null)
+        {
+            throw new PresentationException("HTTP context is null");
+        }
+
+        var customerId = httpContextAccessor.HttpContext.Request.GetCustomerId(logger);
```
donaldgray: This could be

```cs
httpContextAccessor.SafeHttpContext().Request.GetCustomerId(logger);
```

Would take care of the above `httpContextAccessor.HttpContext == null`, diff being an `ArgumentNullException` would be thrown.

src/IIIFPresentation/API/Infrastructure/HttpContextCustomerIdProvider.cs
```diff
@@ -0,0 +1,31 @@
+﻿using API.Infrastructure.Helpers;
+using Core.Exceptions;
+using Repository.Helpers;
+
+namespace API.Infrastructure;
+
+/// <summary>
+/// Retrieves a customer id from HTTPContext to use in
```
donaldgray: use in...

src/IIIFPresentation/BackgroundHandler.Tests/BatchCompletion/BatchCompletionMessageHandlerTests.cs
```diff
@@ -251,4 +252,63 @@ public async Task HandleMessage_UpdatesBatchedImages_WhenOldStyleBatchCompletion
...
+    [Fact]
+    public async Task HandleMessage_SavesResultingManifest_WhenAnotherCustomerIngestingSameManifestId()
+    {
+        // Arrange
+        var initialBatchId = TestIdentifiers.BatchId();
+        var (identifier, canvasPaintingId) = TestIdentifiers.IdCanvasPainting();
+        const int space = 2;
+        var assetId = new AssetId(CustomerId, space, identifier);
+        
+
```
donaldgray:
```suggestion
```

src/IIIFPresentation/Repository/Helpers/MigrationCustomerIdProvider.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Core.Exceptions;
+
+namespace Repository.Helpers;
+
+public class MigrationCustomerIdProvider : ICustomerIdProvider
+{
+    public int GetCustomerId() => 0; // Default for migrations
+    public void SetCustomerId(int customerId) =>
+        throw new PresentationException(); // do not set the customer id for migrations
```
donaldgray: `NotImplementedException` ?

src/IIIFPresentation/Test.Helpers/Integration/PresentationContextFixture.cs
```diff
@@ -335,17 +341,34 @@ public PresentationContext GetNewPresentationContext(
...
+        context.ChangeTracker.QueryTrackingBehavior = queryTrackingBehavior;
+        return context;
+    }
+    
+    /// <summary>
+    /// Get a new <see cref="PresentationContext"/> using connection string of this Postgres docker image.
+    /// Unlike the DbContext property this has full query tracking enabled to mimic 'real' context. This should be
+    /// used in instances where integration tests aren't using WebApplicationFactory to bootstrap environment
+    /// </summary>
+    public PresentationContext GetNewPresentationContext(ICustomerIdProvider customerIdProvider,
```
donaldgray, line 337: There's now 2 `GetNewPresentationContext` methods that are identical bar 1 taking `ICustomerIdProvider`, the other defaults it. Can there be a single method? Or one calls the other?

README.md
```diff
@@ -84,3 +84,11 @@ The general hierarchy of dependencies from lowest to highest are as follows:
...
 | Services |
 | API, BackgroundHandler |
+
+#### Global query filters
+
+This project takes advantage of a global query filter to filter all queries on the database by customer id automatically, so that it does not need to be specified on queries.  This is because the vast majority (if not all) require filtering by customer id and it's better to avoid issues with missing this filter out. 
+
+Some tests do require checking other customers than the one set for retrieval, and this can be done by using `.IgnoreQueryFilters` on the query to allow querying without the global query filter.
+
+Adding this query filter to tables in the database is controlled using the `ICustomerEntity` interface.  By implementing this interface, everything will automatically be pulled through that is required to get the global query filter working.
```
donaldgray: Small but this isn't actually doing anything to the tables in the db, it only affects the queries made _to_ the database.

```suggestion
Adding this query filter to entities in the context is controlled using the `ICustomerEntity` interface.  By implementing this interface, everything will automatically be pulled through that is required to get the global query filter working.
```

src/IIIFPresentation/API/Infrastructure/HttpContextCustomerIdProvider.cs
donaldgray: >the auth handler gets passed in the lower level `HttpRequest` as opposed to `HttpContext`, so I've left it as returning an integer

`HttpContext` is a property of `HttpRequest`, so you could access via `request.HttpContext.Items`.

donaldgray (approved):

src/IIIFPresentation/Repository/Helpers/MigrationCustomerIdProvider.cs
JackLewis-digirati: I use `PresentationException` elsewhere for this, so thought it was best to keep it consistent, but makes sense to be `NotImplemented`

Issue #581, opened 2026-04-09
Final Manifest includes Canvas-level adjuncts from NamedQuery (basic - single asset per Canvas)


donaldgray:
* Update background-handler completion logic. Subscribe to `AdjunctBatch` completion notifications (https://github.com/dlcs/protagonist/issues/1163) and include this in _"is manifest ready to process"_ logic
* Final Manifest includes Canvas-level adjuncts from NamedQuery

Basic - single asset per Canvas only.

christinemacdonald:
This is purely a development sub-task, no testable ACs on this specific ticket, so moving it to Done

PR #587, opened 2026-04-15
Allow adjuncts to be ingested into the DLCS alongside assets

### LINKED ISSUE #578: Accept canvas level adjuncts for Manifests
See RFC https://github.com/dlcs/iiif-presentation/blob/develop/docs/rfcs/0006-adjuncts.md#asset-canvas-level for more details on this approach.

> [!NOTE]
> This ticket is API only - the final Manifest won't reflect Adjuncts until #580 is done.

This is parent ticket is to handle ingesting Canvas level (ie via Assets) adjuncts. E.g. accepting the following payload would ingest 1 asset with 2 adjuncts. The final Manifest would have a single Canvas with 1x `"seeAlso"`, 1x `"annotations"` and content resources for an image.

```jsonc
{
    "type": "Manifest",
    "parent": "https://{{presentationUrl}}/{{customerId}}/parent",
    "slug": "child",
    "paintedResources": [
        {
            "asset": {
                "id": "one",
                "origin": "https://origin.example/image",
                "mediaType": "image/jpeg",
                "adjuncts": [ // new property
                    { 
                        "id": "mets.xml",
                        "externalId": "https://hosted.example/image/mets.xml",
                        "@type": "Dataset",
                        "mediaType": "text/xml",
                        "iiifLink": "seeAlso"
                    },
                    {
                        "id": "annotation-page",
                        "origin": "https://origin.example/annotation-page",
                        "@type": "AnnotationPage",
                        "mediaType": "application/json",
                        "label": { "en": [ "Line-level annotations" ] },
                        "iiifLink": "annotations"
                    }
                ]
            }
        }
    ]
}
```

All of the above tickets refer to both PUT and POST operations.

For now `"adjuncts"` reflects what the final adjuncts will be like. If missing then all existing would be deleted.

Suggested tickets:
* [x] Validate incoming adjunct (verify `id` only? That it doesn't contain invalid characters?)
* [x] Calculate diff of what's required to ingest Assets. If Asset is new then insert all Adjuncts. If Asset exists, workout which ones to ingest or delete. Make appropriate calls to IIIF-CS, recording created `AdjunctBatch` in addition to asset `Batch`. At this stage there is no "optimised updates" logic - we will always ingest adjuncts.
* [x] When returning payload populate the `"adjuncts"` property with full resources. Fetch with `?includes=adjuncts` (is that it?)
* [x] Subsequent GET requests with `x-iiif-show-extras` header return `"adjuncts"` property alongside Asset.

> [!NOTE]
> At this stage canvasOrder, choiceOrder, mixed-manifests etc are irrelevant. It's all just ingestion time.

JackLewis-digirati:
## What does this change?

This PR adds the ability to create adjuncts alongside assets in the same payload like this:

Resolves #578

```json
{
    "type": "Manifest",
    "parent": "https://{{presentationUrl}}/{{customerId}}/parent",
    "slug": "child",
    "paintedResources": [
        {
            "asset": {
                "id": "one",
                "origin": "https://origin.example/image",
                "mediaType": "image/jpeg",
                "adjuncts": [ // new property
                    { 
                        "id": "mets.xml",
                        "externalId": "https://hosted.example/image/mets.xml",
                        "@type": "Dataset",
                        "mediaType": "text/xml",
                        "iiifLink": "seeAlso"
                    },
                    {
                        "id": "annotation-page",
                        "origin": "https://origin.example/annotation-page",
                        "@type": "AnnotationPage",
                        "mediaType": "application/json",
                        "label": { "en": [ "Line-level annotations" ] },
                        "iiifLink": "annotations"
                    }
                ]
            }
        }
    ]
}
```

Additionally, details from the [RFC](https://github.com/dlcs/iiif-presentation/blob/develop/docs/rfcs/0006-adjuncts.md#manifest-payloads) have been implemented.  i.e.:

- `"adjuncts": null` or omitting the "adjuncts" property entirely won't result in any adjuncts being removed from an asset.
- `"adjuncts": []` will result in all adjuncts being removed from the relevant resource - whether they are added via IIIF Presentation or directly in Protagonist. The consumer is explicitly stating they want adjuncts to be empty.
- `"adjuncts": [{...}]` will result in the final Asset having these adjuncts only, any not included in the list will be removed.

Also, the adjuncts will be filled out with details when retrieving `canvasPaintings` as an admin

> [!Note]
> While details of the adjuncts will show up on the `canvasPaintings` records, the details will not show up on final manifest as this is waiting for the implementation of #581/#584.  Additionally, the processing of batches through the `backgroundHandler` is not supported, and instead batches that would be created are automatically finished instead of being handled by the background handler.

## Database Migration

- **What it does:** There are migrations which add a `deliverableType` column to the `batches`
 table and changing the primary key of `id` into a composite key of `id` and `deliverableType`
- **Breaking Change?** No

## Protagonist Dependency

The ability to ingest adjuncts has not yet had a specific release created in Protagonist, but this implementation will likely require v1.13

## Settings changes

`ProhibitedCharacters` has been moved from an internal class to one that can be set through settings, as well as being modified to take in different values for `canvasId` and `adjunctId`.  However, it still retains the same initial value, so should not need to be set in most (if not all) environments.  It can be set like the following:

```json
"ServicesSettings": {                                                                                                                                                                                                                                                                                                                                                                                                          
    "ProhibitedCanvasIdCharacters": ["/", "=", ","] ,
    "ProhibitedAdjunctIdCharacters" : ["/", "\\"]                                                                                                                                                                                                                                                                                                                                                                                  
  }
```


.gitignore
```diff
@@ -357,4 +357,6 @@ MigrationBackup/
 # Generated xml documentation files
 API/API.xml
 
-.env
\ No newline at end of file
+.env
+
+.claude/
```
JackLewis-digirati, line 362: not sure if this is wanted?

donaldgray: Makes sense - if we want to share anything we can remove.

src/IIIFPresentation/Models/DLCS/IIIFLinkType.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.ComponentModel;
+
+namespace Models.DLCS;
+
+/// <summary>
+/// The type of linking property to use for adjunct. Determines how this is output on Manifest.
+/// </summary>
+public enum IIIFLinkType
```
donaldgray: Unused, can go

src/IIIFPresentation/DLCS/Models/AdjunctAssetIdentifier.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace DLCS.Models;
+
+public class AdjunctAssetIdentifier
```
donaldgray, line 7: Quick description as to what this is? Assuming it's an AssetId and AdjunctIds - if so should `Id` be an `AssetId`?

src/IIIFPresentation/Core/Helpers/ProhibitedCharacters.cs
```diff
@@ -0,0 +1,9 @@
+namespace Core.Helpers;
+
+public static class ProhibitedCharacters
```
donaldgray: `ProhibitedAdjunctIdCharacters`?

Edit: noticed this is used for Canvas and Adjunct Ids but could they differ?

src/IIIFPresentation/Core/Helpers/ProhibitedCharacters.cs
```diff
@@ -0,0 +1,9 @@
...
+{
+    public static readonly char[] Characters = ['/', '=', ','];
```
donaldgray: I think it's worth pulling this from config as that's what Protagonist does. 

Protagonist API has default 
```
"RestrictedResourceIdCharacterString": "\\ /",
```

which could be overridden so it seems like we should follow that same pattern here. Also worth noting that the prohibited characters above differ from those currently in Protagonist.

src/IIIFPresentation/Core/Helpers/JObjectX.cs
```diff
@@ -27,7 +28,18 @@ public static T GetRequiredValue<T>(this JObject jObject, string property)
...
-
+        
         return value == null ? default : value.Value<T>();
     }
+    
+    /// <summary>
+    /// Try and get specified property value from jObject. This will throw a <see cref="FormatException"/> if the value
+    /// is not of the specified type
+    /// </summary>
+    public static IEnumerable<T>? TryGetCollectionValue<T>(this JObject jObject, string property)
```
donaldgray, line 39: Small - other methods in this class have unit tests so worth adding one for this?

src/IIIFPresentation/API/Helpers/BatchHelper.cs
```diff
@@ -11,14 +11,18 @@ public static class BatchHelper
     /// This method adds, but does not save batches to the batches table
     /// </summary>
     public static async Task AddBatchesToDatabase(this List<Batch> batches,
-        int customerId, string manifestId, PresentationContext dbContext, CancellationToken cancellationToken = default)
+        int customerId, string manifestId, PresentationContext dbContext, DeliverableType deliverableType = DeliverableType.Asset,
```
donaldgray: Small but it looks like all consumers provide `DeliverableType`, in that case can the default value be removed?
```suggestion
        int customerId, string manifestId, PresentationContext dbContext, DeliverableType deliverableType, 
```

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingRecords.cs
```diff
@@ -16,11 +16,12 @@ public static CanvasPaintingRecords Failure(ModifyEntityResult<PresentationManif
             Error = updateResult
         };
     
-    public static CanvasPaintingRecords Success(List<InterimCanvasPainting>? canvasPaintingsToAdd, List<InterimCanvasPainting>? itemsWithAssets) =>
+    public static CanvasPaintingRecords Success(List<InterimCanvasPainting>? canvasPaintingsToAdd, List<InterimCanvasPainting>? itemsWithAssets, List<AdjunctInteraction>? adjunctInteractions) =>
```
donaldgray: As discussed on call on Friday - should `CanvasPaintingRecords` be renamed to something more descriptive? The docs for class are 

> /// Records class containing details of canvas paintings that require further processing

but is that still true for collections of `InterimCanvasPainting` and `AdjunctInteraction` objects?

src/IIIFPresentation/Services/Manifests/Model/AdjunctInteraction.cs
```diff
@@ -0,0 +1,11 @@
+using Models.DLCS;
+using Newtonsoft.Json.Linq;
+
+namespace Services.Manifests.Model;
+
+public class AdjunctInteraction
```
donaldgray, line 10: Comment to explain what this is?

src/IIIFPresentation/Services/Manifests/Model/InterimCanvasPainting.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Models.Database;
+﻿using Newtonsoft.Json.Linq;
```
donaldgray: Unused using

src/IIIFPresentation/DLCS/API/DlcsApiClient.cs
```diff
@@ -77,16 +85,16 @@ public async Task<Space> CreateSpace(int customerId, string name, CancellationTo
         return space ?? throw new DlcsException("Failed to create space", HttpStatusCode.InternalServerError);
     }
     
-    public async Task<List<Batch>> IngestAssets<T>(int customerId, List<T> assets, CancellationToken cancellationToken = default)
+    public async Task<List<Batch>> IngestDeliverables<T>(int customerId, List<T> deliverables, bool adjunctQueue = false, CancellationToken cancellationToken = default)
```
donaldgray: Is `T` always `JObject`? If so, is it worth making it `JObject` explicitly? Or are there advantages to keeping it generic?

src/IIIFPresentation/Services/Manifests/Model/InterimCanvasPainting.cs
```diff
@@ -23,6 +24,11 @@ public class InterimCanvasPainting : CanvasPainting
     /// Whether the ordering for this record is set explicitly, or set by the API
     /// </summary>
     public bool ImplicitOrder { get; set; }
+    
+    /// <summary>
+    /// Potential adjuncts in the body of the request
+    /// </summary>
+    public AdjunctInteraction? SuspectedAdjuncts { get; set; }
```
donaldgray: Small - are these "suspect"? If we receive an `"adjuncts"` value then that is an indication that the asset should have these adjuncts.

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -164,15 +168,19 @@ private async Task<DlcsInteractionResult> HandlePaintedResourceDlcsInteractions(
...
-            
+
+            if (adjunctInteractions.Count > 0)
+            {
+                SpaceHelper.UpdateAdjuncts(adjunctInteractions, spaceId.Value);
+            }
         }
 
-        return await UpdateDlcsWithAssets(request, manifestId, previousManifestAssetIds, assets, assetsFromItems, spaceId,
+        return await UpdateDlcsWithAssets(request, manifestId, previousManifestAssetIds, assets, adjunctInteractions, assetsFromItems, spaceId,
```
donaldgray, line 178: Does `UpdateDlcsWithAssets` this need renaming to reflect that it also handles adjuncts? Or is this enough, assuming that all adjuncts are part of an asset (and if so, does that hang together with non-Asset adjuncts?).

donaldgray (commented):
Briefly tested locally and seems functionally correct - will need the follow up completion tickets done to run full e2e. 

I think adding some XML/code comments will help with comprehension when following code through.

src/IIIFPresentation/DLCS/Models/AdjunctAssetIdentifier.cs
JackLewis-digirati: this is basically the backing class needed by `deleteAdjuncts` - I'll add some comments to make this clearer

src/IIIFPresentation/Core/Helpers/ProhibitedCharacters.cs
JackLewis-digirati: right now, they're the same, so thought it was best to share - I'll add a comment to match though

src/IIIFPresentation/Core/Helpers/ProhibitedCharacters.cs
JackLewis-digirati: I've added the ability for this to be set in settings, though leaving this open for now to remind myself to update the description

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
JackLewis-digirati: I think that assuming adjuncts are a part of assets is ok? In order to add them into IIIF-CS, they currently need to be a  part of an asset

JackLewis-digirati:
## Adjunct ingestion flow

This is a summary of how adjuncts are processed end-to-end to help navigate the code:

**1. Parsing (`ManifestPaintedResourceParser` → `CanvasPaintingResolver`)**

Adjuncts are declared on entries in `paintedResources[*].asset.adjuncts`. `ManifestPaintedResourceParser` reads them and packages each one into an `AdjunctInteraction`, which binds the adjunct payloads to the `AssetId` of their parent asset. These are then surfaced through the private `ManifestParseResult` inside `CanvasPaintingResolver` and forwarded in the public `ParsedManifestResult` that is handed back to `ManifestWriteService`.

**2. Space hydration (`DlcsManifestCoordinator.HandlePaintedResourceDlcsInteractions`)**

Once a DLCS space is confirmed/created, `SpaceHelper.UpdateAdjuncts` is called to fill in the space number on any adjunct whose asset had a placeholder space (i.e. assets whose space was not yet known at parse time). This mirrors the same hydration step done for assets themselves.

**3. Populating existing adjunct IDs (`ManagedAssetResultFinder.PerformAssetInteractions`)**

When the DLCS is queried for existing assets (to determine what needs ingesting vs patching), `SetExistingAdjunctIds` is called per asset. This piggy-backs on the asset lookup to record which adjunct IDs are already registered in the DLCS for each asset, populating `AdjunctInteraction.ExistingAdjunctIds`. This avoids a separate round-trip for the common case.

**4. Adjunct reconciliation (`DlcsManifestCoordinator.HandleAdjunctInteractions`)**

After assets have been ingested, adjuncts are reconciled in three steps:

- `EnrichExistingAdjunctIds` — for any `AdjunctInteraction` where `ExistingAdjunctIds` is still null (i.e. its asset was not seen in step 3), a targeted DLCS fetch is made to retrieve the current adjunct IDs.
- `DeleteUnusedAdjuncts` — compares existing DLCS adjunct IDs against the IDs declared in the current manifest and bulk-deletes any that are no longer present.
- `IngestDeliverables` (with `DeliverableType.Adjunct`) — sends the adjunct payloads to the DLCS for ingestion, the same pipeline used for assets but flagged as adjuncts.

src/IIIFPresentation/Services/Manifests/Settings/ServicesSettings.cs
```diff
@@ -0,0 +1,12 @@
+namespace Services.Manifests.Settings;
+
+public class ServicesSettings
+{
+    public const string SettingsName = "ServicesSettings";
+
+    // NOTE: this is used by both adjuncts and canvas id's - if a change is made for 1, consider that it will affect both
```
donaldgray, line 7: I'm not sure it's best to share the same set of characters for the different identifier types. The default list here is what we were rejecting for CanvasIds but differs from what Protagonist allows for AdjunctIds. If we update it to align with Protagonist then we would start rejecting different characters for CanvasId.

Validity for Adjunct ids is ultimately governed by the Protagonist API. Previous suggestion to use settings to control this is to allow IIIF-P and Protagonist to be in sync, to avoid IIIF-P blocking things through that would be accepted, or allowing things through only to be knocked back later.

Canvas Id validity it completely controlled by IIIF-P, no other service is involved.

Final Q - if setting `ProhibitedCharacters` via appSettings, what would that look like? I'm asking as I recall you doing some work in Protagonist to make setting the prohibited characters a `string`, which is then mapped to a `char[]` but this is a `char[]` from the off. Thinking that if we did want to share this, it would likely be done via IaC so it would be useful if they could share a similar structure, whichever is easier. https://github.com/dlcs/protagonist/blob/develop/src/protagonist/API/Settings/ApiSettings.cs#L95-L108

src/IIIFPresentation/API/Features/Manifest/ParsedManifestResult.cs
```diff
@@ -6,21 +6,23 @@
 namespace API.Features.Manifest;
 
 /// <summary>
-/// Records class containing details of canvas paintings that require further processing
+/// Records class containing details of various tghings that require further processing, such as canvas paintings
```
donaldgray:
```suggestion
/// Records class containing details of various things that require further processing, such as canvas paintings
```

src/IIIFPresentation/DLCS/Models/AdjunctAssetIdentifier.cs
```diff
@@ -0,0 +1,18 @@
...
+/// <summary>
+/// Class used to allow for IIIF-CS calls that need just a list of adjunct id's attached to an asset
+/// <remarks>Used for things like deleting unneeded adjuncts</remarks>
+/// </summary>
+public class AdjunctAssetIdentifier
+{
+    /// <summary>
+    /// This is the string representation of an asset id i.e.: "{customer}/{space}/{id}"
+    /// </summary>
+    public required string Id { get; set; }
```
donaldgray, line 12: Given comment, should this be an `AssetId` type?

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -217,6 +228,72 @@ await UpdateAssetsWithManifestId(request, manifestId,
...
+            if (adjunctInteraction.ExistingAdjunctIds is not { Count: > 0 }) continue;
+
+            var currentIds = adjunctInteraction.Adjuncts
+                .Select(a => a[AdjunctProperties.Id]?.Value<string>())
+                .Where(id => id != null)
+                .ToHashSet();
+
+            var adjunctsToDelete = adjunctInteraction.ExistingAdjunctIds
+                .Where(id => !currentIds.Contains(id))
+                .ToList();
```
donaldgray, line 286: Small, style. This could be simplified to use `.Except()`, this builds a HashSet internally so perf should be the same.

```suggestion
            var currentIds = adjunctInteraction.Adjuncts
                .Select(a => a[AdjunctProperties.Id]?.Value<string>())
                .Where(id => id != null);

            var adjunctsToDelete = adjunctInteraction.ExistingAdjunctIds
                .Except(currentIds)
                .ToList();
```

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -217,6 +228,72 @@ await UpdateAssetsWithManifestId(request, manifestId,
...
+
+            if (adjunctsToDelete.Count > 0)
+                bulkDelete.Add(new AdjunctAssetIdentifier { Id = adjunctInteraction.AssetId.ToString(), Adjunct = adjunctsToDelete });
```
donaldgray, line 289: Seeing a warning locally for `adjunctsToDelete` potentially being null. Use `!adjunctsToDelete.IsNullOrEmpty()` rather than `.Count > 0`?

donaldgray (approved):

src/IIIFPresentation/Services/Manifests/Settings/ServicesSettings.cs
JackLewis-digirati: I updated the description with the example - 

```
"ServicesSettings": {                                                                                                                                                                                                                                                                                                                                                                                                          
    "ProhibitedCharacters": ["/", "=", ","]                                                                                                                                                                                                                                                                                                                                                                                    
  }
```

There's a slight issue though, with it being `char`, it would silently ignore anything with more than a single character, which would be difficult to debug - it's probably best to modify to `string`  so this can be avoided

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
JackLewis-digirati: This is warning about `List<string?>` rather than `List<string>?`, which is ultimately coming from the above query (`a[AdjunctProperties.Id]?.Value<string>()`), but given it then immediately strips out `where null` values, this is actually an erroneous error

src/IIIFPresentation/DLCS/Models/AdjunctAssetIdentifier.cs
JackLewis-digirati: I was trying to avoid referencing the `Models` project in `DLCS` if possible, but based on the project hierarchy, it makes sense to do that here for clarity

PR #590, opened 2026-04-24
Output adjuncts on Manifests

### LINKED ISSUE #581: Final Manifest includes Canvas-level adjuncts from NamedQuery (basic - single asset per Canvas)

### LINKED ISSUE #582: Final Manifest includes Canvas-level adjuncts from NamedQuery, multiple assets per Canvas (choices and composites). All adjuncts associated with assets on Canvas are hoisted to Canvas level.
Any Adjuncts that are part of an Asset that form a Composite or Choice are hoisted to the Canvas level.

E.g. if 3 assets formed a choice. Each of those 3 assets had 2 `seeAlso` adjuncts, there would be 6 Canvas level `seeAlso` elements

### LINKED ISSUE #584: Final Manifest includes Canvas-level adjuncts from NamedQuery (confirm mixed-manifests work)
Confirmation that mixed-manifests works when populating canvas-level adjuncts.

Any `seeAlso`, `annotation` and `rendering` properties present as valid IIIF are maintained in final manifest. If same property is already present the adjunct is appended to that collection. If an adjunct of the same property exists with a matching `id` the object that already exists on the provided manifests is used (ie the NQ is ignored).

### LINKED ISSUE #591: API returns correct status code for Manifest, taking Adjunct processing into account
Manifest endpoint can return 200/201/202 based on whether there is more work to do with the Manifest.

Manifest API processing logic will need updated to take into account whether there is further Adjunct processing to do.

* [ ] Return 200 if the Manifest can be completed immediately and it's been updated (ie there is no further work to do)
* [ ] Return 201 if the Manifest can be completed immediately and it's a creation.
* [ ] Return 202 if the Manifest cannot be completed immediately, either because of required Adjunct or Asset processing.

donaldgray:
## What does this change?

Overall updates ManifestMerger to handle adjuncts and ensure they are output in the correct location. A single `BatchCompletionMessageHandler` is used for both asset and adjunct batch completion logic. The logic doesn't differ so didn't feel like a worthwhile use of time to split into different handlers.

### Update `ManifestMerger`

Internally refactored some logic into `CanvasLookups` - this builds a couple of lookups. Keeps `ManifestMerger` leaner and should make it easier to track and find stub-assets for later tickets.

> [!WARNING]
> `ManifestMerger` is now a scoped service. Done so as the above `CanvasLookups` class is stored as an instance variable.
>
> This saves threading the class through multiple chains of calls at the expense of introducing state but the trade-off seemed worthwhile.

### Batches can be instantly-finished

In some instances the Batch that is returned from Protagonist API will be auto-finished (e.g. for `none` channel Assets or external Adjuncts). Update `BatchHelper` class to set Finished/Completed property on Batch entity if this is the case. 

`DlcsManifestCoordinator` now tracks all batches, both Adjunct and Asset, created during current request. It now uses this to drive the `canBeBuiltUpfront` decision, rather than relying on the types of requests sent.

Background handler updated to detect these and stop as completion message is still raised.

### Output Adjuncts on final Manifest

Annotations, SeeAlso and Rendering properties are aggregrated from NQ Canvases onto `StructureBase`. This only used for Canvases but `StructureBase` is the base class that `Manifest` etc will use in the future for exposing adjuncts.

### Update `QueueMessage`

Tweak sqs subscription request to get all custom attributes and map these correctly.

## Infrastructure Changes

> [!NOTE]
> This PR requires infrastructure changes.
>
> - **Required:** SNS subscription for Protagonists 'adjunctBatch completed' SNS topic to 'presentation batch completed' SQS. If missing adjunct-bearing Manifests won't complete as the "adjunctBatch completion" notification won't be picked up.

Resolves #581
Resolves #582
Resolves #584
Resolves #591


src/IIIFPresentation/BackgroundHandler.Tests/BatchCompletion/BatchCompletionMessageHandlerTests.cs
```diff
@@ -162,16 +211,17 @@ await dbContext.CanvasPaintings.AddTestCanvasPainting(manifest, id: canvasPainti
...
         var expectedCanvasId = $"https://localhost:5000/1/canvases/{canvasPaintingId}";
-        savedManifest.Items[0].Id.Should().Be(expectedCanvasId, "Canvas Id overwritten");
-        savedManifest.Items[0].Items[0].Id.Should().Be(
+        var firstCanvas = savedManifest.Items![0];
+        firstCanvas.Id.Should().Be(expectedCanvasId, "Canvas Id overwritten");
+        firstCanvas.Items![0].Id.Should().Be(
             $"https://localhost:5000/1/canvases/{canvasPaintingId}/annopages/1",
             "AnnotationPage Id overwritten");
-        var paintingAnnotation = savedManifest.Items[0].Items[0].Items[0].As<PaintingAnnotation>();
+        var paintingAnnotation = firstCanvas.Items![0].Items![0].As<PaintingAnnotation>();
```
JackLewis-digirati: `GetPaintingAnnotations`?

src/IIIFPresentation/Repository/Paths/PathRewriteParser.cs
```diff
@@ -149,6 +149,23 @@ private static (int customerId, string? resourceId) MatchValuesInTemplate(string
     }
 }
 
+/// <summary>
+/// Result of parsing a request path into CustomerId, resource identifier, and path type.
+/// Taking into account canonical paths and customer-specific rewrites.
+/// When Customer and Resource are both null, indicates the path could not be parsed.
```
JackLewis-digirati, line 155: probably not worth doing this here, but is it worth adding an explicit `bool error` to make it more obvious to the caller?

JackLewis-digirati (approved):

src/IIIFPresentation/Repository/Paths/PathRewriteParser.cs
donaldgray: Created https://github.com/dlcs/iiif-presentation/issues/597

src/IIIFPresentation/BackgroundHandler.Tests/BatchCompletion/BatchCompletionMessageHandlerTests.cs
donaldgray: Updated to `firstCanvas.GetFirstPaintingAnnotation()`

PR #598, opened 2026-05-06
Implement manifest level adjuncts into the API

### LINKED ISSUE #594: Manifest level adjuncts - API
> [!NOTE]
> This ticket is API only - the final Manifest won't reflect Adjuncts until further ticket is done.

See parent ticket for additional information.

> [!IMPORTANT]
> See #578 for Canvas-level adjuncts, which will share some basic logic.

This refers to both PUT and POST operations.

Suggested steps:
* [ ] Validate incoming adjunct id, as per #578 
* [ ] Calculate diff of what's required to ingest Adjuncts, as per #578. Following same rules of handling `null`, `[]` and `[{..}]` adjuncts.
* [ ] Ensure stub asset exists, see RFC for naming conventions.
* [ ] When returning payload, either in response to POST or further GET requests the _manifest_ level `"adjuncts"` property should be populated, without details of the stub asset.

JackLewis-digirati:
## What does this change?

Resolves #594

This PR allows manifest level adjuncts to be added to manifests using the new top-level `adjuncts` property like so:

```json
{
    "type": "Manifest",
    "slug": "test-manifest-level-adjuncts",
    "parent": "root",
    "adjuncts": [
                    { 
                        "id": "external",
                        "externalId": "https://external.example/image/mets.xml",
                        "@type": "Dataset",
                        "mediaType": "text/xml",
                        "iiifLink": "seeAlso"
                    },
    ],
}
```

This should then cause the corresponding `adjuncts` field to be filled out on the response.  Additionally, this PR takes advantage of the existing adjuncts code, so the calculation of the diff works the same way as the existing adjuncts code.

Finally, this PR modifies the `AdjunctAssetIdentifier` to use an `AssetId` instead of a `string`  so that it's clearer. 

## Protagonist Dependency

> [!NOTE]
> This PR requires a new version of Protagonist which hasn't been released yet, but is likely to be `1.13.0`

### DIFF
--- a/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
+++ b/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
@@ -513,7 +513,7 @@ public class GetManifestTests : IClassFixture<PresentationAppFactory<Program>>
             ContentBody = TestContent.ManifestJson,
         });
 
-        var manifestAdjunctId = ResourceAdjunctInteractions.GetResourceStubAssetId(PresentationContextFixture.CustomerId, id).Asset;
+        var manifestAdjunctId = ResourceAdjunctInteractions.GetResourceStubAssetId(new PresentationManifest(), PresentationContextFixture.CustomerId, id).Asset;
         A.CallTo(() => DLCSApiClient.GetCustomerImages(PresentationContextFixture.CustomerId, id, A<CancellationToken>._))
             .Returns(Task.FromResult<IList<JObject>>(
             [
--- a/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
+++ b/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
@@ -512,16 +512,16 @@ public class GetManifestTests : IClassFixture<PresentationAppFactory<Program>>
             ContentBody = TestContent.ManifestJson,
         });
 
-        var stubAssetName = CanvasPaintingResolver.ManifestStubAssetId(id);
+        var manifestAdjunctId = CanvasPaintingResolver.ManifestLevelAdjunctId(id);
         A.CallTo(() => DLCSApiClient.GetCustomerImages(PresentationContextFixture.CustomerId, id, A<CancellationToken>._))
             .Returns(Task.FromResult<IList<JObject>>(
             [
                 JObject.Parse($$"""
                     {
-                        "@id": "https://localhost/customers/1/spaces/0/images/{{stubAssetName}}",
-                        "id": "{{stubAssetName}}",
+                        "@id": "https://localhost/customers/1/spaces/0/images/{{manifestAdjunctId}}",
+                        "id": "{{manifestAdjunctId}}",
                         "space": 0,
-                        "adjuncts": [{ "id": "mets.xml", "asset": "1/0/{{stubAssetName}}" }]
+                        "adjuncts": [{ "id": "mets.xml", "asset": "1/0/{{manifestAdjunctId}}" }]
                     }
                     """)
             ]));
--- a/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
+++ b/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
@@ -50,7 +50,6 @@ public class GetManifestTests : IClassFixture<PresentationAppFactory<Program>>
     {
         s3 = storageFixture.LocalStackFixture.AWSS3ClientFactory();
         dbContext = storageFixture.DbFixture.DbContext;
-        Fake.ClearRecordedCalls(DLCSApiClient);
         httpClient = factory.ConfigureBasicIntegrationTestHttpClient(storageFixture.DbFixture,
             appFactory => appFactory.WithLocalStack(storageFixture.LocalStackFixture),
             services => services.AddSingleton(DLCSApiClient));
--- a/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
+++ b/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
@@ -512,7 +512,7 @@ public class GetManifestTests : IClassFixture<PresentationAppFactory<Program>>
             ContentBody = TestContent.ManifestJson,
         });
 
-        var manifestAdjunctId = CanvasPaintingResolver.ManifestLevelAdjunctId(id);
+        var manifestAdjunctId = CanvasPaintingResolver.ManifestStubAssetId(PresentationContextFixture.CustomerId, id).Asset;
         A.CallTo(() => DLCSApiClient.GetCustomerImages(PresentationContextFixture.CustomerId, id, A<CancellationToken>._))
             .Returns(Task.FromResult<IList<JObject>>(
             [
--- a/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
+++ b/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
@@ -3,6 +3,7 @@
 using System.Net;
 using Amazon.S3;
 using API.Converters;
+using API.Features.Common.Helpers;
 using API.Features.Manifest;
 using API.Tests.Integration.Infrastructure;
 using Core.Response;
@@ -512,7 +513,7 @@ public class GetManifestTests : IClassFixture<PresentationAppFactory<Program>>
             ContentBody = TestContent.ManifestJson,
         });
 
-        var manifestAdjunctId = CanvasPaintingResolver.ManifestStubAssetId(PresentationContextFixture.CustomerId, id).Asset;
+        var manifestAdjunctId = ResourceAdjunctInteractions.GetResourceStubAssetId(PresentationContextFixture.CustomerId, id).Asset;
         A.CallTo(() => DLCSApiClient.GetCustomerImages(PresentationContextFixture.CustomerId, id, A<CancellationToken>._))
             .Returns(Task.FromResult<IList<JObject>>(
             [
--- a/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
+++ b/src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
@@ -3,6 +3,7 @@
 using System.Net;
 using Amazon.S3;
 using API.Converters;
+using API.Features.Manifest;
 using API.Tests.Integration.Infrastructure;
 using Core.Response;
 using Core.Web;
@@ -37,7 +38,7 @@ public class GetManifestTests : IClassFixture<PresentationAppFactory<Program>>
     private readonly PresentationContext dbContext;
 
     private readonly IAmazonS3 amazonS3;
-    private readonly IDlcsApiClient dlcsApiClient;
+    private static readonly IDlcsApiClient DLCSApiClient = A.Fake<IDlcsApiClient>();
     private readonly IAmazonS3 s3;
     private readonly JObject sampleAsset;
     private const string PaintedResource = "foo-paintedResource";
@@ -48,13 +49,12 @@ public class GetManifestTests : IClassFixture<PresentationAppFactory<Program>>
     {
         s3 = storageFixture.LocalStackFixture.AWSS3ClientFactory();
         dbContext = storageFixture.DbFixture.DbContext;
-        dlcsApiClient = A.Fake<IDlcsApiClient>();
+        Fake.ClearRecordedCalls(DLCSApiClient);
         httpClient = factory.ConfigureBasicIntegrationTestHttpClient(storageFixture.DbFixture,
             appFactory => appFactory.WithLocalStack(storageFixture.LocalStackFixture),
-            services => services.AddSingleton(dlcsApiClient));
+            services => services.AddSingleton(DLCSApiClient));
 
         amazonS3 = storageFixture.LocalStackFixture.AWSS3ClientFactory();
-
         storageFixture.DbFixture.CleanUp();
 
         sampleAsset = JObject.Parse(
@@ -96,12 +96,12 @@ public class GetManifestTests : IClassFixture<PresentationAppFactory<Program>>
             """
         );
         
-        A.CallTo(() => dlcsApiClient.GetCustomerImages(PresentationContextFixture.CustomerId,
+        A.CallTo(() => DLCSApiClient.GetCustomerImages(PresentationContextFixture.CustomerId,
                 A<string>.That.Matches(l => l.EndsWith("_returnsPainted")),
                 A<CancellationToken>._))
             .ReturnsLazily(() => [sampleAsset]);
         
-        A.CallTo(() => dlcsApiClient.GetCustomerImages(PresentationContextFixture.CustomerId,
+        A.CallTo(() => DLCSApiClient.GetCustomerImages(PresentationContextFixture.CustomerId,
                 A<string>.That.Matches(l => l.EndsWith("_ingestingAssets")),
                 A<CancellationToken>._))
             .ReturnsLazily(() => [ingestingAsset, errorAsset]);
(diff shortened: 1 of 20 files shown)


src/IIIFPresentation/DLCS/Converters/AssetIdConverter.cs
```diff
@@ -0,0 +1,14 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Models.DLCS;
+
+namespace DLCS.Converters;
+
+public class AssetIdConverter : JsonConverter<AssetId>
```
donaldgray, line 7: Should this live alongside the `AssetId`? _"Here is the model and a converter for handling it"_

src/IIIFPresentation/API/Features/Manifest/Validators/PresentationManifestValidator.cs
```diff
@@ -17,6 +17,7 @@ public PresentationManifestValidator(IOptions<ServicesSettings> servicesOptions)
     {
         servicesSettings = servicesOptions.Value;
         When(m => !m.PaintedResources.IsNullOrEmpty(), PaintedResourcesValidation);
+        When(m => m.Adjuncts is { Count: > 0 }, ManifestAdjunctsValidation);
```
donaldgray: Consistency - this line and line above both have _"Use this validator if the list has items"_ logic but is done in 2 different ways,  `!list.IsNullOrEmpty()` and `list is { Count: > 0 }`, we should settle for 1 if possible (particularly on 2 consecutive lines).

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -306,6 +313,33 @@ private async Task<ManifestParseResult> ParseManifest(int customerId,
...
+            })
+            .ToList();
+
+        return new AdjunctInteraction { AssetId = manifestAdjunctId, Adjuncts = hydratedAdjuncts };
+    }
+
+    /// <summary>
+    /// The asset id for the manifest-level adjunct carrier asset in space 0, given the manifest's internal id.
+    /// </summary>
+    public static string ManifestLevelAdjunctId(string manifestId) => $"Manifest_{manifestId}";
```
donaldgray: Couple of points on this:

1. I think it'd make sense if we were to have a single place where we have the stub asset-id building logic. We'll need to call this method when finalising the Manifest too. Having in a central location would also give us a single place to manage any rules around restricting length or hasing (see "caution" in https://github.com/dlcs/iiif-presentation/blob/develop/docs/rfcs/0006-adjuncts.md#examples).
2. The method is `ManifestLevel*Adjunct*Id` but this is the id of the _Asset_ (XML comment reflects intent, method name is incorrect)
3. Given (2) should it return `AssetId`, always Space0? Usages can use `manifestId.Space`, might save a few uses of `0` being dotted around.

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -29,11 +30,11 @@ public class CanvasPaintingResolver(
     /// </summary>
     /// <returns>Tuple of either error OR newly created </returns>
     public async Task<ParsedManifestResult> GenerateCanvasPaintings(
-        int customerId, PresentationManifest presentationManifest, CancellationToken cancellationToken = default)
+        int customerId, PresentationManifest presentationManifest, string manifestId, CancellationToken cancellationToken = default)
```
donaldgray: Can't we get the manifestId from the `PresentationManifest` object? 

_"Here's an object - use it's Id if you need"_ is easier than _"Here's an object and here's the Id separately"_. I'm not sure if there's reasons it would be in a different format from just the plain flat id?

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -245,8 +246,8 @@ private static string GetCanvasLoggingId(InterimCanvasPainting incoming)
         }
     }
     
-    private async Task<ManifestParseResult> ParseManifest(int customerId, 
-        PresentationManifest presentationManifest, string? existingManifestId)
+    private async Task<ManifestParseResult> ParseManifest(int customerId,
+        PresentationManifest presentationManifest, string? existingManifestId, string? manifestId = null)
```
donaldgray: `string? manifestId = null` - when would we not know the `manifestId`? That seems like it shouldn't be possible - we either get given one or generate it ourselves, otherwise it wouldn't be possible to add adjuncts?

src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
```diff
@@ -22,11 +22,16 @@ public class PresentationManifest : IIIF.Presentation.V3.Manifest, IPresentation
     public List<PaintedResource>? PaintedResources { get; set; }
 
     [JsonProperty(Order = 13)] public string? Space { get; set; }
-    
+
+    /// <summary>
+    /// Manifest-level adjuncts to be associated with this manifest via a stub asset.
+    /// </summary>
+    [JsonProperty(Order = 14)] public List<JObject>? Adjuncts { get; set; }
```
donaldgray, line 29: Thinking ahead - I wonder if this would be better solved using the new `AdditionalProperties` stuff from iiif-net? I can see from usages of this in `CanvasPaintingResolver` that `Adjuncts` is accessed in `GetManifestAdjunctInteraction` method. That's good and solves it for `Manifest` - but the the plan is to allow adjuncts to be specified at various levels, not just `Manifest` level. 

Wondering if, rather than `GetManifestAdjunctInteraction` we could have something more generic that we're able to use for different types. Something that takes `ResourceBase` and recursively walked the tree, possibly filtering by known types, and building up `AdjunctInteractions` for different types? The stub asset-id generator could also take a `ResourceBase` and decide on the type of output accordingly.

src/IIIFPresentation/API/Features/Manifest/CanvasPaintingResolver.cs
```diff
@@ -306,6 +313,33 @@ private async Task<ManifestParseResult> ParseManifest(int customerId,
...
+    /// <summary>
+    /// Builds a manifest-level <see cref="AdjunctInteraction"/> for any adjuncts on the manifest request.
+    /// Note: stamps <see cref="AssetProperties.Asset"/> onto each adjunct JObject in-place.
+    /// </summary>
+    private static AdjunctInteraction? GetManifestAdjunctInteraction(int customerId,
+        PresentationManifest presentationManifest, string? manifestId)
+    {
+        if (presentationManifest.Adjuncts == null || manifestId == null) return null;
+
+        var manifestAdjunctId = new AssetId(customerId, 0, ManifestLevelAdjunctId(manifestId));
```
donaldgray: `0` magic number is used in a few places, we should give some meaning

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -130,7 +130,7 @@ private async Task<DlcsInteractionResult> HandlePaintedResourceDlcsInteractions(
     {
         var assets = GetAssetJObjectList(request.PresentationManifest.PaintedResources);
 
-        if (!request.CreateSpace && assets.Count <= 0 && previousManifestAssetIds.IsNullOrEmpty())
+        if (!request.CreateSpace && assets.Count <= 0 && previousManifestAssetIds.IsNullOrEmpty() && adjunctInteractions.Count == 0)
```
donaldgray: Small - there's 3 different approaches to checking if a `List<>` has items in this statement.

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -241,6 +241,12 @@ await UpdateAssetsWithManifestId(request, manifestId,
             // ExistingAdjunctIds may already be set for assets checked against DLCS in FindAssetsThatRequireAdditionalWork;
             // this call fills in any that were missed (e.g. assets that were already known and skipped the IIIF-CS check)
             await EnrichExistingAdjunctIds(request, adjunctInteractions, cancellationToken);
+
+            // Ensure stub assets (space 0) exist before ingesting their adjuncts
+            var manifestAdjunctAssetError = await EnsureManifestAdjunctAssetsExist(request.CustomerId, manifestId, adjunctInteractions,
```
donaldgray: As discussed on call - can there be 1 list of Assets and Adjuncts compiled before hitting this class? This class can then only concern itself with ensuring the Assets/Adjuncts exist, rather than needing to know how to split them up based on where they came from.

src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs
```diff
@@ -257,6 +263,37 @@ await UpdateAssetsWithManifestId(request, manifestId,
...
+        var manifestAdjunctAssets = missingManifestAdjunctAssets
+            .Select(a => new JObject
+            {
+                [AssetProperties.Id] = a.AssetId.Asset,
+                [AssetProperties.Space] = 0,
+                [AssetProperties.Manifests] = new JArray { manifestId }
+            })
+            .ToList();
+
+        var result = await IngestDeliverables(customerId, manifestId, manifestAdjunctAssets, DeliverableType.Asset,
```
donaldgray:
```suggestion
        var error = await IngestDeliverables(customerId, manifestId, manifestAdjunctAssets, DeliverableType.Asset,
```

src/IIIFPresentation/API/Features/Manifest/ManifestReadService.cs
```diff
@@ -79,6 +78,19 @@ public async Task<FetchEntityResult<PresentationManifest>> GetManifest(int custo
         manifest = manifest.SetGeneratedFields(dbManifest, pathGenerator, settingsBasedPathGenerator, assets,
             m => Enumerable.Single(m.Hierarchy!, h => h.Canonical));
 
+        // If the DLCS lookup failed, assets will be null (error already logged in DlcsManifestCoordinator).
+        // Return the manifest without manifest-level adjuncts rather than failing the GET.
+        if (assets != null)
+        {
+            // set manifest level adjuncts
```
donaldgray: Small - would this be better as a method named `SetManifestLevelAdjuncts()`, rather than a comment and a few lines?

src/IIIFPresentation/API/Features/Manifest/ManifestReadService.cs
```diff
@@ -79,6 +78,19 @@ public async Task<FetchEntityResult<PresentationManifest>> GetManifest(int custo
...
+            var manifestAdjunctId = CanvasPaintingResolver.ManifestLevelAdjunctId(dbManifest.Id);
+            var manifestAdjunctAsset = assets.Values.FirstOrDefault(a => a[AssetProperties.Id]?.Value<string>() == manifestAdjunctId);
+            if (manifestAdjunctAsset?[AssetProperties.Adjuncts] is JArray adjunctsArray)
+                manifest.Adjuncts = adjunctsArray.OfType<JObject>()
```
donaldgray: We discussed braces on slack - I think this is case where they'd be useful as it spans multiple lines.

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -265,13 +265,13 @@ private async Task<ResolvedManifestData> ResolveCanvasPaintingsAndParentSlug(Wri
     }
 
     private async Task<(PresUpdateResult? error, ParsedManifestResult? records)> ResolveCanvasPaintings(
-        WriteManifestRequest request, DbManifest? existingManifest, CancellationToken cancellationToken)
+        WriteManifestRequest request, DbManifest? existingManifest, string? manifestId, CancellationToken cancellationToken)
```
donaldgray: Can manifestId be null?

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -255,7 +255,7 @@ private List<CanvasPainting> UpdateCanvasPaintingsAfterDlcsInteractions(List<Can
     private async Task<ResolvedManifestData> ResolveCanvasPaintingsAndParentSlug(WriteManifestRequest request,
         string? manifestId = null, DbManifest? existingManifest = null, CancellationToken cancellationToken = default)
```
donaldgray: I think `string? manifestId = null` is no longer true - it's either came in from PUT or you've created for POST.

This used to be the case before refactoring at line 162 but not the manifestId generation comes first.

donaldgray (commented):
Left a few comments, some are small syntactic/consistency but a few are around whether we can avoid treating Manifest level assets as special. We are implementing this for Manifest now but if we can handle it in a more generic fashion we can add further types easier, without refactoring. Need to make a judgement call on whether we should do this now or later - we are on a fairly tight schedule so let me know if you want to discuss.

Also notes that there is no logging introduced with this PR - are we logging enough elsewhere that troubleshooting after deployment would be straight forward enough?

src/IIIFPresentation/DLCS/Converters/AssetIdConverter.cs
JackLewis-digirati: I think not? The `Model` project is designed to be lightweight and only contain models, whereas this is starting to add in logic.  This converter is (currently) only used in the `DLCS` project, so feels like a good fit to sit there - though this should be revisited if it's used elsewhere

donaldgray: Okay, leave as-is. 

I disagree that this is "logic" though, it's only about how it looks as JSON, similar to how some classes in Models use `[JsonConverter]` and `[JsonProperty]`. Although this is STJ and `[JsonConverter]` and `[JsonProperty]` are newtonsoft, which could get confusing if in the same project.

src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
JackLewis-digirati: As discussed, this has moved to collection level adjuncts

src/IIIFPresentation/API/Features/Common/Helpers/ResourceAdjunctInteractions.cs
```diff
@@ -0,0 +1,40 @@
...
+/// <summary>
+/// Generates adjunct interactions for resource level adjuncts
+/// </summary>
+public static class ResourceAdjunctInteractions
+{
+    /// <summary>
+    /// The <see cref="AssetId"/> for the manifest-level stub asset in space 0, given the resources's internal id.
+    /// </summary>
+    public static AssetId GetResourceStubAssetId(int customerId, string resourceId) =>
+        new(customerId, 0, $"Manifest_{resourceId}");
```
donaldgray: `0` - should we have a const for this?

src/IIIFPresentation/API/Features/Common/Helpers/ResourceAdjunctInteractions.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    /// <summary>
+    /// Builds an <see cref="AdjunctInteraction"/> for the given stub asset and adjunct list.
+    /// Returns null if <see cref="PresentationManifest.Adjuncts"/> is null (treat as "no change").
+    /// Note: stamps <see cref="AssetProperties.Asset"/> onto each adjunct JObject in-place.
+    /// </summary>
+    public static AdjunctInteraction? GetAdjunctInteraction(PresentationManifest presentationManifest, int customerId)
+    {
+        if (presentationManifest.Adjuncts == null) return null;
```
donaldgray, line 30: Should this early exit for null _and_ empty adjuncts? Is there any practical difference in handling?
```suggestion
        if (presentationManifest.Adjuncts.IsNullOrEmpty()) return null;
```

As is, `null` adjuncts will return `null` but empty adjuncts would return `AdjunctInteraction` with empty `Adjuncts` property. Or is this handled higher up?

src/IIIFPresentation/API.Tests/Integration/GetManifestTests.cs
```diff
@@ -48,13 +50,12 @@ public GetManifestTests(StorageFixture storageFixture, PresentationAppFactory<Pr
     {
         s3 = storageFixture.LocalStackFixture.AWSS3ClientFactory();
         dbContext = storageFixture.DbFixture.DbContext;
-        dlcsApiClient = A.Fake<IDlcsApiClient>();
+        Fake.ClearRecordedCalls(DLCSApiClient);
```
donaldgray: Is this necessary? I generally find this to indicate something is off with tests, not enough test isolation or greedy fake assertions.

donaldgray (approved):

src/IIIFPresentation/API/Features/Common/Helpers/ResourceAdjunctInteractions.cs
JackLewis-digirati: no, it being an empty list is used to signal that we should delete all adjuncts, so the short-circuit would break that - will add a comment to explain

PR #599, opened 2026-05-12
Reject distinct adjuncts across duplicate assets

### LINKED ISSUE #596: Reject duplicate Assets with differing Adjuncts
Issue #392 implemented logic to correctly handle duplicate assets in a Manifest payload. Part of this ticket detected and rejected duplicate Assets that had any differing properties

> If we receive duplicate `Asset` then we need to de-dup before passing to DLCS API but if we receive the same asset-id with anything else that differs (metadata, delivery-channels) we need to reject.

This same logic should be applied to the Adjunct property of an Asset - we can accept duplicate Assets with Adjuncts as long as the adjuncts are identical, if not reject with a 400|BadRequest.

```jsonc
// This would be rejected as "profile" differs
{
    "paintedResources": [
        {
            "asset": {
                "id": "example",
                "origin": "https://dlcs.exmaple/sample-origin",
                "mediaType": "image/jpeg",
                "adjuncts": [
                    {
                        "id": "mets.xml",
                        "externalId": "https://hosted.example/image/mets.xml",
                        "@type": "Dataset",
                        "mediaType": "text/xml",
                        "iiifLink": "seeAlso"
                    }
                ]
            }
        },
        {
            "asset": {
                "id": "example",
                "origin": "https://dlcs.exmaple/sample-origin",
                "mediaType": "image/jpeg",
                "adjuncts": [
                    {
                        "id": "mets.xml",
                        "externalId": "https://hosted.example/image/mets.xml",
                        "@type": "Dataset",
                        "mediaType": "text/xml",
                        "iiifLink": "seeAlso",
                        "profile": "this-is-different"
                    }
                ]
            }
        }
    ]
}

// This would be accepted as asset + adjunct are identical
{
    "paintedResources": [
        {
            "asset": {
                "id": "example",
                "origin": "https://dlcs.exmaple/sample-origin",
                "mediaType": "image/jpeg",
                "adjuncts": [
                    {
                        "id": "mets.xml",
                        "externalId": "https://hosted.example/image/mets.xml",
                        "@type": "Dataset",
                        "mediaType": "text/xml",
                        "iiifLink": "seeAlso"
                    }
                ]
            }
        },
        {
            "asset": {
                "id": "example",
                "origin": "https://dlcs.exmaple/sample-origin",
                "mediaType": "image/jpeg",
                "adjuncts": [
                    {
                        "id": "mets.xml",
                        "externalId": "https://hosted.example/image/mets.xml",
                        "@type": "Dataset",
                        "mediaType": "text/xml",
                        "iiifLink": "seeAlso"
                    }
                ]
            }
        }
    ]
}
```

This detection logic should happen as early in the call as possible so as to avoid doing any unnecessary work. 

The current code to handle this is in `DlcsManifestCoordinator` - https://github.com/dlcs/iiif-presentation/blob/develop/src/IIIFPresentation/API/Features/Manifest/DlcsManifestCoordinator.cs#L375-L379. See #578 for original Adjunct handling implementation. This was done in such a way that by the time the `DlcsManifestCoordinator` is reached the `"adjuncts"` property has been processed and stripped from the Asset so the equality check will need to happen before this call.

JackLewis-digirati:
## What does this change?

Resolves #596

This PR extends detection of duplicate assets to check for duplicate adjuncts

  ### The pipeline problem

  Adjuncts are part of the asset payload but are not forwarded to DLCS directly — they are extracted
  from the asset `JObject` by `ManifestPaintedResourceParser` during canvas painting resolution,
  stored on `InterimCanvasPainting.AdjunctInteraction`, and **removed** from the raw `Asset` JObject.
  By the time execution reaches `DlcsManifestCoordinator`, the adjuncts are gone and cannot be
  compared. Any adjunct conflict check must therefore happen after parsing but before the coordinator.

Additionally, due to the changes required for detecting duplicate adjuncts, it makes sense to move the detection of duplicate assets into the same location as detection of duplicate adjuncts.

  ### Where the check lives

  The new `ValidateDuplicates` method is called inside `CanvasPaintingResolver.ParseManifest`,
  immediately after `ManifestPaintedResourceParser.ParseToCanvasPainting` returns. At this point both
  data sources are available:

  - `paintedResourceCanvasPaintings` — the parsed canvas paintings, each carrying an
    `AdjunctInteraction` if the asset had adjuncts.
  - `presentationManifest.PaintedResources` — the original `PaintedResource` list, whose `Asset`
    JObjects have had adjuncts stripped but still contain all other asset fields.

  ### What `ValidateDuplicates` does

  The method runs two sequential checks and returns the first error it finds, or `null` if everything
  is valid.  This is a sequential check as there's no single collection that has both the raw JObject asset data and the parsed adjunct interaction data, so the two checks must live in separate loops

  #### Asset data check

  Iterates `paintedResources` directly (avoiding any fragile positional coupling to the canvas
  painting list). It builds a seen-dictionary keyed on `{space}/{assetId}`, where the space comes
  from the asset's own `"space"` field (falling back to the `-1` sentinel for assets without an
  explicit space). Including the space in the key ensures that the same asset ID in two different
  spaces is not treated as a conflict. If the same key is seen twice with non-equal JObjects, it
  returns `AssetsDataDoesNotMatch` (400) with a JSON diff.

  #### Adjunct check

  Iterates `paintedResourceCanvasPaintings`. It builds a seen-dictionary keyed on the user-facing
  asset key produced by `BuildAssetKey`. Comparison uses `AdjunctsEqual`, which is order-insensitive:
  adjuncts are matched by their `"id"` field using a dictionary, then deep-compared with
  `JToken.DeepEquals`. If the same asset appears twice with non-equal adjunct lists, it returns
  `AssetAdjunctsDoNotMatch` (400) with a JSON diff. Identical duplicates (or no adjuncts on either
  side) are silently allowed through.

  Both checks share `SerializeDiff`, a one-liner that runs `JsonDiffPatch` and serialises the result.

  #### `BuildAssetKey`

  Produces user-facing keys for error messages. It omits the space when the space is the `-1`
  sentinel (meaning it hasn't been allocated yet by DLCS), so users never see a meaningless `-1` in
  error output. A key is either `{customerId}/{assetId}` (no space yet) or
  `{customerId}/{space}/{assetId}` (space known).

  ### Changes to `DlcsManifestCoordinator`

  Because conflicting asset data is now caught earlier, `CreateBatches` no longer needs to
  re-validate duplicates. The 15-line block that detected conflicts, computed a diff, and returned an
  error has been replaced with a simple `TryAdd` for de-duplication.

  The `JsonDiffPatchDotNet` and `Newtonsoft.Json` imports that were only needed for the removed block
  have been cleaned up from `DlcsManifestCoordinator`.

  ### New error type

  `ModifyCollectionType.AssetsAdjunctsDoNotMatch = 27` is added to distinguish adjunct conflicts from
  general asset data conflicts (`AssetsDoNotMatch = 25`). The corresponding
  `UpsertErrorHelper.AssetAdjunctsDoNotMatch` factory method follows the existing pattern in that
  class. `UpsertErrorHelper.AssetsDataDoesNotMatch` is also extracted as a factory method (it was
  previously an inline string in `DlcsManifestCoordinator`) so both error shapes are defined in one
  place.

### DIFF
--- a/src/IIIFPresentation/API/Features/Manifest/Helpers/AssetDuplicateValidator.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/Helpers/AssetDuplicateValidator.cs
@@ -80,6 +80,9 @@ public static class AssetDuplicateValidator
     // Relies on AdjunctValidator having already enforced non-empty ids upstream.
     private static bool AdjunctsEqual(List<JObject>? previous, List<JObject>? current)
     {
+        // makes sure [] and null isn't allowed
+        if (previous is null != current is null) return false;
+        
         var left = previous ?? [];
         var right = current ?? [];
         if (left.Count != right.Count) return false;
--- a/src/IIIFPresentation/API/Features/Manifest/Helpers/AssetDuplicateValidator.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/Helpers/AssetDuplicateValidator.cs
@@ -74,12 +74,16 @@ public static class AssetDuplicateValidator
         return null;
     }
 
-    // Order-insensitive: adjuncts are matched by id, then deep-compared
+    // Compares two adjunct lists without regard to order.
+    // Each adjunct is looked up by its id in the previous list, then deep-compared —
+    // so reordering adjuncts is not treated as a conflict, but any property change is.
+    // Relies on AdjunctValidator having already enforced non-empty ids upstream.
     private static bool AdjunctsEqual(List<JObject>? previous, List<JObject>? current)
     {
         var left = previous ?? [];
         var right = current ?? [];
         if (left.Count != right.Count) return false;
+        // index the previous adjuncts by id for O(n) lookup
         var leftById = left.ToDictionary(j => j.GetRequiredValue<string>(AssetProperties.Id));
         return right.All(j =>
             leftById.TryGetValue(j.GetRequiredValue<string>(AssetProperties.Id), out var match)
--- a/src/IIIFPresentation/API/Features/Manifest/Helpers/AssetDuplicateValidator.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/Helpers/AssetDuplicateValidator.cs
@@ -1,0 +1,97 @@
+using API.Features.Common.Helpers;
+using Core.Helpers;
+using DLCS.Models;
+using JsonDiffPatchDotNet;
+using Models.API.Manifest;
+using Models.DLCS;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Services.Manifests.Helpers;
+using Services.Manifests.Model;
+using PresUpdateResult = API.Infrastructure.Requests.ModifyEntityResult<Models.API.Manifest.PresentationManifest, Models.API.General.ModifyCollectionType>;
+
+namespace API.Features.Manifest.Helpers;
+
+public static class AssetDuplicateValidator
+{
+    public static PresUpdateResult? ValidateDuplicates(
+        List<InterimCanvasPainting> paintedResourceCanvasPaintings,
+        List<PaintedResource>? paintedResources)
+        => ValidateDuplicateAssets(paintedResourceCanvasPaintings, paintedResources)
+           // run the adjunct validator if the asset validator returned no errors
+           ?? ValidateDuplicateAdjuncts(paintedResourceCanvasPaintings);
+
+    private static PresUpdateResult? ValidateDuplicateAssets(
+        List<InterimCanvasPainting> paintedResourceCanvasPaintings,
+        List<PaintedResource>? paintedResources)
+    {
+        if (paintedResources == null) return null;
+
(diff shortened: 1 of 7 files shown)


src/IIIFPresentation/API/Features/Manifest/Helpers/AssetDuplicateValidator.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        return null;
+    }
+
+    // Compares two adjunct lists without regard to order.
+    // Each adjunct is looked up by its id in the previous list, then deep-compared —
+    // so reordering adjuncts is not treated as a conflict, but any property change is.
+    // Relies on AdjunctValidator having already enforced non-empty ids upstream.
+    private static bool AdjunctsEqual(List<JObject>? previous, List<JObject>? current)
```
donaldgray, line 81: This looks like it would treat `null` and `[]` as equal, which isn't the case for adjuncts.

donaldgray (approved):

PR #600, opened 2026-05-14
Handle unknown props for Manifests and Collections

### LINKED ISSUE #588: Improve IIIF serialisation/deserialisation support
Fix for ticket https://github.com/digirati-co-uk/iiif-net/issues/62 will extend iiif-net nuget package serialisation/deserialisation to support round-tripping JSON, including unknown properties. 

Once that fix is ready, we'll need to upgrade to that version used in IIIF-net. However, updating the version with fixes for `#62` alone is not enough for IIIF-Presentations needs. In summary the problem would be:
* We have [`PresentationManifest : Manifest`](https://github.com/dlcs/iiif-presentation/blob/b6aad10c7b64a4e1d4b069fc482d7f3551c71ffc/src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs) class that contains properties custom to IIIF-P (`publicId`, `slug` etc). We deserialise to this type to be able read these properties in addition to the standard IIIF Properties.
* We also deserialise to `Manifest` when saving to S3. By doing this we stripped out any custom properties, only saving "valid" IIIF (however, this didn't quite work as iiif-net isn't completed and consumers could validly supply their own props).
* With the fix for https://github.com/digirati-co-uk/iiif-net/issues/62 the `Manifest` serialised object will contain _all_ properties - whether they are on the underlying object or now.

So, we need a solution that will:
* Allow us to read and parse the known "off-spec" properties (that we currently do via `PresentationManifest`).
* Allow us to save the Manifest to S3, without those known "off-spec" properties, but possibly including some other off-spec properties.
* Allow us to load the Manifest from S3 and re-add the known "off-spec" properties for returning to users. 

Ideally this is handled in iiif-net nuget package, allowing some customisation of ser/deser via a properties object.

There was a discussion around how we implement this in Slack with @JackLewis-digirati , @tomcrane , @p-kaczynski and @frank-kerrigan (https://digirati.slack.com/archives/C0AUF2MGVL6/p1776845037820909).

donaldgray:
## What does this change?

Resolves #588

Upgrades iiif-net from `0.3.12` to `0.4.1` and fixes a regression introduced by the new version.

iiif-net [`0.4.0`](https://github.com/digirati-co-uk/iiif-net/releases/tag/v0.4.0) adds `AdditionalProperties` (a `Dictionary<string, JToken>`) to `JsonLdBase` via Newtonsoft's `[JsonExtensionData]`. This means that when a user submits a manifest or collection containing `PresentationManifest`/`PresentationCollection`-specific fields (`slug`, `parent`, `paintedResources`, etc) and the request body is deserialised as plain `Manifest`/`Collection`, those fields now land in `AdditionalProperties` and are re-serialised when `AsJson()` is called — polluting the IIIF resource stored in S3.

Fix was to strip the known presentation-layer keys at deserialisation time, before the object reaches any S3 write path.

* `PresentationManifest` and `PresentationCollection` each gain a `static readonly string[] PresentationPropertyKeys` listing their JSON property names.
* Two extension methods `ToManifest()` and `ToCollection()` in `Services` deserialise and immediately strip those keys using iiif-net's `WithoutAdditionalProperties()`. Unknown custom properties the user added are preserved.
* `ManifestWriteService` uses `ToManifest()` instead of `FromJson<Manifest>()` directly.
* `ConvertCollectionToIIIF` in `RequestBodyX` is de-generified (all callers used `Collection`) and now calls `ToCollection()` internally, covering all collection write paths (`CreateCollection`, `UpsertCollection`, `PostHierarchicalCollection`).
* Unit tests in `IIIFSerialisationXTests` cover stripping, custom-property preservation, and round-trip serialisation. Integration tests assert that the S3-stored manifest/collection contains none of the presentation property keys.

### DIFF
--- a/src/IIIFPresentation/Models/API/Collection/PresentationCollection.cs
+++ b/src/IIIFPresentation/Models/API/Collection/PresentationCollection.cs
@@ -3,6 +3,12 @@ namespace Models.API.Collection;
 
 public class PresentationCollection : IIIF.Presentation.V3.Collection, IPresentation
 {
+    public static readonly string[] PresentationPropertyKeys =
+    [
+        "slug", "publicId", "flatId", "parent", "itemsOrder", "totalItems",
+        "created", "modified", "createdBy", "modifiedBy", "tags", "totals", "view"
+    ];
+
     public string? PublicId { get; set; }
     public string? FlatId { get; set; }
     
--- a/src/IIIFPresentation/Models/API/Collection/PresentationCollection.cs
+++ b/src/IIIFPresentation/Models/API/Collection/PresentationCollection.cs
@@ -3,6 +3,10 @@ namespace Models.API.Collection;
 
 public class PresentationCollection : IIIF.Presentation.V3.Collection, IPresentation
 {
+    /// <summary>
+    /// A collection of properties that are not part of the IIIF Presentation API spec and are custom to the
+    /// IIIF-Presentation.
+    /// </summary>
     public static readonly string[] PresentationPropertyKeys =
     [
         "slug", "publicId", "flatId", "parent", "itemsOrder", "totalItems",
--- a/src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
+++ b/src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
@@ -6,6 +6,13 @@ namespace Models.API.Manifest;
 
 public class PresentationManifest : IIIF.Presentation.V3.Manifest, IPresentation
 {
+    public static readonly string[] PresentationPropertyKeys =
+    [
+        "slug", "publicId", "parent", "created", "modified",
+        "createdBy", "modifiedBy", "flatId", "paintedResources", "space", "ingesting",
+        "fullPath", "currentlyIngesting"
+    ];
+
     [JsonProperty(Order = 6)] public string? Slug { get; set; }
     [JsonProperty(Order = 7)] public string? PublicId { get; set; }
     [JsonProperty(Order = 8)] public string? Parent { get; set; }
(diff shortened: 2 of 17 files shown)


src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
```diff
@@ -6,6 +6,17 @@ namespace Models.API.Manifest;
...
 {
+    /// <summary>
+    /// A collection of properties that are not part of the IIIF Presentation API spec and are custom to the
+    /// IIIF-Presentation.
+    /// </summary>
+    public static readonly string[] PresentationPropertyKeys =
+    [
+        "slug", "publicId", "parent", "created", "modified",
+        "createdBy", "modifiedBy", "flatId", "paintedResources", "space", "ingesting",
+        "fullPath", "currentlyIngesting"
```
JackLewis-digirati, line 17: think this is missing `adjuncts`?

src/IIIFPresentation/Models/API/Collection/PresentationCollection.cs
```diff
@@ -3,6 +3,16 @@ namespace Models.API.Collection;
...
 public class PresentationCollection : IIIF.Presentation.V3.Collection, IPresentation
 {
+    /// <summary>
+    /// A collection of properties that are not part of the IIIF Presentation API spec and are custom to the
+    /// IIIF-Presentation.
+    /// </summary>
+    public static readonly string[] PresentationPropertyKeys =
+    [
+        "slug", "publicId", "flatId", "parent", "itemsOrder", "totalItems",
+        "created", "modified", "createdBy", "modifiedBy", "tags", "totals", "view"
```
JackLewis-digirati, line 13: A lot of these properties are shared by `PresentationManifest` - worth creating a single list with all the shared properties and referencing them?

JackLewis-digirati (approved):

src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
donaldgray: Added - rebased ontop of #598 and didn't add new prop.

src/IIIFPresentation/Models/API/Collection/PresentationCollection.cs
donaldgray: I considered that initially but felt it was clearer to have 2 separate, explicit lists - easier to see at a glance what the relevant props per type are.

PR #601, opened 2026-05-14
Implementation of manifest-level adjuncts being reflected in items

### LINKED ISSUE #595: Manifest level adjuncts - output on Manifest
See parent ticket for additional information.

> [!IMPORTANT]
> See https://github.com/dlcs/iiif-presentation/issues/580 for Canvas level adjunct implementation.

This refers to both sync and async processed Manifests.

Suggested steps
* [ ] Final Manifest reflects adjuncts, `"seeAlso"`, `"rendering"` or `"annotations"`.
* [ ] If the above properties already exist they are appended to and de-duplicated based in `"id"`.
* [ ] The Manifest level adjuncts will be found via the Manifest-level stub asset.

JackLewis-digirati:
## What does this change?

Resolves #595

This PR is the other side of manifest level adjuncts, and allows the manifest level adjuncts to update top level properties like `seeAlso`, `renderings` and `annotations`.

This mirrors the work done on canvas-level adjuncts, in that it will deduplicate and merge any existing values in these properties with the ones coming from `adjuncts`.

This also fixes an issue where the incorrect status code was recorded for a hosted manifest-level adjunct.  It should now correctly record a `202`


JackLewis-digirati:
## How this works

Manifest-level adjuncts (`seeAlso`, `rendering`, `annotations` on the manifest itself, rather than individual canvases) are stored in DLCS via a **stub asset** in space 0 with the id `Manifest_{internalManifestId}`. This stub carries no binary content — it exists purely to host adjuncts.

When DLCS returns a Named Query manifest, it includes a canvas for every asset it knows about, including this stub. This PR teaches `ManifestMerger` to recognise that stub canvas and apply its IIIF properties up to the manifest level using the existing `AddDistinctById` dedup logic (so user-set values and adjunct-derived values are merged, not overwritten).

### Write path (`canBeBuiltUpfront = true`)

`ManifestMerger.MergeManifest` now calls `ApplyManifestLevelAdjuncts` after `ProcessCanvasPaintings`. The stub canvas is identified by matching the NQ canvas id against the expected stub `AssetId` (`{customer}/0/Manifest_{id}`). Its `seeAlso`/`rendering`/`annotations` are applied to the manifest, and the result is written to S3. `ManifestWriteService` then stamps those fields back onto the `PresentationManifest` so they appear in the API response.

### Write path (async / batch)

The staging manifest is saved without adjuncts (they aren't known yet). When DLCS signals batch completion via SQS, `BatchCompletionMessageHandler` triggers `ManifestS3Manager.UpsertManifestFromStagingInStorage`, which rebuilds via `ManifestMerger` — picking up the stub canvas and its adjuncts at that point.

### Read path (GET)

`ManifestReadService.SetManifestLevelAdjuncts` already populated `manifest.Adjuncts` from the stub asset for the API response. The change here means the IIIF properties (`seeAlso` etc.) are also present in the stored S3 manifest, so they round-trip correctly.

### What the stub canvas is NOT

The stub canvas is filtered out of `manifest.Items` — it never appears as a viewable canvas in the manifest.

src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
```diff
@@ -48,24 +56,49 @@ public class ManifestMerger(SettingsBasedPathGenerator pathGenerator, IPathRewri
...
 
+        ValidateManifests(baseManifest, namedQueryManifest);
         canvasLookups.Initialise(namedQueryManifest!, baseManifest, canvasPaintings);
         BuildItems(baseManifest, canvasPaintings);
         SetManifestContext(baseManifest, namedQueryManifest!);
 
         return baseManifest;
     }
     
+    private Manifest ApplyManifestLevelAdjuncts(Manifest baseManifest, Manifest? namedQueryManifest, int customerId,
```
donaldgray: IMO this is unnecessarily tied to `Manifest`, it could be `ResourceBase`? We will be adding adjuncts to other resources in the future so a small change here can make like easier when we do that.

It's likely we'll need to do more work to "de-manifest" this class in general but the less that's tied now the easier a refactore will be.

src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
```diff
@@ -48,24 +56,49 @@ public class ManifestMerger(SettingsBasedPathGenerator pathGenerator, IPathRewri
...
+        string manifestId)
+    {
+        if (namedQueryManifest?.Items.IsNullOrEmpty() ?? true) return baseManifest;
```
donaldgray: Small - is this/can this be checked before entering here? If not I wonder if it should, we can't really do anything if the NQ has no canvases. Same with `namedQueryManifest` being nullable - can we protect against that before entering this method and therefore make the param non-nullable?

src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
```diff
@@ -48,24 +56,49 @@ public class ManifestMerger(SettingsBasedPathGenerator pathGenerator, IPathRewri
...
+
+        var stubAssetId = ResourceAdjunctInteractions.GetResourceStubAssetId(baseManifest, customerId, manifestId);
+
+        var stubCanvas = namedQueryManifest.Items!.FirstOrDefault(canvas =>
```
donaldgray: `CanvasLookups` was introduced in https://github.com/dlcs/iiif-presentation/pull/590 to try and help with this sort of call. From PR notes

> Internally refactored some logic into `CanvasLookups` - this builds a couple of lookups. Keeps ManifestMerger leaner and should make it easier to track and find stub-assets for later tickets.

I think it'd be better to use that becase:
* It would make callsite easier and more intuitive, less parsing and error handling - just _"give me the Canvas for this resource, if you have one"_
* We've already done the work of iterating through `.Items` to build a Dictionary.
* `int customerId` is added here but there's logic in `CanvasLookups` to determine what the `CustomerId` is so would save passing that down.
* `GetAssetIdFromNamedQueryCanvasId()` is already called for every canvas in the NQ (in `BuildAssetIdToCanvasLookup()`) so this is unnecessary here - the `catch` block would never be hit.

```cs
// Internally CanvasLookup has an AssetId:Canvas dict
// It also can store CustomerId from when parsing
// ResourceAdjunctInteractions only needs ResourceBase, CustomerId and ResourceId to build the identifier
var stubCanvas = canvasesLookup.GetStubCanvas(resourceBase, resourceId);
if (stubCanvas != null)
{
    logger.LogDebug("Found manifest-level stub canvas for {ManifestId}, applying adjuncts", manifestId);
    SetAssetDerivedProperties(resourceBase, stubCanvas, stubAssetId.ToString());
}

return baseManifest;
```

Happy to discuss, let me know if so.

src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
```diff
@@ -18,25 +18,33 @@ namespace Services.Manifests;
...
+        List<CanvasPainting>? canvasPaintings, int customerId, string manifestId);
 }
 
 public class ManifestMerger(SettingsBasedPathGenerator pathGenerator, IPathRewriteParser pathRewriteParser, 
     ILogger<ManifestMerger> logger) : IManifestMerger
 {
     private readonly CanvasLookups canvasLookups = new(pathRewriteParser, logger);
-    
+
+    public Manifest MergeManifest(Manifest baseManifest, Manifest? namedQueryManifest,
```
donaldgray, line 42: XML comments? This now private `ProcessCanvasPaintings` has a desc, does that fully apply here?

src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
JackLewis-digirati: Went with `StructureBase` as it's used by `SetAssetDerivedProperties` that needs it for `annotations`

src/IIIFPresentation/Services/Manifests/ManifestMerger.cs
JackLewis-digirati: I don't think it makes a difference as we have control of the NQ, but the catch would be hit if we had a broken canvas id on the NQ (so we could attempt to fill out _some_ of the manifest), by contrast `BuildAssetIdToCanvasLookup` doesn't build the dictionary if there's any error - but suspect if this does occur it's best to leave everything than try and half build something

donaldgray: The `CanvasLookup` throws on `.Initialise()` if the format isn't known so this code would never be reached

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -291,20 +289,19 @@ private async Task<PresUpdateResult> SaveToS3AndGenerateResult(WriteManifestRequ
...
     private async Task<PresUpdateResult> GeneratePresentationSuccessResult(PresentationManifest presentationManifest,
-        int customerId, DbManifest dbManifest, bool hasAssets, DlcsInteractionResult dlcsInteractionResult,
+        int customerId, DbManifest dbManifest, DlcsInteractionResult dlcsInteractionResult,
         WriteResult writeResult, CancellationToken cancellationToken)
     {
         return PresUpdateResult.Success(
             presentationManifest.SetGeneratedFields(dbManifest, pathGenerator, savedManifestPathGenerator,
                 await dlcsManifestCoordinator.GetAssets(customerId, dbManifest, cancellationToken)),
-            hasAssets && !dlcsInteractionResult.CanBeBuiltUpfront
+            !dlcsInteractionResult.CanBeBuiltUpfront
```
donaldgray: Rather passing a default `WriteResult` and potentially changing it here, `CreateInternal` and `UpdateInternal` can calculate which result to use pass in the correct write result as they know both know `CanBeBuiltUpfront`. 

Currently
```cs
// WriteResult.Created is passed but it might get overwritten depending on CanBeBuiltUpfront 
return await SaveToS3AndGenerateResult(request, dbManifest!, dlcsResult.InteractionResult!, WriteResult.Created);
```

Could be
```cs
// Work out the result initially
var result = dlcsResult.InteractionResult.CanBeBuiltUpfront ? WriteResult.Created : WriteResult.Accepted;
return await SaveToS3AndGenerateResult(request, dbManifest!, dlcsResult.InteractionResult!, result);
```

donaldgray (approved):

PR #606, opened 2026-05-25
Store Incoming Payloads

### LINKED ISSUE #576: Store incoming payloads
There may be instances in the future where it will be necessary to reprocess resources; e.g. due to a change in processing (e.g. switching to P4 support) or due to a general change in output resources (e.g. offering [content search](https://iiif.io/api/search/2.0/) services for Manifests) or for tracking changes to downstream Assets or Adjuncts (see #536).

A simple way to support the sorts of behaviours required by above is to store the incoming payload in addition to the "final" resource. This will make it simpler to reprocess items - we can effectively reprocess the initial payload, without needing to reverse-engineer the final resource.

> [!NOTE]
> Any Manifests created before the implementation of this ticket will be missing the "incoming payload". To help identify these we will add a setting that denotes when this functionality was deployed. Any endpoints that require an "incoming payload" (say, `/refresh`) can use this date to know whether it is supported for a particular resource.

Summary:
* [ ] Store incoming Manifest/Collection payload if the final resource will be different (e.g. if we receive IIIF and store it as-is, there's no value in storing it).
* [ ] We shouldn't overwrite on reception each time - we should manage when this is overwritten, similar to how the staging versions of Manifests work (capture workflow somewhere)
* [ ] Ensure the new payload is removed when Manifest/Collection is deleted
* [ ] Have an appSetting date that denotes when functionality is deployed. By default this is `null` (signifying all resources will have it). If there is a DateTime set then only those resources created after that DateTime will have it.
* [ ] Update https://github.com/dlcs/iiif-presentation/blob/develop/docs/notes/storage-keys.md with chosen storage location.
* [ ] If we fail to save the payload for whatever reason, don't fail the entire request (assuming everything else works!)

p-kaczynski:
## What does this change?

Implements #576 

* New `DateTimeOffset?` setting `StoresPayloadsSince` loaded as a part of new `BehaviourSettings` - declares since when we can expect an entity to have their original payload stored - saves reaching to S3 for nothing. If not set, code assumes all entities should have that payload stored (within other rules)
* Refactored `IIIFS3Service` to allow saving/reading raw streams, as we treat the payload as arbitrary `string`, not serialised/deserialised IIIF entity
* Added logic to API behaviour that includes storing the original, if the `PaintedResources` were included in manifest, as this will make normally-saved manifest differ from the payload
* Expanded `BucketHelperX` to take an enum pointing to a "variant" of location - this expands on the original `isStaging`  flag
* Modified `BatchCompletionMessageHandlerTests` to work without new functionality, and included two-scenario `HandleMessage_SavesResultingManifest_ToS3` that also tests the new behaviour.

## Configuration Changes

> [!NOTE]
>
> |---|---|---|---|---|
> | API |  `StoresPayloadsSince` | N | Timestamp of moment since storing payloads | `null` |

### DIFF
--- a/docs/notes/storage-keys.md
+++ b/docs/notes/storage-keys.md
@@ -7,11 +7,13 @@ IIIF-Presentation makes use of S3 to store manifests and collections. These keys
 
 See `BucketHelperX` for code that generates keys.
 
-| Name             | Format                                                | Example                            | Description                                                                                         |
-| ---------------- | ----------------------------------------------------- | ---------------------------------- | --------------------------------------------------------------------------------------------------- |
-| Manifest         | `{Storage}/{Customer}/manifests/{Manifest-Id}`        | `iiif-p/1/manifests/abc123`        | IIIF manifests, whether saved as-is or generated                                                    |
-| Staged Manifest* | `{Storage}/staged/{Customer}/manifests/{Manifest-Id}` | `iiif-p/staged/1/manifests/abc123` | In-flight generated IIIF manifests. Will contain in-complete `"items"` that will be populated later |
-| IIIF Collections | `{Storage}/{Customer}/collections/{Collection-Id}`    | `iiif-p/1/collections/abc123`      | IIIF Collections                                                                                    |
+| Name                     | Format                                                          | Example                                      | Description                                                                                        |
+|--------------------------|-----------------------------------------------------------------|----------------------------------------------|----------------------------------------------------------------------------------------------------|
+| Manifest                 | `{Storage}/{Customer}/manifests/{Manifest-Id}`                  | `iiif-p/1/manifests/abc123`                  | IIIF manifests, whether saved as-is or generated                                                   |
+| Staged Manifest          | `{Storage}/staging/{Customer}/manifests/{Manifest-Id}`          | `iiif-p/staging/1/manifests/abc123`          | In-flight generated IIIF manifests. Will contain in-complete `"items"` that will be populated later |
+| Orignal Manifest Payload | `{Storage}/original/{Customer}/manifests/{Manifest-Id}`         | `iiif-p/original/1/manifests/abc123`         | Stores the original payload if the it'd have been modified by processing                           |
+| Orignal Manifest Payload | `{Storage}/original-staging/{Customer}/manifests/{Manifest-Id}` | `iiif-p/original-staging/1/manifests/abc123` | Stores the payload in the update scenario that requires background handler
+| IIIF Collections         | `{Storage}/{Customer}/collections/{Collection-Id}`              | `iiif-p/1/collections/abc123`                | IIIF Collections                                                                                   |
 
 * `*` indicates that this is a proposed key and not currently used. `/staged/` key will allow easy clean-up of 'stale' in-flight manifests using prefix lifecycle rules.
 
--- a/docs/notes/storage-keys.md
+++ b/docs/notes/storage-keys.md
@@ -7,15 +7,28 @@ IIIF-Presentation makes use of S3 to store manifests and collections. These keys
 
 See `BucketHelperX` for code that generates keys.
 
-| Name                     | Format                                                          | Example                                      | Description                                                                                        |
-|--------------------------|-----------------------------------------------------------------|----------------------------------------------|----------------------------------------------------------------------------------------------------|
-| Manifest                 | `{Storage}/{Customer}/manifests/{Manifest-Id}`                  | `iiif-p/1/manifests/abc123`                  | IIIF manifests, whether saved as-is or generated                                                   |
-| Staged Manifest          | `{Storage}/staging/{Customer}/manifests/{Manifest-Id}`          | `iiif-p/staging/1/manifests/abc123`          | In-flight generated IIIF manifests. Will contain in-complete `"items"` that will be populated later |
-| Orignal Manifest Payload | `{Storage}/original/{Customer}/manifests/{Manifest-Id}`         | `iiif-p/original/1/manifests/abc123`         | Stores the original payload if the it'd have been modified by processing                           |
-| Orignal Manifest Payload | `{Storage}/original-staging/{Customer}/manifests/{Manifest-Id}` | `iiif-p/original-staging/1/manifests/abc123` | Stores the payload in the update scenario that requires background handler
-| IIIF Collections         | `{Storage}/{Customer}/collections/{Collection-Id}`              | `iiif-p/1/collections/abc123`                | IIIF Collections                                                                                   |
+| Name                             | Format                                                          | Example                                      | Description                                                                                         |
+| -------------------------------- | --------------------------------------------------------------- | -------------------------------------------- | --------------------------------------------------------------------------------------------------- |
+| Manifest                         | `{Storage}/{Customer}/manifests/{Manifest-Id}`                  | `iiif-p/1/manifests/abc123`                  | IIIF manifests, whether saved as-is or generated                                                    |
+| Original Manifest Payload        | `{Storage}/{Customer}/manifests/{Manifest-Id}/original`         | `iiif-p/1/manifests/abc123/original`         | Stores the original payload if the required modification by processing                              |
+| Staged Manifest                  | `{Storage}/staging/{Customer}/manifests/{Manifest-Id}`          | `iiif-p/staging/1/manifests/abc123`          | In-flight generated IIIF manifests. Will contain in-complete `"items"` that will be populated later |
+| Staged Original Manifest Payload | `{Storage}/staging/{Customer}/manifests/{Manifest-Id}/original` | `iiif-p/staging/1/manifests/abc123/original` | Stores the payload in the update scenario that requires background handler                          |
+| IIIF Collections                 | `{Storage}/{Customer}/collections/{Collection-Id}`              | `iiif-p/1/collections/abc123`                | IIIF Collections                                                                                    |
 
-* `*` indicates that this is a proposed key and not currently used. `/staged/` key will allow easy clean-up of 'stale' in-flight manifests using prefix lifecycle rules.
+## Prefixes
+
+* `/{Customer}/` is the general prefix, keeps all resources for a customer together.
+* `/staging/` groups those resources that are in-flight. These are relatively transient to useful to be able to identify independantly.
+
+## Staging / Original
+
+The Manifest can exist in multiple places (using terminology from above)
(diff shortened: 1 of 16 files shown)


docs/notes/storage-keys.md
```diff
@@ -7,11 +7,13 @@ IIIF-Presentation makes use of S3 to store manifests and collections. These keys
...
+| Manifest                 | `{Storage}/{Customer}/manifests/{Manifest-Id}`                  | `iiif-p/1/manifests/abc123`                  | IIIF manifests, whether saved as-is or generated                                                   |
+| Staged Manifest          | `{Storage}/staging/{Customer}/manifests/{Manifest-Id}`          | `iiif-p/staging/1/manifests/abc123`          | In-flight generated IIIF manifests. Will contain in-complete `"items"` that will be populated later |
+| Orignal Manifest Payload | `{Storage}/original/{Customer}/manifests/{Manifest-Id}`         | `iiif-p/original/1/manifests/abc123`         | Stores the original payload if the it'd have been modified by processing                           |
+| Orignal Manifest Payload | `{Storage}/original-staging/{Customer}/manifests/{Manifest-Id}` | `iiif-p/original-staging/1/manifests/abc123` | Stores the payload in the update scenario that requires background handler
+| IIIF Collections         | `{Storage}/{Customer}/collections/{Collection-Id}`              | `iiif-p/1/collections/abc123`                | IIIF Collections                                                                                   |
```
JackLewis-digirati: The ticket mentions `collections` being stored as well, but code seems to omit any mention of them - is this implemented?

src/IIIFPresentation/AWS/Helpers/IIIFS3Service.cs
```diff
@@ -24,6 +25,12 @@ public Task SaveIIIFToS3(ResourceBase iiifResource, IHierarchyResource dbResourc
         bool saveToStaging, CancellationToken cancellationToken);
 
     public Task DeleteIIIFFromS3(IHierarchyResource dbResource, bool fromStaging = false);
+
+    Task SaveToS3(IHierarchyResource dbResource, BucketLocationType locationType,
+        string iiifJson, CancellationToken cancellationToken);
+
+    Task<Stream?> ReadStreamFromS3(IHierarchyResource dbResource,
```
JackLewis-digirati:
```suggestion
    public Task<Stream?> ReadStreamFromS3(IHierarchyResource dbResource,
```

src/IIIFPresentation/Services/Manifests/AWS/ManifestS3Manager.cs
```diff
@@ -67,7 +89,7 @@ await dlcsOrchestratorClient.RetrieveAssetsForManifest(dbManifest.CustomerId, db
             dbManifest.CustomerId,
             dbManifest.Id);
 
-        await SaveManifestInStorage(mergedManifest, dbManifest, false, cancellationToken);
+        await SaveManifestInStorage(mergedManifest, dbManifest,originalPayload, false, cancellationToken);
```
JackLewis-digirati:
```suggestion
        await SaveManifestInStorage(mergedManifest, dbManifest, originalPayload, false, cancellationToken);
```

src/IIIFPresentation/AWS/Helpers/IIIFS3Service.cs
```diff
@@ -33,15 +40,29 @@ public class IIIFS3Service(
...
+    public Task<Stream?> ReadStreamFromS3(IHierarchyResource dbResource,
+        BucketLocationType locationType, CancellationToken cancellationToken) =>
+        ReadStreamFromS3(dbResource.GetResourceBucketKey(locationType), cancellationToken);
 
     public async Task<T?> ReadIIIFFromS3<T>(string bucketKey,
         CancellationToken cancellationToken) where T : ResourceBase, new()
+    {
+        var stream = await ReadStreamFromS3(bucketKey, cancellationToken);
+        if (stream == null)
+            return null;
```
JackLewis-digirati:
```suggestion
        if (stream == null) return null;
```

src/IIIFPresentation/AWS/Helpers/IIIFS3Service.cs
```diff
@@ -24,6 +25,12 @@ public Task SaveIIIFToS3(ResourceBase iiifResource, IHierarchyResource dbResourc
         bool saveToStaging, CancellationToken cancellationToken);
 
     public Task DeleteIIIFFromS3(IHierarchyResource dbResource, bool fromStaging = false);
+
+    Task SaveToS3(IHierarchyResource dbResource, BucketLocationType locationType,
```
JackLewis-digirati:
```suggestion
    public Task SaveToS3(IHierarchyResource dbResource, BucketLocationType locationType,
```

src/IIIFPresentation/Core/Streams/StreamX.cs
```diff
@@ -11,4 +12,18 @@ public static class StreamX
...
+    public static async Task<string> ReadStreamAsStringAsync(
+        this Stream stream,
+        CancellationToken cancellationToken = default)
+    {
+        using var reader = new StreamReader(
+            stream,
+            Encoding.UTF8,
+            detectEncodingFromByteOrderMarks: true,
+            bufferSize: 1024,
+            leaveOpen: true);
```
JackLewis-digirati: why `leaveOpen`?

src/IIIFPresentation/Services/Manifests/AWS/ManifestS3Manager.cs
```diff
@@ -33,19 +37,37 @@ public async Task UpsertManifestFromStagingInStorage(Models.Database.Collections
...
 
-        await UpsertManifest(manifest!, dbManifest, cancellationToken);
+        // Future improvement would be to perform a copy without reading
+        string? stagedOriginal = null;
+        if (behaviour.CurrentValue.ShouldHaveStoredOriginal(dbManifest.Created))
+        {
+            var stagedOriginalStream =
+                await iiifS3.ReadStreamFromS3(dbManifest, BucketLocationType.OriginalStaging, cancellationToken);
+            if (!stagedOriginalStream.IsNull())
+                stagedOriginal = await stagedOriginalStream.ReadStreamAsStringAsync(cancellationToken);
```
JackLewis-digirati, line 50:
```suggestion
            if (!stagedOriginalStream.IsNull())
            {
                stagedOriginal = await stagedOriginalStream.ReadStreamAsStringAsync(cancellationToken);
            }
```

docs/notes/storage-keys.md
p-kaczynski: Ticket says "only if the payload and saved differs" - and technically yes, this might differ as those are stripped from the payload's "additional properties":
```
public static readonly string[] PresentationPropertyKeys =
    [
        "slug", "publicId", "flatId", "parent", "itemsOrder", "totalItems",
        "created", "modified", "createdBy", "modifiedBy", "tags", "totals", "view"
    ];
```
however it's either stuff we add from db anyway, or calculated numbers, leaving perhaps `tags` and `view`.

When I asked Donald on Slack he kinda just spoke about manifests, as the `paintedResources` indeed are replaced and vastly different.

If `tags` and `view` are indeed "worthy preserving" I guess it can be easily added (just extending `UploadToS3IfRequiredAsync` with raw request data), but _determining_ whether it should be stored is a bit of a pain:

1. We can run a top-level JSON scan for those specific props - a tool exists on that old branch for exactly this kind of checks: https://github.com/dlcs/iiif-presentation/blob/5cb284959048962b040066a17e22d047a3c4d509/src/IIIFPresentation/API/Converters/FastJsonPropertyRead.cs - but even though it's pretty fast, it's still kinda lots of operations
2. Alternatively the `IIIFSerialisationX` could be extended to have sth like `out bool shouldPreserveOriginal` and a check being done between deserialization and removal of additional properties.

So, if yous think this is worth the effort rn, it shouldn't take more than a couple hours.

donaldgray: Resolving - as mentioned Collections are saved-as is and not modified (aside from populating DB stored values) so can leave as is. If/when Collections do store _stuff_ that needs processing (say adjuncts) then we can extend.

src/IIIFPresentation/AWS/Helpers/IIIFS3Service.cs
```diff
@@ -62,11 +84,27 @@ public async Task SaveIIIFToS3(ResourceBase iiifResource, IHierarchyResource dbR
...
-            dbResource.GetResourceBucketKey(saveToStaging));
+            dbResource.GetResourceBucketKey(locationType));
         await bucketWriter.WriteToBucket(item, iiifJson, "application/json", cancellationToken);
     }
-    
+
+    /// <summary>
+    /// Delete a single resource from S3 at the specified location
+    /// </summary>
+    public Task DeleteFromS3(IHierarchyResource dbResource, BucketLocationType locationType)
```
JackLewis-digirati, line 103: should this comment go on the interface?

src/IIIFPresentation/Core/Settings/BehaviourSettings.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    /// <summary>
+    /// Marks a point in time since the payload storing functionality was deployed.
+    /// This serves as a way to limit other functionality that depends on storing payloads to items that were
+    /// created afterwards.
+    /// null value means all items are expected to have been created after that functionality was deployed.
+    /// </summary>
+    public DateTimeOffset? StoresPayloadsSince { get; set; }
+
+    public bool ShouldHaveStoredOriginal(DateTime created)
```
JackLewis-digirati, line 16: is there a case when we expected the payload to be stored, but didn't?

JackLewis-digirati (approved):

src/IIIFPresentation/Core/Settings/BehaviourSettings.cs
donaldgray: Added XML comment explaining use. The "should" seems valid as it's only indicating that there may be a payload there based on the date, rather than promising it will be. For that to happen it would need to take into account other aspects of the request - whether it required external content, whether it succeeded etc.

src/IIIFPresentation/AWS/Helpers/IIIFS3Service.cs
donaldgray: Yeah - good shout. Moved this and a few other XML comments from implementation to interface and added missing calls. In doing so realised that 1 public overload of `ReadIIIFFromS3()` was only used in 1 place to replaced usages of that.

PR #609, opened 2026-05-27
Add text-services RFC

donaldgray:
## What does this change?

New RFC to outline how we can use text-services when processing Manifests.

### DIFF
--- a/docs/rfcs/0007-text-services.md
+++ b/docs/rfcs/0007-text-services.md
@@ -40,7 +40,7 @@ When sending a Manifest to IIIF-Presentation we will support a new `"pipeline"` 
 {
     "pipeline": {
         "text": {
-            "index": true
+            "action": "Index"
         }
     }
 }
@@ -49,7 +49,9 @@ When sending a Manifest to IIIF-Presentation we will support a new `"pipeline"` 
 Where:
 * `"pipeline"` is the top level containing property for indicating that there is some further processing to be done to the Manifest as a whole.
 * `"text"` indicates the type of pipeline. At this time this is the only accepted value.
-* `"index": true` signifies that we want to index any text associated with this Manifest.
+* `"action"` accepts a list of possible actions to take for this pipeline.
+* `"Index"` signifies that we want to index any text associated with this Manifest.
+  * This will be extended in the future, see below.
 
 Other properties will be added to the `"text"` property in the future, to give greater control of what is processed (e.g. "content-search only" or "content-search and a PDF, using X as a cover-page").
 
@@ -74,7 +76,10 @@ The IIIF Manifest will serve as the input to text-services. This will be a full,
 
 We need to have the processed Manifest available as the source for text indexing but it won't have been made public yet, therefore IIIF-Presentation will need to store a version of the Manifest that can be used as input. We already have the concept of a "staged" Manifest. Currently, this is interim storage between API receipt and background-handler completion. If text indexing is required, the "staged" Manifest will either be the payload as received, or the Manifest generated from IIIF-CS NamedQuery.
 
-We will need some way to share the non-public resource with text-services. As we are using S3 for storage this can be done via IAM/bucket permissions or a presigned URL. There is nothing private in the Manifest, it's non-public because it hasn't been fully processed yet, not because the content is sensitive. If no `"pipeline"` property is provided then this is where processing would stop.
+We will need some way to share the non-public resource with text-services. As we are using S3 for storage this can be done via IAM/bucket permissions. There is nothing private in the Manifest, it's non-public because it hasn't been fully processed yet, not because the content is sensitive. If no `"pipeline"` property is provided then this is where processing would stop.
+
+> [!IMPORTANT]
+> We could share the S3 uri via a presigned URL but due to their timed nature this would be prohibitive for reingest scenarios.
 
 ### Alternative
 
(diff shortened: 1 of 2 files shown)


docs/rfcs/0007-text-services.md
```diff
@@ -0,0 +1,235 @@
...
+            "index": true
+        }
+    }
+}
+```
+
+Where:
+* `"pipeline"` is the top level containing property for indicating that there is some further processing to be done to the Manifest as a whole.
+* `"text"` indicates the type of pipeline. At this time this is the only accepted value.
+* `"index": true` signifies that we want to index any text associated with this Manifest.
```
JackLewis-digirati: is this the default value?

docs/rfcs/0007-text-services.md
```diff
@@ -0,0 +1,235 @@
...
+    PBH->>PBH: Record job
+    S3-->>TS: Read manifest
+    TS->>TS: Process
+    TS->>PBH: Job Completed
+    note right of PBH: Via message broker
+    PBH->>TS: GET text-augmented/{job_id}
+    TS-->>PBH: response
+    PBH->>PBH: Add search-services etc
+    PBH->>S3: Save Manifest to storage
+```
```
JackLewis-digirati, line 164: Think this is slightly incorrect - we also saved the manifest to staging in the protag API - so is the second step to staging storage an overwrite?

docs/rfcs/0007-text-services.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Text Operations
+
+Text-services allow consumer to specify which services/derivatives should be generated. A future implementation could be allow the consumer to control what is produced and exposed.
+
+This is currently modeled as a flags enum in text-services which may not be the best option to expose for consumers. We can assess the best approach when implementing.
+
+## Open Questions
+
+* If text indexing is requested but no text-bearing sources were found, should we record this anywhere? Something on the Manifest that can be surfaced in `X-IIIF-CS-Show-Extras` view? This is not necessarily an "error" but could be useful to record.
```
JackLewis-digirati: This would be nice to have imo

docs/rfcs/0007-text-services.md
```diff
@@ -0,0 +1,235 @@
...
+* Is the behaviour described in [Updates](#updates) correct when a Manifest with text-services is updated without a `"pipeline"` property?
```
JackLewis-digirati: this does work slightly different from adjuncts, so that could cause confusion?  Seems like there's a few states we want:

- update
- reuse (maybe?)
- remove (maybe?)
- don't advertise (potential use in the future)

docs/rfcs/0007-text-services.md
```diff
@@ -0,0 +1,235 @@
...
+    PA->>S3: GET manifest/{id}
+    S3-->>PA: manifest
+    PA-->>U: manifest
+    U->>TS: GET /search/autocomplete
+    TS-->>U: search response
+```
+
+## Updates
+
+If a Manifest that contains search-services is re-submitted without `pipeline.text.index: true` then the updated Manifest won't contain any generated services. They will persist in text-services and won't be deleted but they will no longer be advertised on the Manifest.
```
JackLewis-digirati: how would these be reenabled without causing the test-services pipeline to be run again? What do we do if somebody updates the source as well?

docs/rfcs/0007-text-services.md
donaldgray: We can discuss but I think maybe not to make the inclusion of text services opt-in only.

ie anything other than `pipeline.text.index: true` would mean no pipeline work is done.

docs/rfcs/0007-text-services.md
donaldgray: Added following to sequence diagram
* Show saving of Manifest to staging in Presentation API
* Highlight the second s3 staging save is an overwrite and added note that API view would see the _updated but not finished_ Manifest

docs/rfcs/0007-text-services.md
donaldgray: Agree - can discuss at refinement. Will leave conversation open until after discussion.

docs/rfcs/0007-text-services.md
donaldgray: Notes from call.

> this does work slightly different from adjuncts, so that could cause confusion?

Assets and adjuncts act differently. 
* Assets
  * Must be included in payload to be included in output.
  * Presentation won't delete assets
* Adjuncts
  * Don't need to be included in payload to be included in output (`null` adjuncts signifies "use what you've got").
  * Presentation can delete adjuncts of `[]` specified.

To we need to focus in MVP option first? Suggest
* If Manifest has `pipeline.text.index: true` then we will create a text-services job and add any output to Manifest.
* Else, no text-services job create. No output on Manifest. Nothing is deleted in text-services.

Need to discuss during refinement.

docs/rfcs/0007-text-services.md
donaldgray: Related to above response, marking as resolved for now. We can discuss during refinement.

docs/rfcs/0007-text-services.md
donaldgray: As discussed in refinement - approach will be to accept an enum as "pipeline" property. Initially only respecting 1 single value (equivalent of the boolean) and expanding in future.

docs/rfcs/0007-text-services.md
donaldgray: As per refinement, added details of `"pipelineHistory"` property that will expose this.

docs/rfcs/0007-text-services.md
donaldgray: See comment on updates + enum value.

tomcrane (approved):

JackLewis-digirati (approved):

PR #611, opened 2026-06-01
Allow null or empty manifest-level adjuncts to go through the serialiser correctly

### LINKED ISSUE #610: BUG: Adjuncts not preserved when further PUT omits adjuncts, or sends empty adjuncts array
Create Manifest with Manifest level adjunct(s) via POT
Update via PUT, send NO adjuncts value, or send empty adjuncts array

Expected response: origin al adjunct(s) persist and are included on the response
Actual response: No seeAlso/annotation/rendering block

JackLewis-digirati:
## What does this change?

Resolves #610 

This PR causes null or empty adjuncts to correctly go through the serialiser (in the case of null with adjuncts) and have the manifest-level `adjuncts` property be correctly filled out.  I.e.: with adjunct for `null` 

> [!Note]
> there's a slight variation from how `[]` works in painted resources if  `"adjuncts": []` is sent in a `canvasPainting` asset the API will respond with `"adjuncts": []`, whereas in a manifest-level adjunct, the response will be to not have an `adjuncts` property.  There is a bug tracking this on #612

### DIFF
--- a/src/IIIFPresentation/API.Tests/Features/Manifest/ManifestWriteServiceTests.cs
+++ b/src/IIIFPresentation/API.Tests/Features/Manifest/ManifestWriteServiceTests.cs
@@ -646,7 +646,7 @@ public class ManifestWriteServiceTests
                 ]
             });
 
-        const string stubManifestAdjunctId = "https://example.com/manifest-adjunct.xml";
+        const string stubManifestAdjunctId = "manifest-adjunct.xml";
         var stubAssetName = $"Manifest_{resourceId}";
         A.CallTo(() => dlcsClient.GetCustomerImages(Customer, A<string>._, A<CancellationToken>._))
             .Returns(Task.FromResult<IList<JObject>>(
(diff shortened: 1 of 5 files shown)


src/IIIFPresentation/API.Tests/Features/Manifest/ManifestWriteServiceTests.cs
```diff
@@ -645,8 +646,20 @@ public async Task Upsert_MergesStubAdjuncts_WithUserSetValues_WhenCanBeBuiltUpfr
                 ]
             });
 
+        const string stubManifestAdjunctId = "https://example.com/manifest-adjunct.xml";
```
donaldgray: I know this is just a test but this isn't a valid adjunct Id, it would be rejected by Protagonist for containing a `/`. Is this supposed to be the `"origin"`?

donaldgray (approved):

src/IIIFPresentation/API.Tests/Features/Manifest/ManifestWriteServiceTests.cs
JackLewis-digirati: this was unintentional - will swap it to make it clearer

PR #614, opened 2026-06-04
Stop duplicate manifests from being created/updated at the same time

### LINKED ISSUE #507: Stop manifests from being created (via PUT) when another manifest with the same id is being created

JackLewis-digirati:
## What does this change?

Resolves #507

Prevents duplicate DLCS batch submissions when the same manifest is submitted concurrently while a previous request is still mid-flight

This PR can be tested in 2 ways:

1. concurrent requests through the API (essentially execute a second call with the same values while waiting for the first to complete)
2. resending the same request while the manifest is waiting for processing in the background handler, but a response has been received from the API for the first request

The third scenario is very hard to force in practice as it requires 2 values to be generated with the same manifest id while both are processing through the API

### DIFF
--- a/docs/ADR/0003-locking.md
+++ b/docs/ADR/0003-locking.md
@@ -1,0 +1,65 @@
+# Locking
+* Status: decided
+* Author: Jack Lewis
+* Decider: Jack Lewis
+* Date: 2026-06-10
+
+## Context and problem statement
+
+When a manifest is ingested through the API there's the possibility of same manifest being "double-submitted" by a caller.  This causes major issues with processing as it can mean batches are double-submitted that means manifests are unable to be processed by the background handler.  [This has occasionally been seen in the live environment](https://github.com/dlcs/iiif-presentation/issues/507).  As such, some way of locking manifests from being processed twice needs to be implemented.
+
+There are a couple of scenarios where this issue can occur:
+
+1. Concurrent requests through the API (essentially execute a second call with the same values while waiting for the first to complete)
+2. Resending the same request while the manifest is waiting for processing in the background handler, but a response has been received from the API for the first request
+
+The second is fairly easy to solve by essentially checking if the manifest is still ingesting, and if it is respond with a `409 Conflict` response instead of accepting the manifest.  However, the first is more complex as it requires a solution to lock a manifest from being processed after the first attempt has been made.
+
+## Decision drivers
+
+- Make the locking function reusable le through a library or shared class
+- Locking needs to be controlled in a thread-safe manner
+
+## Considered options
+
+- Use a library like `AsyncKeyedLock`
+- Create an implementation of a locking library
+- Implement REDIS or a shared cache
+
+## Decision outcome
+
+Final decision was to write an implementation of a locking library so that it is internalized within iiif-presentation.  This allowed for greater flexibility than the other options, while being the most targeted approach for our needs.
+
+The specific implementation of locking can be found in the `ManifestLockManager` class.
+
+It was decided to roll out own due to the bug report.  While the bug didn't affect the current use-case and there was a workaround, the potential of adding a library with a known bug could have caused issues in the future if used in that specific case.  By contrast, REDIs was decided not to be used as it provides an additional burden of supporting infrastructure for something that isn't used yet.
+
+### Limitations
+
+It needs to be noted, that this locking functionality is in-memory and thus cannot be extended to multiple running instances of the API, unlike using REDIS.  This means that if the load balancer is set to share load across multiple instances of the API, it's possible for this to happen again due to API 2 not knowing that API 2 has locked the resource for processing.
+
+This functionality was decided to not be required for now as all production instances of the API currently only run a single instance.  However, this decision should be revisited if requirement changes and REDIS or similar should be implemented.
+
+## Pros and cons of the options
+
+### Locking library
+
+- Many users of the library provide better support
+- A number of additional functions that could be useful in the future
+- Updates are handled by another team
+- The specific library initially used had a major bug that meant locking didn't work in a specific case
+
+
+### Implementation
+
+- Focused only on the current needs of the project
+- Can be extended to support REDIS etc. in the future
+- flexible for needed future modifications
+- Must be maintained by the team (albeit minor)
+
+### REDIS
+
+- Allows for multiple instances of the API to run at the same time
+- Most widely used, so fully tested in production
+- Most complex implementation due to crossing system boundaries
+- Requires maintaining more infrastructure
--- a/docs/ADR/0003-locking.md
+++ b/docs/ADR/0003-locking.md
@@ -32,7 +32,7 @@ Final decision was to write an implementation of a locking library so that it is
 
 The specific implementation of locking can be found in the `ManifestLockManager` class.
 
-It was decided to roll out own due to the bug report.  While the bug didn't affect the current use-case and there was a workaround, the potential of adding a library with a known bug could have caused issues in the future if used in that specific case.  By contrast, REDIs was decided not to be used as it provides an additional burden of supporting infrastructure for something that isn't used yet.
+It was decided to roll our own due to the bug report.  While the bug didn't affect the current use-case and there was a workaround, the potential of adding a library with a known bug could have caused issues in the future if used in that specific case.  By contrast, REDIS was decided not to be used as it provides an additional burden of supporting infrastructure for something that isn't used yet.
 
 ### Limitations
 
--- a/docs/ADR/0003-locking.md
+++ b/docs/ADR/0003-locking.md
@@ -54,7 +54,7 @@ This functionality was decided to not be required for now as all production inst
 
 - Focused only on the current needs of the project
 - Can be extended to support REDIS etc. in the future
-- flexible for needed future modifications
+- Flexible for needed future modifications
 - Must be maintained by the team (albeit minor)
 
 ### REDIS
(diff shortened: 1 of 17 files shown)


JackLewis-digirati:
## How the duplicate-manifest guard works

### Two-layer protection

**Layer 1 — In-process lock (pre-DB)**
An `AsyncKeyedLocker<string>` (singleton, keyed on `{customerId}:{manifestId}`) is acquired at the start of `Upsert` before any DB reads or DLCS calls. If the lock is already held by another in-flight request, `409 Conflict` is returned immediately without touching DLCS. This closes the race window that existed between request entry and the first `SaveChangesAsync`, during which the existing `IsIngesting()` check could not detect a concurrent request.

**Layer 2 — DB batch state check (post-DB)**
The existing `IsIngesting()` guard is retained. It catches cases where a previous request has already persisted batches to the database (i.e. the manifest is legitimately ingesting), regardless of whether a lock is held.

**Layer 3 — DB constraint catch**
`DbUpdateException` on `pk_manifests` violation is caught and returned as `409 Conflict`, covering the rare concurrent-creation race on `POST`.

### Supporting changes
- `ModifyCollectionType.ManifestCurrentlyIngesting` (28) added
- `UpsertErrorHelper.ManifestCurrentlyIngesting<T>()` helper added
- `DbUpdateExceptionX.IsManifestPrimaryKeyViolation()` extension added
- `AddTestBatch` helper made configurable with an optional `BatchStatus` parameter (previously hardcoded to `Ingesting`, which would have triggered the new lock guard in existing tests)
- `AddTestManifest(..., ingested: false)` corrected to `ingested: true` in a test that expected normal write behaviour

src/IIIFPresentation/API.Tests/Features/Manifest/ManifestWriteServiceTests.cs
```diff
@@ -937,6 +942,69 @@ public async Task Upsert_ErrorCreatingManifest_WhenErrorWithPaintableAsset()
         ingestedManifest.WriteResult.Should().Be(WriteResult.BadRequest);
     }
     
+    [Fact]
+    public async Task Upsert_ReturnsConflict_WhenManifestHasIngestingAssetBatch()
+    {
+        // Arrange
+        var (slug, resourceId) = TestIdentifiers.SlugResource();
+
+        var dbManifest = await presentationContext.Manifests.AddTestManifest(resourceId, slug: slug, batchId: 9981);
```
donaldgray: I think this (and below) should use `TestIdentifiers.BatchId()`, rather than a hardcoded batch id.

src/IIIFPresentation/API.Tests/Integration/ModifyManifestUpdateTests.cs
```diff
@@ -40,6 +40,29 @@ public ModifyManifestUpdateTests(StorageFixture storageFixture, PresentationAppF
         storageFixture.DbFixture.CleanUp();
     }
     
+    [Fact]
+    public async Task PutFlatId_Update_Conflict_IfManifestIsCurrentlyIngesting()
+    {
+        // Arrange
+        var dbManifest = (await dbContext.Manifests.AddTestManifest(batchId: 999)).Entity;
```
donaldgray: As above, `TestIdentifiers.BatchId()`

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -93,13 +95,22 @@ public class ManifestWriteService(
...
     IPathRewriteParser pathRewriteParser,
+    AsyncKeyedLocker<string> manifestLocker,
     ILogger<ManifestWriteService> logger) : IManifestWrite
 {
     /// <summary>
     /// Create or update full manifest, using details provided in request object
     /// </summary>
     public async Task<PresUpdateResult> Upsert(UpsertManifestRequest request, CancellationToken cancellationToken)
     {
+        using var manifestLock = manifestLocker.LockOrNull($"M:{request.CustomerId}:{request.ManifestId}", 0);
```
donaldgray: Should we be using the `LockOrNullAsync()` method here, or was `LockOrNull()` chosen specifically?

donaldgray (commented):
Overall LGTM - one question is whether we should have an integration test that exercises the new logic via the API? Making a couple of consecutive requests and confirming one of them succeeds and one fails?

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
JackLewis-digirati: changed to use an in-built locking class

JackLewis-digirati:
I don't think there's a need for a specific integration test as `Upsert_ReturnsConflict_WhenManifestIsAlreadyBeingProcessed` effectively exercises the logic in a repeatable way (as well as various specific locking tests) without having to deal with potential race conditions etc.

docs/ADR/0003-locking.md
```diff
@@ -0,0 +1,65 @@
...
+* Date: 2026-06-10
+
+## Context and problem statement
+
+When a manifest is ingested through the API there's the possibility of same manifest being "double-submitted" by a caller.  This causes major issues with processing as it can mean batches are double-submitted that means manifests are unable to be processed by the background handler.  [This has occasionally been seen in the live environment](https://github.com/dlcs/iiif-presentation/issues/507).  As such, some way of locking manifests from being processed twice needs to be implemented.
+
+There are a couple of scenarios where this issue can occur:
+
+1. Concurrent requests through the API (essentially execute a second call with the same values while waiting for the first to complete)
+2. Resending the same request while the manifest is waiting for processing in the background handler, but a response has been received from the API for the first request
```
donaldgray, line 14: Is this (2) a possible scenario?

> but a response has been received from the API for the first request

If the response has returned the eTag should have changed, as the DB record will have been saved. The 202|Accepted response wouldn't contain an eTag but it would have changed in DB, so if we received another request with the old eTag it would result in a 412

docs/ADR/0003-locking.md
```diff
@@ -0,0 +1,65 @@
...
+- Create an implementation of a locking library
+- Implement REDIS or a shared cache
+
+## Decision outcome
+
+Final decision was to write an implementation of a locking library so that it is internalized within iiif-presentation.  This allowed for greater flexibility than the other options, while being the most targeted approach for our needs.
+
+The specific implementation of locking can be found in the `ManifestLockManager` class.
+
+It was decided to roll out own due to the bug report.  While the bug didn't affect the current use-case and there was a workaround, the potential of adding a library with a known bug could have caused issues in the future if used in that specific case.  By contrast, REDIs was decided not to be used as it provides an additional burden of supporting infrastructure for something that isn't used yet.
```
donaldgray: Small typo

```suggestion
It was decided to roll our own due to the bug report.  While the bug didn't affect the current use-case and there was a workaround, the potential of adding a library with a known bug could have caused issues in the future if used in that specific case.  By contrast, REDIS was decided not to be used as it provides an additional burden of supporting infrastructure for something that isn't used yet.
```

Link to bug mentioned?

docs/ADR/0003-locking.md
```diff
@@ -0,0 +1,65 @@
...
+- A number of additional functions that could be useful in the future
+- Updates are handled by another team
+- The specific library initially used had a major bug that meant locking didn't work in a specific case
+
+
+### Implementation
+
+- Focused only on the current needs of the project
+- Can be extended to support REDIS etc. in the future
+- flexible for needed future modifications
```
donaldgray:
```suggestion
- Flexible for needed future modifications
```

src/IIIFPresentation/API.Tests/Features/Manifest/ManifestWriteServiceTests.cs
```diff
@@ -937,6 +942,69 @@ public async Task Upsert_ErrorCreatingManifest_WhenErrorWithPaintableAsset()
...
+    }
+
+    [Fact]
+    public async Task Upsert_ReturnsConflict_WhenManifestHasIngestingAdjunctBatch()
+    {
+        // Arrange
+        var (slug, resourceId) = TestIdentifiers.SlugResource();
+
+        var dbManifest = await presentationContext.Manifests.AddTestManifest(resourceId, slug: slug);
+        await presentationContext.Batches.AddTestBatch(9982, dbManifest.Entity, DbDeliverableType.Adjunct);
```
donaldgray:
```suggestion
        await presentationContext.Batches.AddTestBatch(TestIdentifiers.BatchId(), dbManifest.Entity, DbDeliverableType.Adjunct);
```

src/IIIFPresentation/API/Infrastructure/ManifestLockManager.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+    IDisposable? TryAcquire(string key);
+}
+
+/// <summary>
+/// Backed by a <see cref="ConcurrentDictionary{TKey,TValue}"/> with ref-counted entries, so entries
+/// are removed once no thread holds or is waiting on a given key, keeping memory bounded to the set
+/// of currently-active keys.
+/// </summary>
+public sealed class ManifestLockManager : IManifestLockManager
```
donaldgray: Implementation LGTM - I don't think `Manifest` should be in the name though. We are currently using it for locking access to write manifests but there's nothing specifically "manifesty" about it, it could be used for anything that requires locking.

donaldgray (approved):

docs/ADR/0003-locking.md
JackLewis-digirati: in this specific scenario, you would have gotten back an etag not matched error - I've changed this to be a more specific "manifest currently ingesting" error

PR #623, opened 2026-06-15
Accept "pipeline" to generate SearchService2 on sync processed Manifests

### LINKED ISSUE #617: Manifests not requiring IIIF-CS ingestion
Handle Manifests that don't require any IIIF-CS ingestion that specify valid text `"pipeline"`.

These are IIIF-only payloads (ie no Assets or Adjuncts) OR those that contain only existing Assets.

Behaviour remains unchanged for those IIIF-only Manifests that don't specify valid text `"pipeline"`, however those that do:
* Create a text-builder job for Manifest, if a job already exists reprocess it.
* Return 202|Accepted without eTag value as there is further work to do.
  * This remains 200/201 if pipeline not specified.
* Manifest will return 202|Accepted in line with existing handling for this status code.
* When text-builder job is completed the Manifest will be updated to include generated IIIF Content Search services
  * Note that if no text bearing resources found there won't be a search service generated.
* Final Manifest updated and will return 200 w/ eTag in future.

> [!Note]
> Further acceptance criteria and description on the parent ticket

## Implementation notes

* Manifest is saved in "staging" area as is requires external content before completion.
* Text builder job should have `id` as defined in RFC.
* Text builder job `sourceUri` is the `s3://` url for staged Manifest, IAM policies will allow text builder to read.
* This will require an additional BackgroundHandler queue listener to complete.

JackLewis-digirati:
Resolves #617

## What does this change?

This PR enables pipelines in iiif-presentation as well as implementing a connection to the text services project.  

> [!Note]
> This PR now wraps database operations in a transaction so that failures to submit a manifest correctly rolls everything back.  Additionally, as text services requires S3 to be available from the moment of submission, the S3 is saved before pipeline submission.  This then needs to roll back by deleting the staging manifest from S3

## Database Migration

> [!NOTE]
> Details of migration. What does it change? Is it breaking/non-breaking?
>
> - **What it does:** adds the `pipeline_jobs` table that tracks jobs being submitted to text-services, while being extendable to other pipelines in the future
> - **Breaking Change?** No

## Configuration Changes

> [!NOTE]
>
> |---|---|---|---|---|
> | API | `TextServices:BuilderApiUri` | N | The location of the text services builder | `null` |
> | API | `TextServices:BuilderApiTimeoutSeconds` | N | Timeout for http requests to builder | 5 |
> | BackgroundHandler | `TextServices:SearchApiUri` | N | The location of the text services search | `null` |
> | BackgroundHandler | `TextServices:SearchApiTimeoutSeconds` | N | Timeout for http requests to builder | 30 |
> | BackgroundHandler | `AWS:SQS:TextJobQueueName` | N | The SQS queue holding competed text services jobs | `null` |


JackLewis-digirati:
## What this PR does

Implements the text-services pipeline integration (issue #617). When a manifest is created or updated with a `pipeline` property containing a `text` step, the API:

1. Creates a `PipelineJob` record (new DB table) tracking the job status
2. Submits the job to the text-services API via `TextServicesClient`
3. Returns `202 Accepted` while the job is in flight (the staged manifest is held in S3)

Once the text service finishes, a background SQS handler (`TextServiceJobCompletionMessageHandler`) picks up the completion message and:
- On success: reads the staged manifest from S3, merges any `SearchService2` entries from the text-augmented manifest into it, writes the final manifest to the public bucket, and marks the job as `Completed`
- On failure: marks the job as `Failed`, deletes the staged manifest from S3

### Key design decisions

**`PipelineJob` entity** follows the same `ManifestId?` / `CollectionId?` pattern as `Hierarchy` - nullable FKs with a check constraint ensuring exactly one is set (`num_nonnulls(manifest_id, collection_id) = 1`). This allows the same table to be reused for collection-level pipeline jobs in future without a schema change.

**Only `SearchService2` is merged** from the text-augmented manifest - no other service types. The Search 2 context URL (`http://iiif.io/api/search/2/context.json`) is also merged into the base manifest's `@context`.

**IManifestAugmentor** is a new interface introduced to abstract operations that modify Manifests. In addition to this the **ManifestS3Manager** class was refactored as it was doing 2 jobs - S3 saving and merging with IIIF-CS NQ. Now that we have an additional source of Manifest content in text-service it felt appropriate to split this down, rather than bloat **ManifestS3Manager**. It also allowed for consistent handling of "original" Manifest.

The `CreateManifest` and `UpsertManifest` handlers strip unknown pipeline values (via `RemoveInvalidPipelines()`) from payloads before passing them down to handling logic. This simplifies handling and avoids any troublesome logic where pipelines look present but they then turn out to be invalid.

Original manifests are saved if pipeline is specified, this follows logic introduced in #576

donaldgray (approved):

PR #625, opened 2026-07-01
Allow submitting pipelines following asset ingestion

### LINKED ISSUE #618: Manifest requiring IIIF-CS ingestion
Handle Manifest containing Asset or Adjuncts that require IIIF-CS ingestion (e.g. has adjuncts, new assets or asset with "reingest") specify valid text `"pipeline".

Behaviour remains unchanged for those Manifests that don't specify valid text `"pipeline"`, however those that do:
* Continue Return 202|Accepted without eTag value as there is further work to do.
* Manifest will return 202|Accepted in line with existing handling for this status code.
* When text-builder job is completed the Manifest will be updated to include generated IIIF Content Search services
  * Note that if no text bearing resources found there won't be a search service generated.
* Final Manifest updated and will return 200 w/ eTag in future.

> [!Note]
> Further acceptance criteria and description on the parent ticket

## Implementation notes

* Manifest is saved in "staging" area as is requires external content before completion.
* Upon completion of IIIF-CS batch completion, the background-handler will save the Manifest with content-resources in staging location and create text-builder job.
* Text builder job should have `id` as defined in RFC.
* Text builder job `sourceUri` is the `s3://` url for staged Manifest, IAM policies will allow text builder to read.
* Background handler added in (#617) will finish Manifest by decorating with search-services

JackLewis-digirati:
## What does this change?

Resolves #618

This PR extends text-services to work with assets that are being ingested through the background handler.  It works by submitting a pipeline job after the assets have been ingested.

## Configuration Changes

> [!NOTE]
>
> |---|---|---|---|---|
> | BackgroundHandler | `TextServices:BuilderApiUri` | N | The location of the text services builder | `null` |
> | BackgroundHandler | `TextServices:BuilderApiTimeoutSeconds` | N | Timeout for http requests to builder | 5 |

### DIFF
--- a/src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
@@ -539,7 +539,6 @@ public class ManifestWriteService(
                 ModifyCollectionType.CannotConnectToTextService, WriteResult.Error);
         }
 
-        job.Status = PipelineJobStatus.Waiting;
         await dbContext.SaveChangesAsync(cancellationToken);
         return null;
     }
--- a/src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
@@ -540,6 +540,8 @@ public class ManifestWriteService(
                 ModifyCollectionType.CannotConnectToTextService, WriteResult.Error);
         }
 
+        job.Status = PipelineJobStatus.Waiting;
+        await dbContext.SaveChangesAsync(cancellationToken);
         return null;
     }
 
@@ -555,7 +557,7 @@ public class ManifestWriteService(
                     ManifestId = dbManifest.Id,
                     JobType = PipelineJobType.TextService,
                     CustomerId = dbManifest.CustomerId,
-                    Status = PipelineJobStatus.Waiting,
+                    Status = PipelineJobStatus.NotSubmitted,
                     Config = pipelineItem.Config,
                     Created = DateTime.UtcNow
                 };
--- a/src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
@@ -99,7 +99,7 @@ public class ManifestWriteService(
     IDlcsManifestMerger dlcsManifestMerger,
     IPathRewriteParser pathRewriteParser,
     ILockManager manifestLockManager,
-    ITextBuilderClient textServicesClient,
+    ITextBuilderClient textBuilderClient,
     ILogger<ManifestWriteService> logger) : IManifestWrite
 {
     /// <summary>
@@ -487,15 +487,15 @@ public class ManifestWriteService(
 
         if (request.PresentationManifest.HasPipelineJob())
         {
-            var registerError = await RegisterPipelineJob(dbManifest, request.PresentationManifest.Pipeline!, cancellationToken);
-            if (registerError != null) return registerError;
+            var job = await RegisterPipelineJob(dbManifest, request.PresentationManifest.Pipeline!, cancellationToken);
+            if (job == null) return null;
 
             // Submission is deferred until DLCS batch completion when assets are still being ingested
             if (canBeBuiltUpfront)
             {
                 logger.LogDebug("Submitting pipeline job for manifest {ManifestId}",
                     dbManifest.Id);
-                return await SubmitPipelineJob(dbManifest, cancellationToken);
+                return await SubmitPipelineJob(dbManifest, job, cancellationToken);
             }
 
             logger.LogDebug("Deferring text-services submission for manifest {ManifestId} until DLCS batch completion",
@@ -509,8 +509,8 @@ public class ManifestWriteService(
     }
 
     // Persists the pipeline job record within the open transaction.
-    // Returns a failure result if no recognised pipeline type is found, otherwise null.
-    private async Task<PresUpdateResult?> RegisterPipelineJob(DbManifest dbManifest,
+    // Returns the created job, or null if no recognised pipeline type was found (nothing registered).
+    private async Task<PipelineJob?> RegisterPipelineJob(DbManifest dbManifest,
         List<PipelineItem> pipeline, CancellationToken cancellationToken)
     {
         var job = BuildPipelineJob(dbManifest, pipeline);
@@ -523,16 +523,16 @@ public class ManifestWriteService(
 
         (dbManifest.PipelineJobs ??= []).Add(job);
         await dbContext.SaveChangesAsync(cancellationToken);
-        return null;
+        return job;
     }
 
-    // Submits the most recently registered pipeline job to text-services.
+    // Submits the given pipeline job to text-services.
     // The HTTP call runs while the DB transaction is still open — the HttpClient should be configured with a short
     // timeout to bound this window. On failure the staged manifest is cleaned up so the caller can retry cleanly.
-    private async Task<PresUpdateResult?> SubmitPipelineJob(DbManifest dbManifest, CancellationToken cancellationToken)
+    private async Task<PresUpdateResult?> SubmitPipelineJob(DbManifest dbManifest, PipelineJob job,
+        CancellationToken cancellationToken)
     {
-        var job = dbManifest.PipelineJobs!.Last();
-        if (!await textServicesClient.UpsertJob(dbManifest, job, cancellationToken))
+        if (!await textBuilderClient.UpsertJob(dbManifest, job, cancellationToken))
         {
             logger.LogError("Failed to submit {JobType} pipeline job for manifest {ManifestId}", job.JobType, dbManifest.Id);
             await manifestStorageManager.DeleteStagedManifest(dbManifest);
--- a/src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
@@ -487,8 +487,20 @@ public class ManifestWriteService(
 
         if (request.PresentationManifest.HasPipelineJob())
         {
-            return await RegisterAndSubmitPipelineJobs(dbManifest, request.PresentationManifest.Pipeline!,
-                cancellationToken);
+            var registerError = await RegisterPipelineJob(dbManifest, request.PresentationManifest.Pipeline!, cancellationToken);
+            if (registerError != null) return registerError;
+
+            // Submission is deferred until DLCS batch completion when assets are still being ingested
+            if (canBeBuiltUpfront)
+            {
+                logger.LogDebug("Submitting pipeline job for manifest {ManifestId}",
+                    dbManifest.Id);
+                return await SubmitPipelineJob(dbManifest, cancellationToken);
+            }
+
+            logger.LogDebug("Deferring text-services submission for manifest {ManifestId} until DLCS batch completion",
+                dbManifest.Id);
+            return null;
         }
 
         // save changes called if there are no pipeline jobs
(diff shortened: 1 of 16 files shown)


JackLewis-digirati:
## Summary of changes

This PR wires the text-service pipeline into `BatchCompletionMessageHandler`, so manifests whose assets are still ingesting via DLCS when a text pipeline job (`pipeline: [{name: text, config: {action: Index}}]`) is requested get that job submitted once ingestion completes, rather than only supporting immediate/upfront submission.

**New `PipelineJobStatus.NotSubmitted`**
Splits the previous single `Waiting` status into two distinct states:
- `NotSubmitted` — job registered but not yet handed to text-services (new initial state, set by `ManifestWriteService.BuildPipelineJob`)
- `Waiting` — submitted to text-services, awaiting its completion notification

`Manifest.HasPendingPipelineJob()` now treats `NotSubmitted or Waiting` as outstanding work.

**Two submission paths**
- *Immediate* (`ManifestWriteService.SubmitPipelineJob`): used when assets don't need ingestion; submits synchronously and flips the job to `Waiting` after a successful `UpsertJob`.
- *Deferred* (`BatchCompletionMessageHandler.CompleteManifestFromStaging`): used when assets were still ingesting. Selects the `NotSubmitted` job for the manifest (ordered by `Created` descending, so the most recent resubmission wins if more than one exists), merges DLCS content into the staged manifest, then submits to text-services and flips the job to `Waiting` in-memory. This deliberately avoids an intermediate `SaveChangesAsync` — a failed submission leaves the job at `NotSubmitted` and rides the single `SaveChangesAsync` already at the end of `TryUpdateManifest`, so the whole batch-completion update (including the DLCS merge) is retried together on failure.

**Tests**
Added/updated coverage in `ManifestWriteServiceTests`, `BatchCompletionMessageHandlerTests` (including a regression test for correctly picking the most-recent job when a manifest has multiple `NotSubmitted` rows), `ManifestXTests`, and `PipelineHelperTests` for the new status and both submission paths.

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -487,23 +487,29 @@ await manifestStorageManager.SaveManifestInStorage(iiifManifest, dbManifest, ori
 
         if (request.PresentationManifest.HasPipelineJob())
         {
-            return await RegisterAndSubmitPipelineJobs(dbManifest, request.PresentationManifest.Pipeline!,
-                cancellationToken);
+            var job = await RegisterPipelineJob(dbManifest, request.PresentationManifest.Pipeline!, cancellationToken);
+            if (job == null) return null;
+
+            // Submission is deferred until DLCS batch completion when assets are still being ingested
```
donaldgray: This comment can be moved down, it's above the check for whether it's _not_ deferred.

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -516,15 +522,25 @@ await manifestStorageManager.SaveManifestInStorage(iiifManifest, dbManifest, ori
...
+    {
+        if (!await textBuilderClient.UpsertJob(dbManifest, job, cancellationToken))
         {
             logger.LogError("Failed to submit {JobType} pipeline job for manifest {ManifestId}", job.JobType, dbManifest.Id);
             await manifestStorageManager.DeleteStagedManifest(dbManifest);
             return PresUpdateResult.Failure("Error connecting to the text service",
                 ModifyCollectionType.CannotConnectToTextService, WriteResult.Error);
         }
 
+        job.Status = PipelineJobStatus.Waiting;
```
donaldgray: Since `textBuilderClient.UpsertJob()` takes the `job` as a parameter, should it mark the `job.Status = PipelineJobStatus.Waiting;`? Means that you couldn't call the method and forget to mark it as waiting, the client does it for you. Any additional future statuses could be set there too.

Tthinking about it - should we have a `FailedToSubmit` status? Would that be surfaced anywhere, or would it always be rolled back?

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -516,15 +522,25 @@ await manifestStorageManager.SaveManifestInStorage(iiifManifest, dbManifest, ori
...
+    // The HTTP call runs while the DB transaction is still open — the HttpClient should be configured with a short
+    // timeout to bound this window. On failure the staged manifest is cleaned up so the caller can retry cleanly.
+    private async Task<PresUpdateResult?> SubmitPipelineJob(DbManifest dbManifest, PipelineJob job,
+        CancellationToken cancellationToken)
+    {
+        if (!await textBuilderClient.UpsertJob(dbManifest, job, cancellationToken))
         {
             logger.LogError("Failed to submit {JobType} pipeline job for manifest {ManifestId}", job.JobType, dbManifest.Id);
             await manifestStorageManager.DeleteStagedManifest(dbManifest);
             return PresUpdateResult.Failure("Error connecting to the text service",
```
donaldgray: I know this is from previous PR but not sure if this error message could be tidied a little to make it more obvious as to what's not working. The consumer sends 

```json
{
    "pipeline": [ {
            "name": "text",
            "config": { "action": "Index" }}
    ]
}
```

but gets back `"Error connecting to the text service"` - should it say something more related to what they've sent? (`"Error running text pipeline"` or `"Error creating text pipeline job/runthing"` ??).

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -487,23 +487,29 @@ await manifestStorageManager.SaveManifestInStorage(iiifManifest, dbManifest, ori
 
         if (request.PresentationManifest.HasPipelineJob())
```
donaldgray, line 488: Should the handling of pipelines be in another class? That would mean that the `ManifestWriteService` knows that it needs to handle pipelines but not _how_ pipelines are handled - seems like this is maybe too many concerns in this class?

src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
```diff
@@ -89,17 +93,44 @@ private async Task<bool> TryUpdateManifest(BatchCompletionMessage batchCompletio
...
+        var pendingPipelineJob = dbManifest.PipelineJobs?
+            .Where(p => p.Status == PipelineJobStatus.NotSubmitted)
+            .OrderByDescending(p => p.Created)
+            .FirstOrDefault();
+        if (pendingPipelineJob != null)
+        {
+            Logger.LogInformation(
+                "Manifest {ManifestId} has pending text pipeline; saving merged content to staging and submitting job",
+                dbManifest.Id);
+            await manifestS3Manager.SaveManifestInStorage(merged, dbManifest, staged.Original, saveToStaging: true,
```
donaldgray: Do we need to pass `staged.Original` here? If left out, will it have the same outcome without an additional write happening?

src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
```diff
@@ -89,17 +93,44 @@ private async Task<bool> TryUpdateManifest(BatchCompletionMessage batchCompletio
         return true;
     }
 
-    // Read the staged manifest, merge in the DLCS content, save the final manifest (promoting any stored original
-    // payload) then remove the staging artifacts.
-    private async Task CompleteManifestFromStaging(Manifest dbManifest, CancellationToken cancellationToken)
+    // Read the staged manifest, merge in the DLCS content, then either hand off to the text pipeline or
+    // save the final manifest directly. Returns false if text-services submission fails (caller should retry).
+    private async Task<bool> CompleteManifestFromStaging(Manifest dbManifest, CancellationToken cancellationToken)
```
donaldgray: `TryComplete*` as it might not complete?

src/IIIFPresentation/Models/Database/General/PipelineJob.cs
```diff
@@ -40,7 +40,8 @@ public enum PipelineJobStatus
     Waiting = 0,
     Running = 1,
     Completed = 2,
-    Failed = 3
+    Failed = 3,
+    NotSubmitted = 4
```
donaldgray: Is this mapping the same as enum from text-service? I don't think it is but just checking to see if there's a  chance that text-service could add a new `4` that would have a different value and they'd then be out of sync?

src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
```diff
@@ -33,8 +35,10 @@ protected override async Task<bool> HandleMessage(BatchCompletionMessage batchCo
     private async Task<bool> TryUpdateManifest(BatchCompletionMessage batchCompletionMessage, int approximateReceiveCount, CancellationToken cancellationToken)
     {
         // Load batch the incoming message is referring to
-        var batch = await dbContext.Batches.Include(b => b.Manifest)
-            .ThenInclude(m => m.CanvasPaintings)
+        var batch = await dbContext.Batches
+            .Include(b => b.Manifest).ThenInclude(m => m.CanvasPaintings)
+            .Include(b => b.Manifest).ThenInclude(m => m.PipelineJobs)
```
donaldgray: Can we minimise the amount of data being pulled back, rather than all the history?

```suggestion
            .Include(b => b.Manifest).ThenInclude(m => m.PipelineJobs.Where(p => p.Status == PipelineJobStatus.NotSubmitted))
```

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -487,23 +487,29 @@ await manifestStorageManager.SaveManifestInStorage(iiifManifest, dbManifest, ori
...
-    // Submitting after SaveChangesAsync ensures DB state is consistent if the HTTP call fails and
-    // the transaction is rolled back by the caller.
-    // Trade-off: the HTTP call to text-services runs while the DB transaction is still open.
-    // This keeps rollback simple (no compensating transaction needed) at the cost of holding
-    // the transaction for the duration of the HTTP round-trip. The HttpClient should be
-    // configured with a short timeout to bound this window.
-    private async Task<PresUpdateResult?> RegisterAndSubmitPipelineJobs(DbManifest dbManifest,
+    // Persists the pipeline job record within the open transaction.
+    // Returns the created job, or null if no recognised pipeline type was found (nothing registered).
+    private async Task<PipelineJob?> RegisterPipelineJob(DbManifest dbManifest,
```
donaldgray: Small, I think the XML comment (`Persist`) explains what this is doing more than `Register`.
```suggestion
    private async Task<PipelineJob?> PersistPipelineJob(DbManifest dbManifest,
```

src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
JackLewis-digirati: there's no changes to `staged.Original`, so there would be nothing done  to it - added a comment explaining

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
JackLewis-digirati: Changed to `Error submitting text pipeline job`

src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
donaldgray: Wouldn't it still result in `s3.PutObject()` request happening?

JackLewis-digirati: sorry, should mention, I've explicitly passed `null` now, which will cause the `s3.PutObject()` to not happen

src/IIIFPresentation/Models/Database/General/PipelineJob.cs
JackLewis-digirati: What's passed is the name of the value, rather than the value itself, so it would continue to work.  However, you're correct that there could be a collision on the int if it updates - I'll change it to be `1000`

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
JackLewis-digirati: yeah, makes sense - also implemented `FailedToSubmit`

src/IIIFPresentation/BackgroundHandler/BatchCompletion/BatchCompletionMessageHandler.cs
```diff
@@ -89,17 +93,44 @@ private async Task<bool> TryUpdateManifest(BatchCompletionMessage batchCompletio
...
+            // originalPayload omitted: staged.Original was just read unchanged from OriginalStaging, so re-saving it
+            // here would be a no-op write. TextServiceJobCompletionMessageHandler re-reads it fresh when promoting
+            // the final manifest, so nothing downstream depends on it being re-written at this staging step.
+            await manifestS3Manager.SaveManifestInStorage(merged, dbManifest, originalPayload: null, saveToStaging: true,
+                cancellationToken);
+            if (!await pipelineJobService.SubmitPipelineJob(dbManifest, pendingPipelineJob, cancellationToken))
+            {
+                return false;
+            }
+            return true;
```
donaldgray: Can be simplified, we're not logging etc based on return value so can just return the result of `.SubmitPipelineJob` directly.
```suggestion
            return await pipelineJobService.SubmitPipelineJob(dbManifest, pendingPipelineJob, cancellationToken);
```

donaldgray (approved):
LGTM, small styling point. The PR title is still default.

PR #627, opened 2026-07-02
Query for all assets, instead of per asset

### LINKED ISSUE #592: Picked up by AI agent review
🔴 Critical Issues (must fix)

  1. N+1 query in CheckAssetsFromItemsExist — ManagedAssetResultFinder.cs
  - Current code calls dbContext.CanvasPaintings.Any() once per asset in a Where predicate
  - For manifests with many items-identified assets, this is dozens of separate DB roundtrips
  - Fix: materialise the lookup first, then use a set lookup in the filter

JackLewis-digirati:
## What does this change?

Resolves #592

This PR modifies a DB query in the `ManagedAssetResultFinder` so that assets are queried together in advance and then used in a lookup, avoiding potentially thousands of individual DB calls that would have been executed for each individual asset

### DIFF
--- a/src/IIIFPresentation/API/Features/Manifest/ManagedAssetResultFinder.cs
+++ b/src/IIIFPresentation/API/Features/Manifest/ManagedAssetResultFinder.cs
@@ -5,6 +5,7 @@ using Core.Exceptions;
 using Core.Helpers;
 using API.Settings;
 using DLCS.API;
+using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Options;
 using Models.API.Manifest;
 using Models.DLCS;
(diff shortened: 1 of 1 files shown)


JackLewis-digirati:
Fixes #592.

`CheckAssetsFromItemsExist` was calling `dbContext.CanvasPaintings.Any(cp => cp.AssetId == asset)` once per asset inside a `Where` predicate, causing an N+1 query — one DB roundtrip per items-identified asset in a manifest.

This replaces that with a batched, set-based lookup: asset IDs are chunked (500 at a time, matching the existing pattern in `FindAssetsInAnotherManifest`) and queried in a single `IN (...)`-style query per chunk, with results materialised into a `HashSet<AssetId>` for O(1) in-memory containment checks. Behaviour is unchanged; existing `ManagedAssetResultFinderTests` all pass.

donaldgray (approved):

PR #629, opened 2026-07-02
Add PublishReadyToRun to API

### LINKED ISSUE #328: API slow startup
dotnet can be slow to serve first requests when just started up. However, it's been noted that when deployed to dev environment the API is _really_ slow to serve the first request. Often ~16secs, where after that requests are served in ~100ms, sometimes less.

The 16s value is for `root` collection request so this should be quick as it's all DB driven. We should identify what is causing the slowdown (is a _lot_ of stuff cached on first request? Is it dependency resolution?), whether it's something about dev environment, what we can do to speed it up.

JackLewis-digirati:
## What does this change?

Resolves #328 

This PR adds `PublishReadyToRun` to the API csproj to improve first request times.

> [!Note]
> There's additional follow-ups that could be considered for implementation in the comment below, along with discussion of options presented in the ticket, but discarded

### DIFF
--- a/src/IIIFPresentation/API/API.csproj
+++ b/src/IIIFPresentation/API/API.csproj
@@ -1,11 +1,13 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
-
     <PropertyGroup>
         <TargetFramework>net8.0</TargetFramework>
         <Nullable>enable</Nullable>
         <ImplicitUsings>enable</ImplicitUsings>
         <DockerDefaultTargetOS>Linux</DockerDefaultTargetOS>
         <RootNamespace>API</RootNamespace>
+        <RuntimeIdentifier>linux-x64</RuntimeIdentifier>
+        <SelfContained>false</SelfContained>
+        <PublishReadyToRun>true</PublishReadyToRun>
     </PropertyGroup>
 
     <ItemGroup>


JackLewis-digirati:
## Investigation notes on cold-start options considered

Verified the `PublishReadyToRun` change actually takes effect (it silently didn't, before adding `RuntimeIdentifier`/`SelfContained` — R2R needs an explicit RID or it infers one from the build host rather than the deploy target). Confirmed by building the real `Dockerfile.API` image and finding the `RTR\0` ReadyToRun signature embedded in `API.dll` inside the container.

Also looked into a few other angles raised in the original issue / my earlier comment:

**EF Core compiled models — not viable.** Ran `dotnet ef dbcontext optimize` against `PresentationContext` directly. It fails:
```
The entity type 'CanvasPainting' has a query filter configured. Compiled model can't be generated, because query filters are not supported.
```
Our multi-tenant `HasQueryFilter` (applied to every `ICustomerEntity` in `ApplyGlobalFilters`) is fundamentally incompatible with EF Core compiled models in EF Core 8 — this isn't a config issue, it's a hard unsupported combination. Not pursuing unless we're willing to redesign customer-scoping away from global query filters, which is a much bigger change than this PR's scope.

**FluentValidation / MediatR assembly scanning** — real reflection cost at startup, but the API only has 4 validators and ~15 MediatR request/handler types, so the scan itself is cheap here; the cost scales with assembly size not match count, so the saving from hand-rolling explicit registration is marginal and trades away the safety net of auto-registration (easy to forget to register a new handler). Not pursuing.

**Possible follow-ups not implemented in this PR (flagging for visibility):**
- `RunMigrations: true` is set in `appsettings.Development.json`, which is force-copied to publish output. If the deployed dev environment runs with `ASPNETCORE_ENVIRONMENT=Development`, the API runs `TryRunMigrations` on every boot — building the migration model from 69 migration files plus a DB round-trip, before the app is ready to serve. Worth checking whether that's actually the env dev runs under, since this would land squarely on "why is dev specifically slow" from the original issue thread.
  - This is not an issue in prod
- Server GC (`System.GC.Server: true`, the ASP.NET Core default) pre-allocates a heap per detected CPU core at startup, which is disproportionately expensive on a CPU-constrained container and can over-allocate if cgroup CPU-limit detection is off. Testing with `DOTNET_gcServer=0` set as an env var on the dev deployment (no code change needed) would validate/rule this out without affecting other environments.
- A startup warm-up request (hit the app in-process once after `app.Build()`/before accepting real traffic) would shift the JIT/EF-model-build cost from the first real user request into invisible startup time — the most direct fix for the exact symptom in the issue, but adds startup-path complexity so left out of this PR.

JackLewis-digirati:
## Measured impact of the R2R fix

Built the actual images from git history (before this PR vs. with the `RuntimeIdentifier`/`SelfContained`/`PublishReadyToRun` fix), ran each fresh against the same already-migrated Postgres database, and timed the first real HTTP request (`GET /1`, the root collection route) — 3 runs each, fresh container per run:

| | Baseline (no R2R) | Current (R2R fix) |
|---|---|---|
| First-request handling time | 340, 342, 340 ms | 169, 193, 166 ms |
| **Average** | **~341ms** | **~176ms** |
| Steady-state (2nd request) | 5–9ms | 5–7ms |

**~48% reduction in first-request latency**, consistent across all three runs on each side.

**Caveat:** this was measured on a fast, unconstrained dev machine — total cold-start-to-first-response was ~1.8s (baseline) vs ~1.2s (fixed), nowhere near the 16s reported in the issue. JIT cost scales inversely with available CPU, and the original issue notes dev is "underpowered" relative to prod, so the *absolute* seconds saved in the real dev environment are almost certainly larger than what I measured here — I can't quantify that without testing in the actual environment. The *relative* ~50% cut in JIT-bound first-request work is the number I'd trust to carry over as-is.


### Tradeoffs

The image has increased in size from 280MB to 350MB ~25%.  While larger, the increase is still acceptable for the increase in first request time

donaldgray (approved):
> RunMigrations: true is set in appsettings.Development.json, which is force-copied to publish output. If the deployed dev environment runs with ASPNETCORE_ENVIRONMENT=Development, the API runs TryRunMigrations on every boot

Flagging that this isn't possible - `appsettings.Development.json` isn't in the main docker-images, this could only be a possibility when running locally-built docker images. If it's of concern we should add to `.dockerignore`

PR #630, opened 2026-07-06
Search Across MVP RFC

donaldgray:
## What does this change?

Add new RFC looking at how we can support a basic search-across function using stored labels.

### DIFF
--- a/docs/rfcs/0008-search-across-mvp.md
+++ b/docs/rfcs/0008-search-across-mvp.md
@@ -1,0 +1,196 @@
+# Search Across MVP
+
+Manifests can utilise [Text Services](0007-text-services.md) and pipelines to allow searching within an individual Manifest via the [IIIF Content Search API](https://iiif.io/api/search/2.0/). This document
+outlines how we can support a customer/tenant searching across multiple resources.
+
+> [!NOTE]
+> This document outlines a simple MVP approach initially - this will be further expanded and enhanced in the future.
+
+## Outline
+
+A high level overview of the approach is:
+
+* Search will be run against the root collection, available to authenticated users only.
+* The capability will be advertised as a new `"service"` on the root collection.
+* It will search all label values of every resource in the root collection, regardless of nesting depth.
+* A synthetic collection will be returned supporting paging.
+
+### Search Endpoint
+
+> * Search will be run against the root collection.
+
+Search will be available at a new endpoint: `GET /{customerId}/collections/{collectionId}/search`. Search criteria and paging controls will be specified by query parameters.
+
+For the initial MVP:
+* The only accepted `{collectionId}` will be "root", searching across _all_ Manifests and Collections in the root collection - regardless of depth.
+* `label` query parameter will contain the search criteria.
+
+In addition to the above we can support standard query parameters available for all storage collections, namely `page`, `pageSize`, `orderBy`, `orderByDescending`
+
+Example queries
+* `https://presentation-api.example/99/collections/root/search?label=medicine`
+* `https://presentation-api.example/99/collections/root/search?label=19th+century+novels&pageSize=10&page=2`
+
+### New Search Service 
+
+> * The capability will be advertised as a new `"service"` on the root collection.
+
+Any storage-collections that support search-across will have a `"service"` block added. The custom IIIF-Presentation `@context` will define this, e.g.
+
+```jsonc
+{
+    "@context": [
+        "http://tbc.org/iiif-repository/1/context.json",
+        "http://iiif.io/api/presentation/3/context.json"
+    ],
+    "id": "https://presentation-api.example/99/collections/root",
+    "type": "Collection",
+    "totals": {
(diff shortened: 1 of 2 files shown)


docs/rfcs/0008-search-across-mvp.md
```diff
@@ -0,0 +1,196 @@
...
+### New Search Service 
+
+> * The capability will be advertised as a new `"service"` on the root collection.
+
+Any storage-collections that support search-across will have a `"service"` block added. The custom IIIF-Presentation `@context` will define this, e.g.
+
+```jsonc
+{
+    "@context": [
+        "http://tbc.org/iiif-repository/1/context.json",
```
tomcrane, line 43: (more of a note to self... should do something about this! - not a blocker for this work).

docs/rfcs/0008-search-across-mvp.md
```diff
@@ -0,0 +1,196 @@
...
+> The former is `jsonb`, the latter `text` - they should both be the same type (`jsonb`).
+
+> [!NOTE]
+> The below are query examples and might not prove to be the most efficient.
+
+##### Simple
+
+An initial, simple implementation of search can use basic `ILIKE` across all values per label. If search terms are multi-word they are split into individual tokens so word order doesn't matter, e.g.
+
+```sql
```
tomcrane, line 96: Would you construct this sql programmatically from the query param? Or use a library.

docs/rfcs/0008-search-across-mvp.md
```diff
@@ -0,0 +1,196 @@
...
+LIMIT 100; --?pageSize
+```
+
+Downside of this approach is that indexes won't be used so performance will need to be monitored. A standard jsonb GIN index would not accelerate `ILIKE` search as it only supports exact key/value queries.
+
+#### Advanced
+
+A more advanced approach would be to use trigram word search [`pg_trgm`](https://www.postgresql.org/docs/14/pgtrgm.html), or Levenshtein [`fuzzystrmatch`](https://www.postgresql.org/docs/14/fuzzystrmatch.html) matching, both can have different weightings to improve search results. The former seems like it would be a better fit as it would handle missing/reordered tokens better. This, in addition to a [GIN index](https://www.postgresql.org/docs/14/gin.html), should provide more efficient querying.
+
+Both `pg_trgm` and `fuzzystrmatch` work on `text`, rather than `jsonb` data. To handle `jsonb` we would need to pre-process, one option would be to create a new generated column containing the flattened values. This new column can be indexed to allow for efficient searching, it would also improve the above `ILIKE` queries if we wanted to keep this same approach.
```
tomcrane, line 120: This approach makes sense for `label` but might get unwieldy if we have to support searching across arbitrary `metadata` values.

docs/rfcs/0008-search-across-mvp.md
```diff
@@ -0,0 +1,196 @@
...
+
+We may also want to look at [`unaccent`](https://www.postgresql.org/docs/14/unaccent.html) extension to handle labels containing accented characters.
```
tomcrane, line 122: I think this would be necessary, if we do the generated column the values in that generated column should be normalised/unaccented (removed diacritics) and then the search term is also normalised before searching.

tomcrane (approved):
So we're going to go with the ILIKE approach initially?

docs/rfcs/0008-search-across-mvp.md
donaldgray: I'm not sure whether npgsql will natively support such a query but worth investigating. My assumption is that you'd need to do so with EFs `FromSql` or `FromSqlRaw`, so that you get the sql injection protection but can construct the query as you want. 

I'm also not sure how you'd marshal that into a concrete type, `Collection` and `Manifest` share a common interface in `IHierarchyResource` but we might be better with a condensed result set that only contains what we need to render search results.

docs/rfcs/0008-search-across-mvp.md
donaldgray: Agree, I think that we'd want to re-look at the overall approach when we move beyond basic `label` search or it could get unwieldy. For Manifest/Collection `"metadata"` searches we'd need to be indexing the IIIF content anyway, so we could include the `label` as part of that indexing (whether that's into pg or something else).

At that point the `label` stored in the DB goes back to only being used for rendering storage-collections, rather than for searching.

docs/rfcs/0008-search-across-mvp.md
donaldgray: Yeah - that might fall under same category as the previous comment where we need to have a specific search-index for handling. I'll add something around this and your other comment to RFC

donaldgray:
> So we're going to go with the ILIKE approach initially?

Yeah, I think that's simplest in the short term. We only need basic `ilike` searching to meet the immediate use case, the volume should be relatively low as it's effectively a backend admin search, rather than public facing.

PR #633, opened 2026-07-07
Add previous pipelines to the authenticated manifest

### LINKED ISSUE #620: Return details of previous pipelines
Add a `"finishedPipelines"` property to Manifests, this will allow API consumers to see any historical pipelines run on the Manifest, including errors and warnings.

Each "finishedPipelines" will be in format

```jsonc
{
  "name": "text",
  "config": {
      "action": "Index"
   },
  "status": "Completed|Waiting|Error|Etc",
  "error": "Something went wrong",
  "warning": "No text resources found",
  "created": "2026-07-30 08:00:00",
  "finished": "2026-07-30 08:03:12"
}
```

## ACs

* `"finishedPipelines"` only returned for requests with x-iiif-cs-show-extras header and valid auth. 
* `"finishedPipelines"` will be missing/null if there have never been any pipelines.
* `"finishedPipelines"` will be populated with details of previous (ie not currently in-flight) pipelines.
  * Completed/failed pipelines only
* `"finishedPipelines"` will be limited to the last X most recent pipelines.
  * X should pick a reasonable default (20?) and be controlled via configuration.
* Current (ie in-flight) pipeline details will still be returned in `"pipeline"` property, as per parent ticket.
  * This property will only appear while Manifest is returning 202.
  * `"pipeline"` echos the above format (for now it'll only contain `"status"` and `"created"`).

-------

## Implementation note

Add a `invocationCount` to each pipelineJob. This will be allow us to tie up "completed" notifications with the exact pipelineJob record. See https://github.com/dlcs/text-services/issues/57

JackLewis-digirati:
## What does this change?

Resolves #620

This PR implements the previous pipelines property to tack completed pipelines.  Additionally, this uses the new text services `InvocationCount` property to help track matching invocations.

Additionally, fixes an issue where the API built to linux from Rider, stopping it from running on Windows

## Database Migration

> [!NOTE]
> Migration added to support `InvocationCount`
>
> - **Migration name:** `AddInvocationCountToPipelineJob`
> - **What it does:** adds `invocation_count` (int, default 1), and adds a unique index on `(customer_id, manifest_id, collection_id, job_type, invocation_count)`
> - **Breaking Change?** No

## Configuration Changes

> [!NOTE]
>
> |---|---|---|---|---|
> | API | `FinishedPipelinesLimit` | N | the maximum amount of finished pipelines returned to the user | `20` |

## Text Services Dependency

> [!NOTE]
> This PR requires a version of the text services deployed that has `InvocationCount`


JackLewis-digirati:
## Summary

Implements #620 — adds a `finishedPipelines` property to Manifests so API consumers can see the history of previous text-service pipeline runs (status, error, timestamps), gated behind the existing `X-IIIF-CS-Show-Extras` header + auth convention.

## Key changes

**API surface**
- `PresentationManifest.FinishedPipelines`: history of terminal (`Completed`/`Failed`/`FailedToSubmit`) pipeline jobs, newest first, capped at `ApiSettings.FinishedPipelinesLimit` (default 20).
- `PresentationManifest.Pipeline` behaviour change: now only shows a job if it's still in-flight (`Waiting`/`NotSubmitted`/`Running`). Previously it always showed the most-recently-created job regardless of status, so a completed job used to linger in `pipeline` on a 200 response — it now moves to `finishedPipelines` instead.
- `PipelineItem` gained `error`, `created`, `finished` fields to match the shape in the issue.

**InvocationCount correlation (from the issue's implementation note, tied to `dlcs/text-services#57`/PR#56)**
- `TextServiceJobCompletionMessageHandler` now matches a completion notification to the specific `PipelineJob` row by `InvocationCount`, falling back to "newest wins" only for rows created before this migration. This fixes a real latent bug: with multiple historical jobs per manifest (each resubmission creates a new row for history), "newest wins" could complete the wrong row.
- `TextBuilderClient` also now persists the actual error text-services returns (`JobResponse.Errors`) into `PipelineJob.Error` on a failed submission — previously a `FailedToSubmit` job's `Error` column was always `null`.

**Migration:** `AddInvocationCountToPipelineJob` — adds `invocation_count` (int, default 1), backfills pre-existing rows with sequential per-resource values (rather than a flat default) to avoid ties, and adds a unique index on `(customer_id, manifest_id, collection_id, job_type, invocation_count)` filtered to exclude `NotSubmitted`/`FailedToSubmit` rows (which never get a real value from text-services). Non-breaking — additive column + backfill.

## Test plan
- Full solution test suite passes (1291 tests), including `Repository.Tests` which applies the new migration (backfill SQL + unique index) against a real Postgres instance.
- New/updated coverage: `ManifestConverterTests`, `GetManifestTests` (integration), `PipelineJobServiceTests`, `TextBuilderClientTests`, `TextServiceJobCompletionMessageHandlerTests` (including a regression test proving invocation-count correlation picks the right row when it isn't the newest).

src/IIIFPresentation/API/Converters/ManifestConverter.cs
```diff
@@ -33,7 +33,7 @@ public static class ManifestConverter
     /// </param>
     public static PresentationManifest SetGeneratedFields(this PresentationManifest iiifManifest,
         Manifest dbManifest, IPathGenerator pathGenerator, SettingsBasedPathGenerator settingsBasedPathGenerator, Dictionary<string, JObject>? assets = null,
-        Func<Manifest, Hierarchy>? hierarchyFactory = null)
+        Func<Manifest, Hierarchy>? hierarchyFactory = null, int finishedPipelinesLimit = 20)
```
donaldgray, line 36: Not required for this but this signature is starting to indicate that this shouldn't be a static method.

src/IIIFPresentation/Models/Database/General/PipelineJob.cs
```diff
@@ -33,6 +33,12 @@ public class PipelineJob : ICustomerEntity
     public DateTime? Finished { get; set; }
 
     public PipelineConfig? Config { get; set; }
+
+    /// <summary>
+    /// Number of times text-services has been asked to (re)process this job. Mirrors text-services'
```
donaldgray: I don't think this comment should be text-services specific, it's currently used for text-services but the intention is for it to contain general pipeline details, not only text-services.

src/IIIFPresentation/Models/Database/General/PipelineJob.cs
```diff
@@ -33,6 +33,12 @@ public class PipelineJob : ICustomerEntity
...
+    /// own `InvocationCount`, so a completion notification can be matched back to this exact record.
+    /// </summary>
+    public int InvocationCount { get; set; } = 1;
```
donaldgray: IMO `int InvocationCount` ties this to text-services too much, we're modelling this table against a concrete implementation rather than a general one that could handle variations. What we want is some way of identifying a particular run of a particular pipeline - so that when we get details of individual pipeline invocations we can tie them directly to this DB row, on something more than just 'latest' or date.

TextServices models this as `InvocationCount`. Initially we discussed modelling this as a more complex Id but that would have meant modelling full job-history in text-service so we opted for a simple invocation count - other services could differ.

* TextService returns `InvocationCount: 1`, `InvocationCount: 2` etc
* OCR could be `AttemptId: fbcf50ea-c6cb-4c7f-8728-d6f4c25559d3`
* HTR could be `Invocation: ab12324-tre`
* EntityExtraction could be `Run: 85409835434`
* etc

I think something like `string InvocationId` will be more flexible. That way we could handle different types

src/IIIFPresentation/Repository/Migrations/20260707150303_AddInvocationCountToPipelineJob.cs
```diff
@@ -0,0 +1,58 @@
...
+                ) AS ranked
+                WHERE pj.id = ranked.id;
+                """);
+
+            migrationBuilder.CreateIndex(
+                name: "ix_pipeline_jobs_customer_id_manifest_id_collection_id_job_typ",
+                table: "pipeline_jobs",
+                columns: new[] { "customer_id", "manifest_id", "collection_id", "job_type", "invocation_count" },
+                unique: true,
+                filter: "status NOT IN ('NotSubmitted', 'FailedToSubmit')");
```
donaldgray: Could this be `invocation_count not null`, or something that won't involve changing over time? This current approach could mean that more statuses would need to be included as they are added.

src/IIIFPresentation/Models/API/Manifest/Pipeline.cs
```diff
@@ -51,6 +59,10 @@ public static bool HasPipelineJob(this PresentationManifest manifest) =>
     {
         Name = job.JobType == PipelineJobType.TextService ? TextPipeline.Name : "unknown",
         Config = job.Config,
-        Status = job.Status.ToString()
+        Status = job.Status.ToString(),
+        Error = job.Error,
+        Warning = job.Status == PipelineJobStatus.CompletedNoOperation ? TextPipeline.NoTextFoundWarning : null,
```
donaldgray, line 59: Is there any value in modelling `Warning` as a column we store in the DB, rather than deriving purely from the status?

src/IIIFPresentation/Models/API/Manifest/Pipeline.cs
```diff
@@ -12,6 +12,13 @@ public class PipelineItem
     public PipelineConfig? Config { get; set; }
     [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
     public string? Status { get; set; }
+    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+    public string? Error { get; set; }
+    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+    public string? Warning { get; set; }
+    public DateTime Created { get; set; }
+    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
```
donaldgray: I don't think these are required as there's defaults setup in Program.cs, they're not used elsewhere so assume  they're not required.

```cs
builder.Services.AddControllers().AddNewtonsoftJson(options =>
{
    options.SerializerSettings.NullValueHandling = NullValueHandling.Ignore;
});
```

src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
```diff
@@ -52,6 +52,11 @@ public class PresentationManifest : IIIF.Presentation.V3.Manifest, IPresentation
     [JsonIgnore] public bool CurrentlyIngesting { get; set; }
 
     [JsonProperty(Order = 16)] public List<PipelineItem>? Pipeline { get; set; }
+
+    /// <summary>
+    /// History of previous (completed/failed) pipeline runs for this manifest, most recent first.
+    /// </summary>
+    [JsonProperty(Order = 17)] public List<PipelineItem>? FinishedPipelines { get; set; }
```
donaldgray, line 59: Consider making this an `IOrderedEnumerable` , this makes it obvious that the ordering is important (looking at "most recent first" xml comment)

src/IIIFPresentation/BackgroundHandler/TextCompletion/TextServiceJobCompletionMessageHandler.cs
```diff
@@ -38,18 +38,29 @@ protected override async Task<bool> HandleMessage(TextServiceJobCompletionMessag
...
+        var candidateJobs = await dbContext.PipelineJobs
             .Where(p => p.ManifestId == jobId.ResourceId && p.JobType == PipelineJobType.TextService)
             .Include(p => p.Manifest)
             .OrderByDescending(p => p.Created)
-            .FirstOrDefaultAsync(cancellationToken);
+            .ToListAsync(cancellationToken);
+
+        // Match the specific invocation this notification refers to; fall back to "newest wins" for any
+        // job created before InvocationCount existed, or if text-services ever omits it.
+        var pipelineJob = candidateJobs.FirstOrDefault(p => p.InvocationCount == completionMessage.InvocationCount);
```
donaldgray: Wondering if we can do this in the DB, let it do the work? It might not be worth the effort, given the number of records that are likely to exist, just a thought.

Also - what is the reasoning for handling the case where we get a completed notification for an Invocation that we don't know about? Is this a possible scenario - what would happen to end up there?

src/IIIFPresentation/Models/API/Manifest/PresentationManifest.cs
JackLewis-digirati: I think there might be issues with doing this - if somebody `GET` then `PUT` the manifest to update it, `previousPipleines` would be smoothly deserialized being a `List`, but if it's an `IOrderedEnumerable`, then it would fail because there's no default implementation of an `IOrderedEnumerable` in Newtonsoft

I could write up a converter to allow it, but feels like more work than it's worth?  Alternatively, it might be better to set it as non-writable on deserialize.  For now, I've removed the further comment around most recent first

src/IIIFPresentation/Models/API/Manifest/Pipeline.cs
JackLewis-digirati: I don't think so? This might change in the future, but for now its coming just from Status - this *could* change in the future, but that would likely be as part of a large PR with other changes - though feel it's fairly unlikely

src/IIIFPresentation/BackgroundHandler/TextCompletion/TextServiceJobCompletionMessageHandler.cs
JackLewis-digirati: Moved to grab the record fro the DB, rather than load all and find the correct one.  For the question, horribly unlikely - either something being inflight during the migration for this, or if the text-services job somehow manages to complete before the API calls to save changes in the DB from job submission (which tracks the invocation count)

src/IIIFPresentation/BackgroundHandler/TextCompletion/TextServiceJobCompletionMessageHandler.cs
```diff
@@ -38,18 +38,41 @@ protected override async Task<bool> HandleMessage(TextServiceJobCompletionMessag
...
         var pipelineJob = await dbContext.PipelineJobs
-            .Where(p => p.ManifestId == jobId.ResourceId && p.JobType == PipelineJobType.TextService)
+            .Where(p => p.ManifestId == jobId.ResourceId && p.JobType == PipelineJobType.TextService
+                        && p.InvocationId == invocationId)
             .Include(p => p.Manifest)
-            .OrderByDescending(p => p.Created)
             .FirstOrDefaultAsync(cancellationToken);
 
+        if (pipelineJob == null)
+        {
```
donaldgray, line 49: Might be worth adding some logging to show if we need to do this, if it's happening often it's highlighting that something isn't quite right.

donaldgray (approved):

PR #634, opened 2026-07-08
Delete text-services job when manifest is deleted

### LINKED ISSUE #626: Delete text-builder job when Manifest deleted
When Manifest is deleted and it has a text-builder job, ensure the text-builder job is also deleted.

This can happen immediately or after a slight delay.

JackLewis-digirati:
## What does this change?

Resolves #626

Allows the text-services job to be deleted when a manifest is also deleted

### DIFF
--- a/src/IIIFPresentation/Services/TextServices/PipelineJobService.cs
+++ b/src/IIIFPresentation/Services/TextServices/PipelineJobService.cs
@@ -1,3 +1,4 @@
+using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Logging;
 using Models.API.Manifest;
 using Models.Database.General;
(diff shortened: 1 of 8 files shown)


src/IIIFPresentation/Services/TextServices/PipelineJobService.cs
```diff
@@ -40,6 +41,23 @@ public async Task<bool> SubmitPipelineJob(DbManifest dbManifest, PipelineJob job
         return false;
     }
 
+    public async Task DeletePipelineJob(DbManifest dbManifest, CancellationToken cancellationToken)
+    {
+        var job = await dbContext.PipelineJobs
+            .FirstOrDefaultAsync(p => p.ManifestId == dbManifest.Id && p.CustomerId == dbManifest.CustomerId,
```
donaldgray: Global filter should handle Customer but we'll want to filter by Text type - or handle that further down so that when we introduce new pipelines we don't try and delete those.

```suggestion
            .FirstOrDefaultAsync(p => p.ManifestId == dbManifest.Id && p.JobType == PipelineJobType.TextService,
```

src/IIIFPresentation/Services/TextServices/PipelineJobService.cs
```diff
@@ -40,6 +41,23 @@ public async Task<bool> SubmitPipelineJob(DbManifest dbManifest, PipelineJob job
...
+                cancellationToken);
+
+        // NotSubmitted/FailedToSubmit jobs never reached text-services, so there's nothing to delete there
+        if (job == null || job.Status is PipelineJobStatus.NotSubmitted or PipelineJobStatus.FailedToSubmit) return;
```
donaldgray: This isn't necessarily true, is it? The above query is only pulling the first pipelineJob but not specifying any explicit ordering. If that wasn't submitted, or failed to submit, could there be other jobs that did successfully submit (e.g. in the past) and would therefore need to be deleted?

Can the above query instead be an `.Any()` - if there's every been a text job that was finished, try to delete it

donaldgray (approved):

Issue #632, opened 2026-07-06
Support ordering in search across


donaldgray:
Tickets follows on from #631 to add support for `orderBy` and `orderByDescending` query parameters.

We should use the same 3 properties that are already supported: `id`, `slug` and `created`.

Issue #631, opened 2026-07-06
Search Across (MVP)


donaldgray:
#616 added support for search _within_ a Manifest, this ticket is to address searching _across_ multiple Manifests and Collections. See [RFC 0008](https://github.com/dlcs/iiif-presentation/blob/develop/docs/rfcs/0008-search-across-mvp.md) for further discussion of requirement. From the RFC, the broad approach will be 

> * Search will be run against the root collection, available to authenticated users only.
> * The capability will be advertised as a new `"service"` on the root collection.
> * It will search all label values of every resource in the root collection, regardless of nesting depth.
> * A synthetic collection will be returned supporting paging.

All ACs are for authenticated users providing `x-iiif-cs-show-extras` header only.

ACs
* Add new endpoint, `GET /{customerId}/collections/{collectionId}/search`.
  * The only supported `collectionId` is "root" (404 for any others)
  * Support `label` query parameter. If the value has < 3 characters it's a bad request. If >= 3 then run search. The default characters numbers should be controllable via configuration (default 3).
  * Support `page` and `pageSize` query parameters.
  * Hardcode ordering to `created` date.
* Handle search as outlined in [RFC#label-search](https://github.com/dlcs/iiif-presentation/blob/develop/docs/rfcs/0008-search-across-mvp.md#label-search), summarised as
  * Search is case insensitive
  * Search logic is like/contains ("hat" would match "hat", "that", "hatchet" etc)
  * If multiple words are supplied use multiple like/contains with _AND_.
  * _Search against unindexed field so need to verify performance. Run SQL query manually against production instances to check._
* Add new `"service"` to root collection
* Render results in synthetic collection
  * If no results, return a 200 with empty collection
  * Render `"id"`, `"label"` and `"type"` for Manifests and Collections. Include `"behavior"` for Collections (this is the same as current storage-collection rendering).
  * Include `"PartialCollectionView"`

donaldgray:
The returned synthetic collection contains a minimal number of properties. `Created`, `CreatedBy`, `Modified`, `Slug`, `PublicId` etc are all dropped - these are properties of the collection being searched, they are not relevant for the search results. The `"label"` property is _not_ the label of the collection but a synthetic one highlighting this is for search results.

Included a `"seeAlso"` link back to the collection being searched.

PR #635, opened 2026-07-13
Search across

### LINKED ISSUE #466: @context should be first property in API response
Flat/API responses contain standard IIIF properties and some custom element (`publicId`, `slug` etc).

All of the custom elements appear first - we should change this to ensure `@context` is the first property returned. As detailed in https://iiif.io/api/presentation/3.0/#46-linked-data-context-and-extensions

>The top level resource in the response MUST have the `@context` property, and it SHOULD appear as the very first key/value pair of the JSON representation. This tells Linked Data processors how to interpret the document. 

I think it makes sense to keep `"id"` and `"type"` should be elements 2 + 3 too.

donaldgray:
## What does this change?

Implements #632
Implements #631
Fixes #466 

Adds a new search endpoint for root collection, `{customer}/collections/root/search` with paging and ordering. This is surfaced as a `"service"` on the root collection. Summary of changes

* Slight refactoring of `GetCollection` to have some shared helpers for dealing with paged and orderable collections. We can take this further if more endpoints will use. Allowed pagination processing out of controller to keep it thin.
* Audit dates on `IPresentation` are now nullable, as the synthetic search-results collection won't have them.
* `CollectionController` uses `HandleFetch()` handler. This is the only place these are used so had to alter to fit our needs, I suspect they were migrated from Protagonist but never used.
* Added `[RequiredShowExtras]` attribute, which allowed removal of `if (!Request.HasShowExtraHeader()) return this.Forbidden();` in a few places. 
  * Also noticed a bug in the generated problem `Instance` value, which is resolved. Previously it was just the hostname and wouldn't output the path.

> [!WARNING]
> Search is MVP implementation and runs unindexed scan twice per search - once for count and once for getting the actual results. For large result sets and deeper pages we will need to monitor performance.

## Database Migration

> [!NOTE]
> Details of migration. What does it change? Is it breaking/non-breaking?
>
> - **Migration name:** `20260713084444_Manifest label is jsonb`
> - **What it does:** Converts `manifest.label` to `jsonb` type. Note that this takes an `ACCESS EXCLUSIVE` lock and rewrites the whole table, this is a write-blocking outage for the duration.
> - **Breaking Change?** No. 

## Configuration Changes

> [!NOTE]
>
> |---|---|---|---|---|
> | API | `MinSearchLength` | N | Minimum characters required for a search term to be run | `3` |
> | API | `SlowSearchThresholdMs` | N | Search queries taking longer than this are logged as a warning | `1000` |

### DIFF
--- a/src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
+++ b/src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
@@ -594,8 +594,8 @@ public class CollectionConverterTests
         searchCollection.Behavior.Should().BeNull();
         searchCollection.Totals.Should().BeNull("descendant counts can't be derived from a page of results");
         searchCollection.ItemsOrder.Should().BeNull();
-        searchCollection.Created.Should().Be(default);
-        searchCollection.Modified.Should().Be(default);
+        searchCollection.Created.Should().BeNull();
+        searchCollection.Modified.Should().BeNull();
         searchCollection.CreatedBy.Should().BeNull();
         searchCollection.ModifiedBy.Should().BeNull();
     }
--- a/src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
+++ b/src/IIIFPresentation/API.Tests/Converters/CollectionConverterTests.cs
@@ -4,6 +4,7 @@ using DLCS;
 using IIIF.Presentation.V3;
 using IIIF.Presentation.V3.Strings;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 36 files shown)


donaldgray:
> [!NOTE]
> As discussed internally with @tomcrane and @JackLewis-digirati , will merge this without review due to team annual leave.

Confirming I used Claude Opus 4.8 for review, which noted a couple of minor issues, resolved in https://github.com/dlcs/iiif-presentation/pull/635/commits/56a93bf764a343e78fa301cf7f36975a68c7a3f0

PR #638, opened 2026-07-16
Resolve issues with generated error URIs containing queryparams

### LINKED ISSUE #637: `"instance"` and `"type"` in problem details response can contain query-params
The helper for generating problem responses isn't handling query parameters correctly. 

* `"type"` is rendered in format `https://{baseURL}?{query-params}/{error-type}`.
* `"instance"` is rendered in format `https://{baseURL}?{query-params}`.

E.g. if I make a request to `/99/collections/root/search?label=as+xx&page=10`, this will be rejected as `label` must contain at least 1 token that is 3+ characters. The returned error is:

```json
{
    "title": "Bad request",
    "detail": "At least one search term must be 3 characters or more",
    "instance": "https://presentation-api.example?label=as+xx&page=453",
    "type": "https://presentation-api.example?label=as+xx&page=453/errors/ModifyCollectionType/ValidationFailed",
    "status": 400
}
```

The inclusion of the query-parameters is incorrect, this should be something like:

```json
{
    "title": "Bad request",
    "detail": "At least one search term must be 3 characters or more",
    "instance": "https://presentation-api.example",
    "type": "https://presentation-api.example/errors/ModifyCollectionType/ValidationFailed",
    "status": 400
}
```

> [!NOTE]
> The code that generates this is a generic, reusable helper so check call sites to see if this can be happening elsewhere.

donaldgray:
## What does this change?

Resolves #637 

Resolve issue where automatically generated "instance" and "type" URLs will include query-parameters. For the latter, also tweak how URL is constructed to avoid using simple string concatenation, which resulted in query-param being output mid-url.

Also removed `PathBase` appSetting. This isn't used anywhere (ie `.UsePathBase()` isn't called) but inclusion in settings makes it look like it could be.

Not strictly part of fix but switched failed collection searches to use a custom type, InvalidSearchQuery, rather than generic validation error (felt okay to add as bug was found as part of testing search).

### DIFF
--- a/src/IIIFPresentation/API.Tests/Infrastructure/ControllerBaseXTests.cs
+++ b/src/IIIFPresentation/API.Tests/Infrastructure/ControllerBaseXTests.cs
@@ -119,7 +119,6 @@ public class ControllerBaseXTests
     [Fact]
     public void PresentationProblem_Correct_WithDefaults_RequestHasPathBase()
     {
-        // Setup a basic request with no query params
         var sut = GetController(request =>
         {
             request.Scheme = "https";
@@ -142,7 +141,6 @@ public class ControllerBaseXTests
     [Fact]
     public void PresentationProblem_Correct_WithDefaults_RequestHasQueryParam()
     {
-        // Setup a basic request with no query params
         var sut = GetController(request =>
         {
             request.Scheme = "https";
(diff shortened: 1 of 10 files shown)


donaldgray:
> [!NOTE]
> As discussed internally with @tomcrane and @JackLewis-digirati , will merge this without review due to team annual leave.

Confirming I used Claude Opus 4.8 for review, which noted a couple of minor issues (incorrect placement of `[MemberNotNullWhen]` and a copy/paste comment in tests.

PR #641, opened 2026-07-30
Hierarchical consistent PUT/POST

### LINKED ISSUE #291: Refactor Collection write requests
There are currently 3 mediatr handlers that deal with writing Collections

* `CreateCollection`
* `UpsertCollection`
* `PostHierarchicalCollection`

Consider refactoring these to share more code as there is a lot of duplication across these requests, particularly the first 2.

### LINKED ISSUE #464: Consistent API PUT and POST behaviour

JackLewis-digirati:
## What does this change?

Resolves #464 
Resolves #291

This PR is a reimplemented version of the stale #503 which had gotten very out of date.

It has 2 main goals:

- Refactor write collections to share more code
- Implement hierarchical PUT and change the hierarchical POST to be more consistent for collections
- Implement PUT/POST for collections

In order to complete this goal, 3 new handlers have been introduced and `postHierarchical` is modified extensively.  Additionallly, a new central class for collection requests has been introduced to mirror the `ManifestWriteService` called the `CollectionWriteService` to reduce duplication.

Finally, several helpers have been introduced to help validate and respond to the new/modified hierarchical requests.

> [!Note]
> The original PR implemented a hierarchical `DELTE` endpoint which has not been reimplemented here to reduce the size/scope of an already lrge PR

> [!Note]
> The ticket notes a breaking change in the stale branch.  This has been fixed so that `Post` works with the `into` logic, requiring a `publicId`, `id` and/or `slug` property to correctly derive the slug  (with `id` being ignored if the host doesn't match the value from settings)


JackLewis-digirati:
## Summary

Implements #464 (consistent PUT/POST behaviour). Old PR #503 was used purely as a design reference (unmergeable/stale) — this is a fresh implementation that extends the existing `ParentSlugParser`/flat-endpoint machinery rather than replacing everything with a per-type mega-handler.

**New/changed capabilities:**
- Hierarchical PUT now exists for both Collections and Manifests (`PutHierarchicalCollection`, `PutHierarchicalManifest`) — previously only hierarchical POST existed, and only for Collections.
- Hierarchical POST now exists for Manifests (`PostHierarchicalManifest`).
- `id`/`parent`/`slug` can now be supplied consistently from the URL, the body's `parent`/`slug`/`publicId`, or the body's `id` (own-host, resolved via new `RequestIdResolver`) across all 6 endpoint combinations (flat/hierarchical × POST/PUT × Collection/Manifest). Conflicting sources are rejected (400) rather than one silently winning.
- New `CollectionWriteService`/`ICollectionWrite`, mirroring the existing `ManifestWriteService` shape (`Create`/`Upsert` → shared internals). `CreateCollection`/`UpsertCollection` are now thin MediatR wrappers, matching `CreateManifest`/`UpsertManifest`.
- Common hierarchical prologue (deserialize → validate → resolve id → reconcile parent/slug → resolve internal id for PUT) extracted into `HierarchicalRequestHelper.PrepareForPost`/`PrepareForPut`, shared by all 4 hierarchical handlers so future validation/rule changes only need to be made once.

## Bugs fixed (pre-existing, on `develop`)

The old `PostHierarchicalCollection` handler had several bugs that are fixed here, not just extended:
- It treated the **entire** POSTed-to URL path as the new resource's own slug, instead of as the *parent* it should be created inside (so `POST /1/foo/bar` created a resource at slug `bar` under `foo` — correct by accident — but `POST /1/foo` created a resource *named* `foo`, colliding with/replacing the parent's own address, instead of creating a child of `foo`).
- It ignored the body's `slug`/`parent`/`publicId` entirely, and deserialized the body as a bare IIIF `Collection`, not `PresentationCollection`.
- It hardcoded `IsStorageCollection = false` regardless of what the client sent — storage collections could not be created via hierarchical POST at all. (Confirmed against #503's own reference implementation that this was unintended — it computes `IsStorageCollection` from the body the same way flat endpoints do.)
- No ETag/update handling — every hierarchical POST minted a brand-new id, so there was no way to hit an existing resource.

## ⚠️ Breaking change

**Collection PUT-create now returns `201 Created` instead of `200 OK`.** Previously, `UpsertCollectionHandler` had a bug where both the create-via-PUT and update-via-PUT branches fell through to one shared return with the default `WriteResult.Updated` (200) — so a PUT that created a brand-new collection incorrectly returned 200, unlike POST-create (correctly 201) and unlike Manifests (`ManifestWriteService` already returned 201 correctly for both paths). This is now consistent: any PUT that creates returns 201, for both resource types. Any consumer relying on the old 200-on-create-via-PUT behaviour for Collections will see a status code change.

## Notes on hierarchical response shape

- Hierarchical POST/PUT responses are always plain IIIF (no `Prefer`/show-extras support) — out of scope for this ticket per earlier discussion.
- For **Manifests**, the response is built via `PresentationIIIFCleaner.OnlyIIIFProperties` — a reflection-based copy of only base-IIIF properties off the enriched `PresentationManifest`, since manifests' enrich step never rewrites `Behavior`.
- For **Collections**, the enrich step (`EnrichPresentationCollection`) *does* unconditionally regenerate `Behavior` from just two DB booleans (`IsPublic`/`IsStorageCollection` — the `Collection` DB entity has no column for arbitrary behaviour strings), which would silently drop custom values like `auto-advance`. So non-storage-collection hierarchical responses instead re-parse the client's raw request body directly (new shared `HierarchicalCollectionResponse.Build`, used by both `PostHierarchicalCollection`/`PutHierarchicalCollection`) to preserve those. This is a real, pre-existing asymmetry between Collections and Manifests, not something introduced here — flagged in case it's worth a follow-up ticket to let Collections persist arbitrary behaviours too.

## Test coverage added

- Full flat/hierarchical × POST/PUT × id/parent/slug-source matrix (`ManifestPaths_ResolveIdParentSlug_FromAllDocumentedSources`, 15 cases).
- Hierarchical PUT create/update, multi-segment path (parent-splitting), body-`id`-vs-URL conflict rejection, and body-`id`-honoured-on-create — previously only exercised by isolated unit tests on `HierarchicalRequestHelper`'s now-private `SplitPath`/`ReconcileId`/`ResolveIdForPut` methods; those unit tests were removed once caller-level coverage closed the gap.
- Storage collections created via hierarchical POST and PUT (previously untested at the hierarchical level).

Full solution test suite (1561 tests) passes.

src/IIIFPresentation/API.Tests/Integration/ModifyManifestCreateTests.cs
```diff
@@ -405,7 +405,96 @@ public async Task CreateManifest_CreatesManifest_ParentIsValidHierarchicalUrl()
...
+        // hierarchical responses are always plain IIIF, regardless of the show-extras header
+        var responseManifest = await response.ReadAsPresentationJsonAsync<IIIF.Presentation.V3.Manifest>();
+        responseManifest!.Id.Should().Be($"http://localhost/{Customer}/{slug}");
+
+        var hierarchyFromDatabase = dbContext.Hierarchy.First(h => h.Slug == slug);
+        hierarchyFromDatabase.Parent.Should().Be(RootCollection.Id);
+    }
+
+    [Fact]
+    public async Task PostHierarchical_WithAsset_IngestsAsset_WithoutShowExtrasHeader()
```
JackLewis-digirati, line 438: I wasn't sure about this, but erred on the side it should be allowed - this test is basically proving you can get asset ingestion from a hierarchical path (and it will do it without the show-extras header, but the `[Authorize]` means you need to be authed)

src/IIIFPresentation/API/Features/Manifest/Requests/PostHierarchicalManifest.cs
```diff
@@ -0,0 +1,62 @@
...
+using Models.API.Manifest;
+using Services.Manifests.Settings;
+
+namespace API.Features.Manifest.Requests;
+
+/// <summary>
+/// Create a new Manifest as a child of the container addressed by the request URL
+/// (POST /{customer}/{parent-path}).
+/// </summary>
+public class PostHierarchicalManifest(
```
donaldgray: The Mediatr requests should describe the use-case being actioned rather than the HTTP verb used to trigger it. `CreateHierarchicalManifest` better documents what this is doing, `PostHierarchicalManifest` is telling you how it was triggered. 

Naming after intent would allow it to be triggered by an alternative means, say a message queue listener, and it would still make sense. If on message receipt we ran `mediatr.Send(new PostHierarchicalManifest(...));` it doesn't hold up but `mediatr.Send(new CreateHierarchicalManifest(...));` does.

For this reason I think the HTTP route shouldn't be in the comments, it should describe what this class does - not how it is called.

src/IIIFPresentation/API/Features/Storage/Helpers/HierarchicalCollectionResponse.cs
```diff
@@ -0,0 +1,53 @@
+using API.Features.Common.Helpers;
+using API.Infrastructure.Requests;
+using Core.Helpers;
+using IIIF.Presentation;
+using Models.API.Collection;
+
+namespace API.Features.Storage.Helpers;
+
+/// <summary>
+/// Builds the plain-IIIF response shared by the hierarchical Collection POST/PUT handlers.
```
donaldgray: Mentioned similar in handlers but noticing it across the PR, comments should document what this does, not what uses it. If we introduce a hierarchical collection PATCH that calls this class we shouldn't need to update the comments. The only thing that should result in a comment being updated is a change in behaviour of the thing carrying the comment, not a change in caller.

src/IIIFPresentation/API/Features/Storage/Helpers/PresentationContextX.cs
```diff
@@ -50,6 +51,11 @@ public static class PresentationContextX
                     ModifyCollectionType.ManifestCurrentlyIngesting, WriteResult.Conflict);
             }
 
+            if (ex.IsCollectionPrimaryKeyViolation())
```
donaldgray, line 47: If we're including this here and have refactored the collection write logic, should we get rid of `TrySaveCollection` and just call `TrySave` directly? I suspect the former was left to avoid breaking when the latter was introduced? If we've refactored both Manifest and Collection there's no need for them to differ in how the call this method.

src/IIIFPresentation/API/Features/Storage/StorageController.cs
```diff
@@ -94,12 +95,54 @@ public async Task<IActionResult> GetHierarchical(int customerId, string slug = "
...
-    public async Task<IActionResult> PostHierarchicalCollection(int customerId, string slug)
+    public async Task<IActionResult> PostHierarchical(int customerId, string slug = "")
     {
         // X-IIIF-CS-Show-Extras is not required here, the body should be vanilla json
         var rawRequestBody = await Request.GetRawRequestBodyAsync();
-        return await HandleUpsert(new PostHierarchicalCollection(customerId, slug, rawRequestBody));
+
+        IRequest<PresentationResult>? request = JsonPropertyReader.ReadJsonProperty(rawRequestBody, "type", logger: logger) switch
+        {
+            "Manifest" => new PostHierarchicalManifest(customerId, slug, rawRequestBody,
```
donaldgray: Small, consider `nameof(Manifest)` and `nameof(Collection)`?

src/IIIFPresentation/API/Features/Storage/StorageController.cs
```diff
@@ -94,12 +95,54 @@ public async Task<IActionResult> GetHierarchical(int customerId, string slug = "
...
+            "Collection" => new PutHierarchicalCollection(customerId, slug, rawRequestBody, Request.Headers.IfMatch),
+            _ => null
+        };
+
+        if (request == null) return UnrecognisedTypeProblem();
+
+        return await HandleUpsert(request, invalidatesEtag: Request.Headers.IfMatch);
+    }
+
+    private ObjectResult UnrecognisedTypeProblem() =>
```
donaldgray, line 145: `ControllerBaseX` has `PresentationNotFound` already - should we use `PresentationBadRequest`, small but would tidy this up a little and be reusable elsewhere.

```cs
this.PresentationProblem(
            "Could not determine resource 'type' from the request body - expected 'Collection' or 'Manifest'", 
            this.GetErrorType(ModifyCollectionType.CannotDeserialize));
```

src/IIIFPresentation/API/Helpers/HierarchicalRequestHelper.cs
```diff
@@ -0,0 +1,229 @@
...
+    public static DeserializedRequest<T> Success(T presentation, ResolvedRequestId resolvedId) =>
+        new() { Presentation = presentation, ResolvedId = resolvedId };
+}
+
+/// <summary>
+/// Shared orchestration for hierarchical POST/PUT request handling, used identically by the Collection and Manifest
+/// hierarchical handlers. Centralising this here means any future rule that should apply to hierarchical writes of
+/// both resource types (a new validation step, a new id/parent/slug source, etc.) only needs to be added once.
+/// </summary>
+public static class HierarchicalRequestHelper
```
donaldgray: The idea behind this is good but the public methods have 8 and 11 parameters, which seems excessive. 

Is it a sign that this shouldn't be a static class, and instead something that is injected or forms a base class that can be inherited from? `ILogger,` `IRequestIdValidator`, `IValidator<T>` etc are dependencies - these can be injected rather than passed in.

src/IIIFPresentation/API/Features/Storage/CollectionController.cs
```diff
@@ -90,28 +92,42 @@ public async Task<IActionResult> Search(int customerId, string id, string? label
...
     [RequireShowExtras]
     [HttpPost("collections")]
-    public async Task<IActionResult> Post(int customerId, [FromServices] PresentationValidator validator)
+    public async Task<IActionResult> Post(int customerId, [FromServices] PresentationValidator validator,
+        [FromServices] IRequestIdResolver requestIdResolver)
     {
         var deserializeValidationResult = await DeserializeAndValidate(validator, null, null);
         if (deserializeValidationResult.HasError) return deserializeValidationResult.Error;
 
+        var resolvedId = requestIdResolver.Resolve(customerId, deserializeValidationResult.ConvertedIIIF.Id);
```
donaldgray: Should this happen beyond the controller? It happens in the controller for flat requests but beyond for hierarchical.

src/IIIFPresentation/API.Tests/Integration/ModifyManifestCreateTests.cs
donaldgray: @tomcrane flagging this for opinion.

src/IIIFPresentation/API/Features/Manifest/Requests/PostHierarchicalManifest.cs
JackLewis-digirati: changed these + `HierarchicalRequestHelper`

tomcrane:
ShowExtras is not required on a PUT or POST, because the server can tell whether the request payload is vanilla or extras. ShowExtras is not an auth header, it's an instruction about what to include in responses.

BUT - if you make a PUT or POST that eventually returns a body, showExtras on the original PUT or POST determines whether or not extras are included in that body.

This means that a "vanilla" authed manifest editor that only submits vanilla IIIF doesn't get an extras response it won't understand. 

This applies to flat or hierarchical PUT and POST. (needs new ticket)

tomcrane:
**Handoff from the DLCS docs/API hygiene sprint** (cross-cutting session, 2026-08-06) — two conventions were ratified today that this PR's PUT/POST standardisation is the natural landing point for on the iiif-presentation side:

1. **Create-POST returns `201 Created` + Location.** The only sanctioned exceptions are *action POSTs* returning an ephemeral payload with no canonical URL (in protagonist: API-key creation and application setup, which return 200). Any new exception must be named explicitly, never copied silently.

2. **PUT upsert: `201` on create, `200` on replace** — the handler must report Updated vs Created truthfully (never hard-code Created on a known-update path; protagonist had exactly one such bug, live since 2023, fixed today in dlcs/protagonist#1236). A true-upsert PUT annotates/documents both 200 and 201; an *update-only* PUT (404s rather than creates) annotates exactly what it does, e.g. 200/400/404.

Also ratified, for completeness: **every DELETE returns `204 No Content` with an empty body** (404/409/500 as error bodies) — protagonist's two legacy 200-with-body deletes are migrating in the same PR (breaking change, signposted there).

No action requested beyond alignment — the room's intent is that this PR carries the convention on this repo rather than a parallel ruling being minted. Register trail: `scratch/hygiene-sprint/session-0-cross-cutting.md` in dlcs/public-docs (cards XC-01/02/03).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

src/IIIFPresentation/API.Tests/Integration/ModifyManifestCreateTests.cs
JackLewis-digirati: leaving as-is for now, but https://github.com/dlcs/iiif-presentation/issues/648 created to follow up on this

src/IIIFPresentation/API/Converters/PresentationIIIFCleaner.cs
```diff
@@ -0,0 +1,44 @@
...
+/// Hierarchical write responses are always plain IIIF, regardless of headers. This strips Presentation-only
+/// properties (e.g. Slug, Parent, PublicId) off a Presentation* model, copying only the base IIIF properties across
+/// to a fresh instance of the base type.
+/// </summary>
+public static class PresentationIIIFCleaner
+{
+    private static readonly PropertyInfo[] ManifestProperties = GetCopyableProperties(typeof(Manifest));
+    private static readonly PropertyInfo[] CollectionProperties = GetCopyableProperties(typeof(Collection));
+
+    public static Manifest OnlyIIIFProperties(PresentationManifest presentationManifest)
```
donaldgray: Not a huge change but IMO this would read better to caller as a `ToManifest()` extension method.

```cs
// What is "manifest" - is it a PresentationManifest with props removed, or a Manifest?
var manifest = PresentationIIIFCleaner.OnlyIIIFProperties(presentationManifest);

// More self documenting, method explains what it returns. Follows ToArray(), ToList() etc. 
var manifest = presentationManifest.ToManifest();
```

src/IIIFPresentation/API/Features/Manifest/Helpers/HierarchicalManifestResponse.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace API.Features.Manifest.Helpers;
+
+/// <summary>
+/// Builds the plain-IIIF response body for a hierarchical Manifest write.
+/// </summary>
+public static class HierarchicalManifestResponse
+{
+    /// <summary>
+    /// Guards a <see cref="ManifestWriteService"/> result and, on success, strips it down to plain IIIF via
```
donaldgray: What does `"Guards a <see cref="ManifestWriteService"/> result"` mean, is that a left over, or agent "here is where this is used" comment?

src/IIIFPresentation/API/Features/Manifest/ManifestWriteService.cs
```diff
@@ -56,21 +58,24 @@ public class WriteManifestRequest
     public WriteManifestRequest(int customerId,
         PresentationManifest presentationManifest,
         string rawRequestBody,
-        bool createSpace)
+        bool createSpace,
+        ResolvedLocation? location = null)
```
donaldgray: It doesn't look like any 'real' call paths can pass a null location here, it's only tests that use this path, feels like if that is the case it should be removed to not-null (same with overload and CollectionWriteService)

src/IIIFPresentation/API/Features/Storage/CollectionWriteService.cs
```diff
@@ -0,0 +1,348 @@
...
+                    Slug = parsedParentSlug!.Slug,
+                    Canonical = true,
+                    ItemsOrder = request.Collection.ItemsOrder,
+                    Parent = parsedParentSlug.Parent!.Id
+                }
+            ]
+        };
+        SetCommonProperties(collection, request.Collection, dateCreated);
+
+        dbContext.Collections.Add(collection);
```
donaldgray, line 199: IMO this would read better if this was collapsed to a method, similar to ManifestWriteService. 

```cs
var (error, dbManifest) = await CreateCollection(...);
if (error != null) return error;

await using var transaction = await dbContext.Database.BeginTransactionAsync(cancellationToken);
```

src/IIIFPresentation/API/Helpers/ParentSlugParserX.cs
```diff
@@ -0,0 +1,26 @@
...
+using Models.API;
+
+namespace API.Helpers;
+
+public static class ParentSlugParserX
+{
+    /// <summary>
+    /// Resolves the parent/slug for a write request, unwrapping the result into an error-or-value pair.
+    /// </summary>
+    public static async Task<(PresentationResult? error, ParsedParentSlug? parsedParentSlug)> ParseParentSlug<T>(
```
donaldgray: Could this just be a method on `IParentSlugParser`?

src/IIIFPresentation/API/Helpers/ParentSlugParserX.cs
```diff
@@ -0,0 +1,26 @@
...
+        this IParentSlugParser parentSlugParser,
+        T presentation,
+        int customerId,
+        string? id,
+        string? urlParentPath,
+        string? urlSlug,
```
donaldgray: Small, these always come from a `ResolvedLocation` so should we just pass that?

donaldgray (approved):

src/IIIFPresentation/API/Converters/PresentationIIIFCleaner.cs
JackLewis-digirati: `ToCollection()` also done

src/IIIFPresentation/API/Features/Manifest/Helpers/HierarchicalManifestResponse.cs
JackLewis-digirati: it's based on the check for if it's a presentation manifest - though it's unclear, so clarified the comment

src/IIIFPresentation/API/Helpers/ParentSlugParserX.cs
JackLewis-digirati: I left it separate as it would bloat the interface, but it's also a fairly meaningless abstraction as it can still be seen on the call sites - will fold it in

PR #645, opened 2026-08-04
Disallow slugs from having slashes or FQDN's in them

### LINKED ISSUE #571: Slugs can contain forward slashes/FQDNs (low priority)
Following a discussion on https://github.com/dlcs/iiif-presentation/issues/527, from @JackLewis-digirati:

>Values that need to be updated to check customer id:
>- canvas id
>- parent
>- public id
>- _**slug (accepts an FQDN, which should be invalid? - for another ticket perhaps?)**_
>- additionally, forward slashes are allowed in the slug, which then can't be retrieved via the public id
>- mixing customer id between parent + public id also doesn't throw an error - but should be trumped by mismatches between the calling customer

Slugs can contain forward slashes or be a FQDN, which prevents the Presentation resource from being retrieved via its public ID. 

> [!NOTE]
> Low priority

JackLewis-digirati:
## What does this change?

Resolves #571

Makes it so slugs aren't allowed to have FQDN's or forward slashes.  This is also extendable, due to adding a setting controlling characters that are disallowed from use in slugs.

## Configuration Changes

> [!NOTE]
>
> |---|---|---|---|---|
> | API | `ApiSettings:ProhibitedSlugCharacters` | N | Prohibited characters in the slug | `["/"]` |

### DIFF
--- a/src/IIIFPresentation/API/Settings/ApiSettings.cs
+++ b/src/IIIFPresentation/API/Settings/ApiSettings.cs
@@ -36,7 +36,15 @@ public class ApiSettings
     /// The maximum number of historical pipeline jobs returned in a Manifest's "finishedPipelines" property
     /// </summary>
     public int FinishedPipelinesLimit { get; set; } = 20;
-    
+
+    /// <summary>
+    /// Characters that are not permitted in a 'slug'
+    /// </summary>
+    public string[] ProhibitedSlugCharacters { get; set; } = ["/"];
+
+    public string ProhibitedSlugCharactersDisplay =>
+        string.Join(", ", ProhibitedSlugCharacters.Select(p => $"'{p}'"));
+
     public required AWSSettings AWS { get; set; }
 
     public required DlcsSettings DLCS { get; set; }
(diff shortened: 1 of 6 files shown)


src/IIIFPresentation/API/Settings/ApiSettings.cs
```diff
@@ -36,7 +36,15 @@ public class ApiSettings
     /// The maximum number of historical pipeline jobs returned in a Manifest's "finishedPipelines" property
     /// </summary>
     public int FinishedPipelinesLimit { get; set; } = 20;
-    
+
+    /// <summary>
+    /// Characters that are not permitted in a 'slug'
+    /// </summary>
+    public string[] ProhibitedSlugCharacters { get; set; } = ["/"];
```
donaldgray, line 43: Given the comment is `Characters that are not permitted in a 'slug'` should this be `char[]`?

donaldgray (commented):
> |---|---|---|---|---|
> | API | `ApiSettings:ProhibitedSlugCharacters` | Y | Prohibited characters in the slug | `["/"]` |

I think `"Required?"` should be `N` for this. It's not required as it has a default, and (untested) it looks like if you set it to an empty array it'd just let through every character.

src/IIIFPresentation/API/Settings/ApiSettings.cs
JackLewis-digirati: As discussed, swapping to say `Strings that are not permitted in a 'slug`

JackLewis-digirati:
Changed **Required** to be N following discussion

donaldgray (approved):

PR #646, opened 2026-08-04
Space not being an integer should return a bad request

### LINKED ISSUE #572: Supplying a `paintedResource.asset` with a string value for `"space"` returns 500 (low priority)
Discovered while testing https://github.com/dlcs/iiif-presentation/pull/559:

If a painted resource is supplied with a string-based "space" value, like so:
```JSON
...
"asset": {
        "id": "world-of-cheese",
        "mediaType": "image/jpeg",
        "origin": "https://dlcsstage-public-test-objects.s3.eu-west-1.amazonaws.com/wellcome-images/cheese/b15839643_0001.jpg",
        "space": ""
      }
...
```
Presentation will throw an unknown 500 error:
```JSON
{
    "title": "Operation failed",
    "detail": "Unexpected error creating manifest",
    "instance": "https://presentation-api.dlcs.digirati.io/",
    "type": "https://presentation-api.dlcs.digirati.io/errors/ModifyCollectionType/Unknown",
    "status": 500
}
```

> [!NOTE]
> Low priority

JackLewis-digirati:
## What does this change?

Resolves #572

Makes it so that a space being passed that isn't an integer will respond correctly with a `BadRequest`

### DIFF
--- a/src/IIIFPresentation/Services/Manifests/ManifestPaintedResourceParser.cs
+++ b/src/IIIFPresentation/Services/Manifests/ManifestPaintedResourceParser.cs
@@ -179,13 +179,31 @@ public class ManifestPaintedResourceParser(
         var adjuncts = asset.TryGetCollectionValue<JObject>(AssetProperties.Adjuncts);
         asset.Remove(AssetProperties.Adjuncts);
         var id = asset.GetRequiredValue<string>(AssetProperties.Id);
-        var space = asset.TryGetValue<int?>(AssetProperties.Space);
+        var space = GetSpace(asset, id);
         return new AssetDetails
         {
             Space = space,
             AdjunctInteraction = adjuncts != null ? HydrateAdjuncts(adjuncts, space, customerId, id) : null,
             Id = id
         };
+    }
+
+    /// <summary>
+    /// Parse the "space" property of an asset, throwing an <see cref="AssetException"/> rather than an
+    /// unhandled <see cref="FormatException"/> if a non-numeric value (e.g. a string) is supplied
+    /// </summary>
+    private static int? GetSpace(JObject asset, string assetId)
+    {
+        try
+        {
+            return asset.TryGetValue<int?>(AssetProperties.Space);
+        }
+        catch (FormatException)
+        {
+            var value = asset[AssetProperties.Space]?.ToString();
+            throw new AssetException($"The space for asset '{assetId}' is '{value}' and is not a valid integer",
+                assetId);
+        }
     }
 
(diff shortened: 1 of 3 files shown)


src/IIIFPresentation/Services/Manifests/ManifestPaintedResourceParser.cs
```diff
@@ -188,6 +189,36 @@ private static AssetDetails GetCanvasPaintingDetailsForAsset(JObject asset, int
         };
     }
 
+    /// <summary>
+    /// Parse the "space" property of an asset, throwing an <see cref="AssetException"/> rather than an
+    /// unhandled <see cref="FormatException"/>/<see cref="OverflowException"/> if a non-integer value (e.g. a
+    /// string, decimal, boolean, array, object, or a number too large to fit in an int) is supplied
+    /// </summary>
+    private static int? GetSpace(JObject asset, string assetId)
```
donaldgray: Is it worth including the `if (assetDetails.Space < 0)` check in this method, then this will only return `null` or a valid integer? The caller can rely on this to do any validation.

What happens with a space of `0`?

src/IIIFPresentation/Services/Manifests/ManifestPaintedResourceParser.cs
```diff
@@ -188,6 +189,36 @@ private static AssetDetails GetCanvasPaintingDetailsForAsset(JObject asset, int
...
+        }
+
+        try
+        {
+            return asset.TryGetValue<int?>(AssetProperties.Space);
+        }
+        catch (Exception ex) when (ex is FormatException or OverflowException)
+        {
+            throw new AssetException(
+                $"The space for asset '{assetId}' is '{spaceToken}' and is not a valid integer", assetId);
```
donaldgray: Can't remember what the other error messages are like in project but should the requirement come first?

Something like 
* This `Space for asset 'foo' must be a positive integer (got 'hello')`
* Instead of `The space for asset 'foo' is 'hello' and is not a valid integer` 

If this is the type of error we are outputting we should keep it consistent.

donaldgray (approved):

src/IIIFPresentation/Services/Manifests/ManifestPaintedResourceParser.cs
JackLewis-digirati: it would be parsed here, but you would eventually get an error because of the expectation that the manifest builder has that asset in it - likely a good idea to disallow specifying space 0 though as the error is the default `unexepected error`, so I'll make the change

